PR #7, opened 2018-04-05
Fix build on Windows using msys2/mingw-w64

LukeUsher:
The implementation is nasty in places, but it should be fine for now.

Hardcoded opengl libs in the mingw32 code-path because setting opengl = 'yes' was unable to locate them & manually setting opengl_libs broke the build.


mborgerson:
Thanks!

PR #13, opened 2018-06-01
WIP: do not include glew.h and gl.h twice

illwieckz:
`gloffscreen_wgl.c` includes `glew.h` and `gl.h` then includes `gloffscreen.h` that includes `glew.h` and `gl.h` again.

### DIFF
--- a/hw/xbox/nv2a/gl/gloffscreen.h
+++ b/hw/xbox/nv2a/gl/gloffscreen.h
@@ -41,10 +41,12 @@
 #include <GL/gl.h>
 #include <GL/wglext.h>
 #else                                  /* Assume GLX */
+#ifndef __gl_h_                        /* gloffscreen_wgl.c calls them first */
 #include <GL/glew.h>
 #include <GL/glx.h>
 #include <GL/glxext.h>
 #include <GL/gl.h>
+#endif
 #endif
 
 /* Used to hold data for the OpenGL context */


hw/xbox/nv2a/gl/gloffscreen.h
```diff
@@ -41,11 +41,13 @@
 #include <GL/gl.h>
 #include <GL/wglext.h>
 #else                                  /* Assume GLX */
+#ifndef __gl_h_                        /* gloffscreen_wgl.c calls them first */
```
JayFoxRox: This feels incredibly hacky.
The proper solution would be to arrange our includes properly or have some custom "gl.h" which handles including everything platform specific in the right order, so we have a functional GL implementation (which is what glo tries to be, I think. So we should fix its internal use of headers).

illwieckz: yep, this is horribly hacky :grin: but such hacky patchy patch is a good way to disclose the issue :smiley: 

Well it would raise some concerns about the project if this PR get merged as is without any comment or without any better attempt. :wink:

illwieckz: Added `WIP:` to the PR topic to make it more obvious

MayeulC: On the other hand, such hacks would be necessary at some point, either here or in an other file.

MayeulC:
Well, that solved my compilation issue, at least.

```
In file included from /home/mayeul/Documents/projets/xbox/xqemu/hw/xbox/nv2a/gl/gloffscreen.h:44,
                 from /home/mayeul/Documents/projets/xbox/xqemu/hw/xbox/nv2a/nv2a_shaders.h:25,
                 from /home/mayeul/Documents/projets/xbox/xqemu/hw/xbox/nv2a/nv2a.h:38,
                 from /home/mayeul/Documents/projets/xbox/xqemu/hw/xbox/nv2a/nv2a.c:27:
/usr/include/GL/glew.h:85:2: error: #error gl.h included before glew.h
 #error gl.h included before glew.h
  ^~~~~
/usr/include/GL/glew.h:97:2: error: #error glext.h included before glew.h
 #error glext.h included before glew.h
  ^~~~~
In file included from /home/mayeul/Documents/projets/xbox/xqemu/hw/xbox/nv2a/gl/gloffscreen.h:44,
                 from /home/mayeul/Documents/projets/xbox/xqemu/hw/xbox/nv2a/nv2a_shaders.h:25,
                 from /home/mayeul/Documents/projets/xbox/xqemu/hw/xbox/nv2a/nv2a.h:38,
                 from /home/mayeul/Documents/projets/xbox/xqemu/hw/xbox/nv2a/nv2a.c:27:
/usr/include/GL/glew.h:18730:28: error: conflicting types for 'PFNGLFRAGMENTLIGHTMODELFVSGIXPROC'
 typedef void (GLAPIENTRY * PFNGLFRAGMENTLIGHTMODELFVSGIXPROC) (GLenum pname, GLfloat* params);
```
I'm puzzled as to why this is necessary, however. Shouldn't the include guard prevent this from happening?

illwieckz:
> I'm puzzled as to why this is necessary, however. Shouldn't the include guard prevent this from happening?

That question crossed my mind too…

illwieckz:
> I'm puzzled as to why this is necessary, however. Shouldn't the include guard prevent this from happening?

That's probably because libepoxy included `GL.h` first (abstract from `epoxy/gl.h`):

```c
/** @file gl.h
 *
 * Provides an implementation of a GL dispatch layer using either
 * global function pointers or a hidden vtable.
 *
 * You should include `<epoxy/gl.h>` instead of `<GL/gl.h>` and `<GL/glext.h>`.
 */

#ifndef EPOXY_GL_H
#define EPOXY_GL_H

#include "epoxy/common.h"

#if defined(__gl_h_) || defined(__glext_h_)
#error epoxy/gl.h must be included before (or in place of) GL/gl.h
#else
#define __gl_h_
#define __glext_h_
#endif
```

mborgerson:
Probably going to ditch glew soon for epoxy, as upstream has done. Going to close this for now!

PR #15, opened 2018-06-05
Update .gitmodules

gandalfthewhite19890404:
Fixed submodules URLs - there was  git.qemu-project.org, it moved to  git.qemu.org.
Also for some reasons it does not work with git:// - so I edited it to https://.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,48 +1,48 @@
 [submodule "roms/vgabios"]
 	path = roms/vgabios
-	url = git://git.qemu-project.org/vgabios.git/
+	url = https://git.qemu.org/git/vgabios.git
 [submodule "roms/seabios"]
 	path = roms/seabios
-	url = git://git.qemu-project.org/seabios.git/
+	url = https://git.qemu.org/git/seabios.git/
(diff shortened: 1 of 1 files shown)


mborgerson:
Hmm, I'm pretty sure these submodules are not actually needed for build based on my testing. Let me come back to this later. As far as `git://` vs `https://`, are you behind a proxy perhaps? That can sometimes cause issues with the Git protocol.

gandalfthewhite19890404:
I am behind Stonesoft VPN HW Firewall.
I checked cxbx-r - it also uses https. So maybe it will be better to accept PR with `https:// `as working solution instead of `git://` (which could not work behind FWs)

JayFoxRox:
[This is from QEMU upstream](https://git.qemu.org/?p=qemu.git;a=blob;f=.gitmodules;h=49e9c2e3f4abc0321c6be09053445f5cb372cee9;hb=HEAD) (Link is to HEAD at time of writing). Please report it with them.
If we change it it will just harder to keep our version up-to-date if QEMU upstream also changes something.

For now, you can change this locally.

Sorry for the inconvenience; Closing.

PR #21, opened 2018-06-17
Build scripts: have them return the same error as make on failure

MayeulC:
This is done trough the pipefail option, that returns the first-nonzero
return value when programs are piped together.
This commit also succinctly comments the other options used.

I am not too happy with the amount of code duplication there. Couldn't the build options be saved somewhere else, or even in the same script which would autodetect the platform?

### DIFF
--- a/build_linux.sh
+++ b/build_linux.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
-set -e
-set -x
+
+set -e # exit if a command fails
+set -x # Print commands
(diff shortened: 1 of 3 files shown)


MayeulC:
Oh, I also didn't mention that this also makes the scripts output the compilation log to the logfile. On my system at least, gcc outputs the errors to stderr, and only stdout was redirected to build.log.

Feel free to add it to the log when merging.

mborgerson:
Cool, thanks!

PR #52, opened 2018-06-27
Add support for window clipping

### LINKED ISSUE #44: NV2A Window clipping is broken
Running https://github.com/xqemu/xqemu/commit/5d4058d6f25ec09be8a3e581cc02b419a89e2fe0 ; the right half of the display is broken in this screenshot:

Right half broken

*(Screenshot provided by @DarkGabbz)*

This is due to broken window clipping (which was merged in a premature state).
Even if it's not actually a regression, it will be observed as one in most applications (as window clipping usually contributes little, but can break a lot).

---

I believe this issue is due to anti-aliasing in combination with the clipping; which was never tested.

As the actual clipping uses OpenGL pixel coordinates, it has to respect anti-aliasing (which it currently doesn't). So [these uniforms need `pgraph_apply_anti_aliasing_factor`](https://github.com/xqemu/xqemu/commit/c50738b732bd830223691cc317fa60dc112031cc#diff-d936f94a7fe1cbc634cedf01266670e5R3157).

The independence of the clipping optimization (which only scans for redundant regions) and implementation / uniform upload (which does clipping) should also be made more clear in the source code, by adding a comment near the optimization loop.
There should be a short review here though, because I have written this code a long time ago, and it might actually do more than optimizing I could have forgotten about.

That said, I believe the clipping optimization should not respect anti-aliasing. It works with coordiantes (`surface_shape.clip_width` and `surface_shape.clip_height`) which are not affected by anti-aliasing (unless our `surface_shape` is bad).
If optimization / implementation are really independent, then regardless of this issue, [this comment should be removed](https://github.com/xqemu/xqemu/commit/c50738b732bd830223691cc317fa60dc112031cc#diff-d936f94a7fe1cbc634cedf01266670e5R3066) or moved to the actual implementation (but that has been solved by my above observation probably).

mborgerson:
Resolves #44

This PR adds partial support for window clipping. This fixes games which utilize clipping, such as Halo:CE which uses clipping in multiplayer to render multiple player screens.

Research and original PoC work done by Jannik Vogel:
- JayFoxRox/xqemu-espes@c05d91b
- JayFoxRox/xqemu-espes@a254830

This figure illustrates how (inclusive) clipping is performed with this PR, and how clipping region coordinates are transformed from D3D/NV2A coordinates to the OpenGL viewport coordinate system:

image

**Note:** this PR does not add complete clipping support. It will `assert` if "exclusive" clipping is used, and seems to have an off-by-1 issue most evident in Halo:CE with a new 1-pixel border along the top and right edges of the screen. However, these issues can be resolved after the PR is merged.

[Smiley face image source](https://commons.wikimedia.org/wiki/File:Mr._Smiley_Face.svg)

### DIFF
--- a/hw/xbox/nv2a/nv2a_pgraph.c
+++ b/hw/xbox/nv2a/nv2a_pgraph.c
@@ -874,6 +874,20 @@ static void pgraph_method(NV2AState *d,
         SET_MASK(pg->regs[NV_PGRAPH_FOGCOLOR], NV_PGRAPH_FOGCOLOR_ALPHA, alpha);
         break;
     }
+    case NV097_SET_WINDOW_CLIP_TYPE:
+        SET_MASK(pg->regs[NV_PGRAPH_SETUPRASTER],
+                 NV_PGRAPH_SETUPRASTER_WINDOWCLIPTYPE, parameter);
+        break;
+    case NV097_SET_WINDOW_CLIP_HORIZONTAL ...
+            NV097_SET_WINDOW_CLIP_HORIZONTAL + 0x1c:
+        slot = (method - NV097_SET_WINDOW_CLIP_HORIZONTAL) / 4;
+        pg->regs[NV_PGRAPH_WINDOWCLIPX0 + slot * 4] = parameter;
+        break;
+    case NV097_SET_WINDOW_CLIP_VERTICAL ...
+            NV097_SET_WINDOW_CLIP_VERTICAL + 0x1c:
+        slot = (method - NV097_SET_WINDOW_CLIP_VERTICAL) / 4;
+        pg->regs[NV_PGRAPH_WINDOWCLIPY0 + slot * 4] = parameter;
+        break;
     case NV097_SET_ALPHA_TEST_ENABLE:
         SET_MASK(pg->regs[NV_PGRAPH_CONTROL_0],
                  NV_PGRAPH_CONTROL_0_ALPHATESTENABLE, parameter);
@@ -2409,6 +2423,8 @@ static void pgraph_method(NV2AState *d,
         /* FIXME: Should this really be inverted instead of ymin? */
         glScissor(scissor_x, scissor_y, scissor_width, scissor_height);
 
+        /* FIXME: Respect window clip?!?! */
+
         NV2A_DPRINTF("------------------CLEAR 0x%x %d,%d - %d,%d  %x---------------\n",
             parameter, xmin, ymin, xmax, ymax, d->pgraph.regs[NV_PGRAPH_COLORCLEARVALUE]);
 
@@ -2917,7 +2933,6 @@ static void pgraph_shader_update_constants(PGRAPHState *pg,
     if (binding->clip_range_loc != -1) {
         glUniform2f(binding->clip_range_loc, zclip_min, zclip_max);
     }
-
 }
 
 static void pgraph_bind_shaders(PGRAPHState *pg)
@@ -2942,6 +2957,8 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
     ShaderState state = {
         .psh = (PshState){
             /* register combier stuff */
+            .window_clip_exclusive = pg->regs[NV_PGRAPH_SETUPRASTER]
+                                       & NV_PGRAPH_SETUPRASTER_WINDOWCLIPTYPE,
             .combiner_control = pg->regs[NV_PGRAPH_COMBINECTL],
             .shader_stage_program = pg->regs[NV_PGRAPH_SHADERPROG],
             .other_stage_input = pg->regs[NV_PGRAPH_SHADERCTL],
@@ -3037,6 +3054,45 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
         }
     }
 
+    /* Window clip
+     *
+     * Optimization note: very quickly check to ignore any repeated or zero-size
+     * clipping regions. Note that if region number 7 is valid, but the rest are
+     * not, we will still add all of them. Clip regions seem to be typically
+     * front-loaded (meaning the first one or two regions are populated, and the
+     * following are zeroed-out), so let's avoid adding any more complicated
+     * masking or copying logic here for now unless we discover a valid case.
+     */
+    assert(!state.psh.window_clip_exclusive); /* FIXME: Untested */
+    state.psh.window_clip_count = 0;
+    uint32_t last_x = 0, last_y = 0;
+
+    for (i = 0; i < 8; i++) {
+        const uint32_t x = pg->regs[NV_PGRAPH_WINDOWCLIPX0 + i * 4];
+        const uint32_t y = pg->regs[NV_PGRAPH_WINDOWCLIPY0 + i * 4];
+        const uint32_t x_min = GET_MASK(x, NV_PGRAPH_WINDOWCLIPX0_XMIN);
+        const uint32_t x_max = GET_MASK(x, NV_PGRAPH_WINDOWCLIPX0_XMAX);
+        const uint32_t y_min = GET_MASK(y, NV_PGRAPH_WINDOWCLIPY0_YMIN);
+        const uint32_t y_max = GET_MASK(y, NV_PGRAPH_WINDOWCLIPY0_YMAX);
+
+        /* Check for zero width or height clipping region */
+        if ((x_min == x_max) || (y_min == y_max)) {
+            continue;
+        }
+
+        /* Check for in-order duplicate regions */
+        if ((x == last_x) && (y == last_y)) {
+            continue;
+        }
+
+        NV2A_DPRINTF("Clipping Region %d: min=(%d, %d) max=(%d, %d)\n",
+            i, x_min, y_min, x_max, y_max);
+
+        state.psh.window_clip_count = i + 1;
+        last_x = x;
+        last_y = y;
+    }
+
     for (i = 0; i < 8; i++) {
         state.psh.rgb_inputs[i] = pg->regs[NV_PGRAPH_COMBINECOLORI0 + i * 4];
         state.psh.rgb_outputs[i] = pg->regs[NV_PGRAPH_COMBINECOLORO0 + i * 4];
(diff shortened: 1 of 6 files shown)


hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -3041,6 +3059,40 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
         }
     }
 
+    /* Window clip */
```
JayFoxRox: nit: should mention that this first part of the algorithm is mostly an optimization:

- Don't do anything if the entire window  [`all_x`, `all_y`] is clipped/unclipped
- Remove duplicated clipping regions to avoid useless logic in fragment shader
    - D3D clones the last active clip region into all following slots (if I remember correctly)

The code should still work without this (just `state.psh.window_clip_count = 8;`).
Wether I was going to keep this optimization was still left to decide I believe (due to added code complexity / likely little performance gain).

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -3087,6 +3139,28 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
...
+        GLuint xmin = GET_MASK(x, NV_PGRAPH_WINDOWCLIPX0_XMIN);
+        GLuint xmax = GET_MASK(x, NV_PGRAPH_WINDOWCLIPX0_XMAX);
+        uint32_t y = pg->regs[NV_PGRAPH_WINDOWCLIPY0 + i * 4];
+        GLuint ymin = pg->surface_shape.clip_height - 1 -
+                      GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMIN);
+        GLuint ymax = pg->surface_shape.clip_height - 1 -
+                      GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMAX);
+
+        snprintf(name, sizeof(name), "clipRegion[%d]", i);
+        GLint loc = glGetUniformLocation(pg->shader_binding->gl_program, name);
```
JayFoxRox: This uniform location should be cached, as done for other shader uniforms (I believe?).
This can have quite a large performance impact.

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -3041,6 +3059,40 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
...
+        if ((x == all_x) && (y == all_y)) {
+            state.psh.window_clip_count = 0;
+            break;
+        }
+        if (x != last_x || y != last_y) {
+            state.psh.window_clip_count = i + 1;
+            last_x = x;
+            last_y = y;
+        }
+    }
```
JayFoxRox: nit: Please review this optimization logic. I'm a little worried about it to be honest.
In particular the `all_x`, `all_y` part. the `last_x` / `last_y` should work fine.

I remember this stuff making my brain hurt (especially as window clip is not really documented anywhere and it's hard to find good tests)

hw/xbox/nv2a/nv2a_pgraph.c
mborgerson: Done

hw/xbox/nv2a/nv2a_pgraph.c
mborgerson: I think comment about this being an optimization is already pretty clear. Here's what the code says:

    /* Check if the last region is duplicated and ignore duplicates. We'll also
     * early out on regions which cover the entire surface.
     */

To be this is pretty clear, but I can emphasize that this is an optimization by us to detect any blatantly overlapping regions.

hw/xbox/nv2a/nv2a_pgraph.c
mborgerson: I'm pretty sure the logic is fine and fast enough. It might be fine just to let the shader do all 8 clipping regions. This definitely won't simplify everything, but it'll catch duplicates at least.

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -3083,6 +3134,23 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
 
     glUseProgram(pg->shader_binding->gl_program);
 
+    /* Clipping regions */
+    for (i = 0; i < state.psh.window_clip_count; i++) {
```
mborgerson: Seems to be adding an artifact (in Halo:CE at least, where we are clipping a 1-pixel line around the top and right of the screen. Likely an off-by-1 problem somewhere here.

hw/xbox/nv2a/nv2a_pgraph.c
JayFoxRox: Yeah, that'd be good. I think it's not clear that this code would still work fine if that block was removed.

Maybe also mention that we are not sure if it helps with performance?
If future-us will refactor this we might want to get rid of it to reduce complexity at 0-cost.

Wether you do this change is up to you - it's a minor doc issue.

hw/xbox/nv2a/nv2a_pgraph.c
JayFoxRox: The results in Halo imply that `max` bounds are inclusive. So `(0,0)-(0,0)` would still clip a single pixel. This needs more testing on real hardware though (also with anti-aliasing). We can't test in XQEMU because our pixel centers are wrong (I think), so we get off-by-half and off-by-one errors anyway.
It's up to you how you implement that case, we should have an issue about this ("Clarify clip region bounds" or something).

I believe the idea behind the `all` was to be able to avoid some draw call altogethers, and avoid *any* use of `discard`, because that could mean taking a slow path in the rasterizer of some older GPUs. It's probably neglible though, so I'm fine with this change; especially because it mostly adds complexity.

The  `last_x` and `last_y` should probably remain. I believe Dead or Alive 3 repeated clip regions. Not sure if this has any impact, but it's such a simple optimzation that it can be kept - probably more as an inspiration for better optimizations or if we move the clipping to the VSH (clipplanes or something).
But I'll also leave this up to you.

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -3083,6 +3135,28 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
...
+                      GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMIN);
+        GLuint ymax = pg->surface_shape.clip_height - 1 -
+                      GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMAX);
+
+        snprintf(name, sizeof(name), "clipRegion[%d]", i);
+        GLint loc = glGetUniformLocation(pg->shader_binding->gl_program, name);
+        if (loc != -1) {
+            pgraph_apply_anti_aliasing_factor(pg, &xmin, &ymin);
+            pgraph_apply_anti_aliasing_factor(pg, &xmax, &ymax);
+            glUniform4i(loc, xmin, ymax, xmax, ymin);
```
JayFoxRox: The order of these parameters is quite confusing, but it is because D3D and GL use different coordinate systems.
Are you going to add a commnt?

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -3083,6 +3135,28 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
...
+        char name[32];
+
+        uint32_t x = pg->regs[NV_PGRAPH_WINDOWCLIPX0 + i * 4];
+        GLuint xmin = GET_MASK(x, NV_PGRAPH_WINDOWCLIPX0_XMIN);
+        GLuint xmax = GET_MASK(x, NV_PGRAPH_WINDOWCLIPX0_XMAX);
+        uint32_t y = pg->regs[NV_PGRAPH_WINDOWCLIPY0 + i * 4];
+        GLuint ymin = pg->surface_shape.clip_height - 1 -
+                      GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMIN);
+        GLuint ymax = pg->surface_shape.clip_height - 1 -
+                      GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMAX);
```
JayFoxRox: nit: These inversions are necessary for GL / D3D differences. I think it would be better to do these in 2 steps:

```c
GLuint ymin = GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMIN);
GLuint ymax = GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMAX);

// <comment about flipping coordinate system upside down as we'll use GL coordinates in shader, which work different from D3D>
ymin = pg->surface_shape.clip_height - 1 - ymin;
ymax = pg->surface_shape.clip_height - 1 - ymax;
```

JayFoxRox:
Left comments about what we discussed via IM recently.
I don't want to bikeshed this either - so I'll wait for your response, test this myself, review one last time and then merge this.

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -3083,6 +3134,23 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
...
+        if (pg->shader_binding->clip_region_loc[i] != -1) {
+            pgraph_apply_anti_aliasing_factor(pg, &xmin, &ymin);
+            pgraph_apply_anti_aliasing_factor(pg, &xmax, &ymax);
+            glUniform4i(pg->shader_binding->clip_region_loc[i], xmin, ymax, xmax, ymin);
```
JayFoxRox: The order of these parameters is confusing, but it is because D3D and GL use different coordinate systems.
Are you going to add a comment?

mborgerson:
@JayFoxRox I'm not going to be able to work on this until this tor a few days. You're welcome to fix it up as we discussed and merge it in if you'd like, or I'll try to get to it this weekend.

hw/xbox/nv2a/nv2a_pgraph.c
PatrickvL: What's strange about this remark, it's that in most cases NV2A hardware is closely mapped to OpenGL. So why are the clipping coordinates following the Direct3D coordinate system all of a sudden?

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -3037,6 +3054,45 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
...
+    state.psh.window_clip_count = 0;
+    uint32_t last_x = 0, last_y = 0;
+
+    for (i = 0; i < 8; i++) {
+        const uint32_t x = pg->regs[NV_PGRAPH_WINDOWCLIPX0 + i * 4];
+        const uint32_t y = pg->regs[NV_PGRAPH_WINDOWCLIPY0 + i * 4];
+        const uint32_t x_min = GET_MASK(x, NV_PGRAPH_WINDOWCLIPX0_XMIN);
+        const uint32_t x_max = GET_MASK(x, NV_PGRAPH_WINDOWCLIPX0_XMAX);
+        const uint32_t y_min = GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMIN);
+        const uint32_t y_max = GET_MASK(y, NV_PGRAPH_WINDOWCLIPX0_XMAX);
```
JayFoxRox: nit: maybe add `NV_PGRAPH_WINDOWCLIPY0_YMAX` (and `_YMIN`) after all. I didn't add those because they are the same as the ones for `X` (and we always tried to keep register defines short).
But when reading over the code, it's still messing with my head

mborgerson: Sure

JayFoxRox:
Looks good; thanks for the changes.
I'll try the code this week and then I'll merge if it works.

Issue #50, opened 2018-06-26
Add lazy surfaces (aka surface cache)


mborgerson:
One major performance hinderance is time spent copying between RAM and graphics memory (glReadPixels). Ideally, we should avoid this if at all possible.

PR #55, opened 2018-06-28
nv2a: Recycle FIFO command queue memory

### LINKED ISSUE #54: Eliminate calls to malloc in pfifo_run_pusher
Typical frame of Halo:CE multiplayer calls `g_malloc0` ~53,419 times! Experimental results suggest noticeable performance improvement.

mborgerson:
This patch adds an additional "retired" queue in which FIFO command
entry objects are placed after execution. This queue of objects is then
returned to the pusher's new "available" queue for re-use.

This improves the performance of the system by avoiding the costly
overhead associated with the general-purpose use of `malloc` and `free`
for previous allocation of FIFO command queue objects.

Resolves #54, but TBH performance improvement is *visibly* minor, compared to our other slowdowns (#50).

I think there is still a lot of room to optimize this puller/pusher setup. But this is pretty low-hanging fruit and provides good returns, so let's leave more architectural changes for another PR.

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -507,10 +507,21 @@ static void nv2a_realize(PCIDevice *dev, Error **errp)
     }
 
     /* init fifo cache1 */
+    qemu_spin_init(&d->pfifo.cache1.alloc_lock);
     qemu_mutex_init(&d->pfifo.cache1.cache_lock);
     qemu_cond_init(&d->pfifo.cache1.cache_cond);
     QSIMPLEQ_INIT(&d->pfifo.cache1.cache);
     QSIMPLEQ_INIT(&d->pfifo.cache1.working_cache);
+    QSIMPLEQ_INIT(&d->pfifo.cache1.available_entries);
+    QSIMPLEQ_INIT(&d->pfifo.cache1.retired_entries);
+
+    /* Pre-allocate memory for CacheEntry objects */
+    for (i=0; i < 100000; i++) {
+        CacheEntry *command = g_malloc0(sizeof(CacheEntry));
+        assert(command != NULL);
+        QSIMPLEQ_INSERT_TAIL(&d->pfifo.cache1.available_entries,
+                             command, entry);
+    }
 }
 
 static void nv2a_exitfn(PCIDevice *dev)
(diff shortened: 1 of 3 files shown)


hw/xbox/nv2a/nv2a.c
```diff
@@ -507,10 +507,21 @@ static void nv2a_realize(PCIDevice *dev, Error **errp)
...
+    qemu_spin_init(&d->pfifo.cache1.alloc_lock);
     qemu_mutex_init(&d->pfifo.cache1.cache_lock);
     qemu_cond_init(&d->pfifo.cache1.cache_cond);
     QSIMPLEQ_INIT(&d->pfifo.cache1.cache);
     QSIMPLEQ_INIT(&d->pfifo.cache1.working_cache);
+    QSIMPLEQ_INIT(&d->pfifo.cache1.available_entries);
+    QSIMPLEQ_INIT(&d->pfifo.cache1.retired_entries);
+
+    /* Pre-allocate memory for CacheEntry objects */
+    for (i=0; i < 100000; i++) {
```
mborgerson: Possibly overkill but the structure is small, and a frame in MP of Halo:CE uses ~50k.

mborgerson:
I think there is still a lot of room to optimize this puller/pusher setup. But this is pretty low-hanging fruit and provides good returns, so let's leave more architectural changes for another PR.

hw/xbox/nv2a/nv2a.c
JayFoxRox: For what it's worse, the real Xbox has CACHE1 which holds 128 elements (method + data pair).
This actually important, because that means a method causing an interrupt being hit guarantees that a command in 128+ methods, has not been loaded yet, and can still be modified.
If I remember correctly, we had cases where the games were unable to inject further commands, because XQEMU pretended to have loaded *all* of them already.

I think our bottleneck would be 0x1800 methods which are used during draw calls (easily a couple of thousand per draw), so our cache has to be fast - and I'm not sure how fast it is right now.

If we do this properly, we'd even use the CACHE1 registers (array) in PGRAPH for this, no longer a list (how does QEMU even implement it?).
However, the CACHE1 behaves.. strangely. I've been experimenting with it for [nv2a-trace](https://github.com/XboxDev/nv2a-trace), and I couldn't really figure out how it works (the data looked good, but it's PUT and GET pointers were... odd). Also issues where the entire CACHE1 has the same method+data pair 128 times suddenly.
More research necessary (or doc reading?!).

Issue #53, opened 2018-06-28
"#error gl.h included before glew.h" during compilation


JayFoxRox:
Running MESA on Arch Linux I get the error which was described in https://github.com/xqemu/xqemu/pull/13 .
epoxy has not been integrated yet.

As master is currently broken, this issue should be fixed immediately.
I'll mark it as a regression, because XQEMU compilation used to work, now it doesn't.
Furthermore regression shows the importance of this issue.

Fetched from origin and checkout out current master (7f7dc95761dae0c67975b4a40bb2e246b20dcf1e)
Then ran build_linux.sh:

```
[fox@x230 xqemu2]$ ./build_linux.sh 
+ set -o pipefail
+ ./configure --enable-debug '--extra-cflags=-march=native -g -O0 -Wno-error=redundant-decls -Wno-error=unused-but-set-variable -DXBOX=1' --disable-werror --target-list=i386-softmmu --enable-sdl --enable-kvm --disable-xen --with-sdlabi=2.0 --disable-curl --disable-vnc --disable-docs --disable-tools --disable-guest-agent --disable-tpm --disable-live-block-migration --disable-replication --disable-capstone --disable-fdt --disable-libiscsi --disable-spice --disable-user

...

make: *** [Makefile:478: subdir-i386-softmmu] Error 2
```

Re-run to get clean error log:

```
[fox@x230 xqemu2]$ make
  CC      i386-softmmu/gdbstub-xml.o
  CC      i386-softmmu/hw/xbox/nv2a/nv2a.o
In file included from /home/fox/Data/Projects/xqemu2/hw/xbox/nv2a/gl/gloffscreen.h:44,
                 from /home/fox/Data/Projects/xqemu2/hw/xbox/nv2a/nv2a_shaders.h:25,
                 from /home/fox/Data/Projects/xqemu2/hw/xbox/nv2a/nv2a.h:38,
                 from /home/fox/Data/Projects/xqemu2/hw/xbox/nv2a/nv2a.c:27:
/usr/include/GL/glew.h:85:2: error: #error gl.h included before glew.h
 #error gl.h included before glew.h
  ^~~~~
/usr/include/GL/glew.h:97:2: error: #error glext.h included before glew.h
 #error glext.h included before glew.h
  ^~~~~
In file included from /home/fox/Data/Projects/xqemu2/hw/xbox/nv2a/gl/gloffscreen.h:44,
[...]

Issue #20, opened 2018-06-17
Run 2.x rebase on linux


illwieckz:
If you build xqemu 2.x rebase successfully and get that error at runtime:

```
No provider of glFrameTerminatorGREMEDY found.  Requires one of: GL extension "GL_GREMEDY_frame_terminator
```

Then you should try to comment out the `glFrameTerminatorGREMEDY(); ` line there and rebuild:
https://github.com/xqemu/xqemu/blob/a57c1db2c1065d8bbd313503e2522e10b0c941f2/hw/xbox/nv2a/nv2a_pgraph.c#L651-L653

Commenting out this line is not a proper fix btw, xqemu should not crash if this extension is missing, but it's enough to get xqemu working.

mborgerson:
Confirmed. Need to use proper extension handling query.

PR #56, opened 2018-06-28
Unify build scripts

MayeulC:
Hopefully this way we won't have to propagate build script changes across 3 scripts.
I've only tested it on Linux, but shouldn't be problematic to merge if the CI passes.


mborgerson:
Awesome, I was planning on doing this. Thanks for beating me to it! I'll merge it once the CI passes.

build.sh
```diff
@@ -0,0 +1,67 @@
...
+        ;;
+esac
+
+set -x # Print commands from now on
+
+./configure \
+	--enable-debug \
+	--extra-cflags="$CFLAGS_COMMON $CFLAGS" \
+	$CONFIGURE \
+    --target-list=i386-softmmu \
```
JayFoxRox: nit: weird indentation?

JayFoxRox:
Tested on Linux and Windows, rebased on #61 .

Code looks excellent and works like a charm on both platforms. Thanks!
Tiny nit aside from previous feedback: comment spelling goes back and forth between "# Starts uppercase" and "# starts lowercase", please try to be consistent.


LGTM

*(Waiting for regression fix on #53 and #20 (Solved by #61) before merging this one. Sorry for the inconvience*)

PR #57, opened 2018-06-28
Window title "XQEMU [git hash]"

### LINKED ISSUE #47: Git commit id is hard to find
Suggest adding it to title bar of XQEMU window

### LINKED ISSUE #49: Main window should show "XQEMU" in title bar and XQEMU icon
Need to update the main window titlebar with XQEMU and the XQEMU icon (tbd)

dracc:
Closes #49 

  

Screenshot of dashboard with "new" titlebar

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -442,7 +442,7 @@ qemu-version.h: FORCE
 			pkgvers="$(PKGVERSION)"; \
 		else \
 			if test -d .git; then \
-				pkgvers=$$(git describe --match 'v*' 2>/dev/null | tr -d '\n');\
+				pkgvers=$$(git rev-parse --short HEAD 2>/dev/null | tr -d '\n');\
 				if ! git diff-index --quiet HEAD &>/dev/null; then \
 					pkgvers="$${pkgvers}-dirty"; \
 				fi; \
--- a/ui/sdl2.c
+++ b/ui/sdl2.c
@@ -25,6 +25,7 @@
 
 #include "qemu/osdep.h"
 #include "qemu-common.h"
+#include "qemu-version.h"
 #include "ui/console.h"
 #include "ui/input.h"
 #include "ui/sdl2.h"
@@ -135,6 +136,8 @@ static void sdl_update_caption(struct sdl2_console *scon)
     char win_title[1024];
     char icon_title[1024];
     const char *status = "";
+    const char *project_name = "XQEMU";
+    const char *hash_indicator = "Revision";
 
     if (!runstate_is_running()) {
         status = " [Stopped]";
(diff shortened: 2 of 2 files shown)


ui/sdl2.c
```diff
@@ -149,12 +182,12 @@ static void sdl_update_caption(struct sdl2_console *scon)
...
-        snprintf(win_title, sizeof(win_title), "QEMU (%s-%d)%s", qemu_name,
+        snprintf(win_title, sizeof(win_title), "XQEMU (%s) (%s-%d)%s", QEMU_PKGVERSION, qemu_name,
                  scon->idx, status);
-        snprintf(icon_title, sizeof(icon_title), "QEMU (%s)", qemu_name);
+        snprintf(icon_title, sizeof(icon_title), "XQEMU (%s) (%s)", QEMU_PKGVERSION, qemu_name);
     } else {
-        snprintf(win_title, sizeof(win_title), "QEMU%s", status);
-        snprintf(icon_title, sizeof(icon_title), "QEMU");
+        snprintf(win_title, sizeof(win_title), "XQEMU (%s)%s", QEMU_PKGVERSION, status);
+        snprintf(icon_title, sizeof(icon_title), "XQEMU (%s)", QEMU_PKGVERSION);
```
JayFoxRox: nit: Yikes, so many variations..

I suggest fixing this "QEMU" mess by adding a seperate variable name `project_title` or something.

Just out of interest: Do you know what these variations do?

Makefile
```diff
@@ -442,7 +442,7 @@ qemu-version.h: FORCE
 			pkgvers="$(PKGVERSION)"; \
 		else \
 			if test -d .git; then \
-				pkgvers=$$(git describe --match 'v*' 2>/dev/null | tr -d '\n');\
+				pkgvers=$$(git rev-parse --short HEAD 2>/dev/null | tr -d '\n');\
```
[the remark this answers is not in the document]
dracc: It switches the version string from `[release]-[tag]-[commit]` to just the commit hash.
I only managed to get `-dirty` before changing it as my local branch didn't have any tags. As far as I've noticed, nothing has broken because of that change.

ui/sdl2.c
dracc: Two pairs for title bar and window switcher, with and without `name` as provided by the `-name` flag during startup.

dracc:
Re @JayFoxRox;
* Would `(Version: [commit])` suffice?
* I'm sorry about pulling the gun on the logo, needed something to put there. It was meant to be highly temporary - I'll remove the screenshot if it doesn't belong. Cyan is probably because of compression, the array describes a 110*110 pixel image and my window manager screwed up the colors.
* I have no clue how QEMU solved that part - in fact I have only ever seen my default window icon when running (X)QEMU. The image array was created from an .html-table-image generated by GIMP. I'll provide a quick guide if wanted/needed.

JayFoxRox:
Suggestions: 

* Use "Revision" instead of "Version"; that way we can always ask for revision (which is the proper term and same for developers and users) + we keep "Version" available for stable releases.
* Remove the logo portion of this PR (but hold on to the code, possibly improve it for later)

The logo can be integrated [once the poll closes](https://github.com/xqemu/xqemu.com/issues/1#issuecomment-401174766), and we can quickly get the critical part of this merged (which is revision information in screenshots / movies).

*Due to importance, it might be possible that we'll wait for a fix for #53 , so don't be surprised if we don't immediately merge.*

mborgerson:
Thanks! I've tested this patch against the latest master and it works as expected.

One nit pick that I'd like you to fix before I merge this though is the Git commit message. Currently it reads:

    Change window name to "XQEMU", print git commit short-hash.

I'd prefer you `git commit --amend` it to something like:

    ui: Set window title to 'XQEMU (Revision: <id>)'

This alternative message is shorter, does not include a trailing period, includes subsystem prefix, and fixes ambiguous part that says the commit short-hash is printed.

dracc:
@mborgerson I'll do that as soon as I get home from work. Thanks for your input!

Makefile
```diff
@@ -442,7 +442,7 @@ qemu-version.h: FORCE
...
 				if ! git diff-index --quiet HEAD &>/dev/null; then \
```
dracc: This command never runs cleanly for me, I get the help message because syntax is wrong.
Can anyone else confirm?

I've also checked with [upstream](https://git.qemu.org/?p=qemu.git;a=blob;f=Makefile;h=7ed9cc4a211aef25e30bb4b2354308d2c18a46c3;hb=HEAD#l449), same command there.

dracc: I can not reproduce this anymore. Must have been a bug stemming from a software upgrade or something.

PR #72, opened 2018-07-09
Perform some cleanup on Xbox code

### LINKED ISSUE #70: Remove "HACK HACK HACK" code
The nv2a_pgraph code contains some garbage in NV097_FLIP_STALL, that is in an `#if 0` block marked with `// HACK HACK HACK`:

https://github.com/xqemu/xqemu/blob/a9db47323103300847864a062e14e3abd755d19b/hw/xbox/nv2a/nv2a_pgraph.c#L659

@mborgerson [said this](https://github.com/xqemu/xqemu/pull/69#issuecomment-403312689):

> That `#if 0 // HACK HACK HACK` garbage needs to be pulled out of the source. That was added by me during my refactoring work and inadvertently got merged back into the project during my rebase work.

mborgerson:
This PR is to cleanup some of the Xbox-specific code, specifically.
- Removing dead code (Resolves #70)
- Eliminating some duplication through use of basic macros
- Ideally get all code passing through scripts/checkpatch.pl (I fixed some files during rebase, but there are more...)
- Additional cleanup tbd

## checkpatch.pl
Note: I'm not fixing up stuff we get from other places (e.g. wlgext.h), just our core Xbox stuff.

Many of our files are coming back clean now from checkpatch.pl. Still need to fixup:
- nv2a_pfifo.c
- nv2a_pgraph.c
- nv2a_shaders.c

checkpatch.pl will still complain about our usage of:
- `ERROR: use ctz32() instead of ffs()`
- C99 comment style (`// comment`)
- `#if 0`/`#endif` disabled code
- Lines with more than 80 characters

I've been using clang-format to do the extremely tedious bits, but it doesn't always get things right so I've had to selectively add the changes.

### DIFF
--- a/hw/xbox/lpc47m157.c
+++ b/hw/xbox/lpc47m157.c
@@ -47,7 +47,12 @@
 #define CONFIG_DEVICE_BASE_ADDRESS_LOW      0x61
 #define CONFIG_DEVICE_INETRRUPT             0x70
 
-#define DEBUG_LPC47M157
+// #define DEBUG
+#ifdef DEBUG
+# define DPRINTF(format, ...) printf(format, ## __VA_ARGS__)
+#else
+# define DPRINTF(format, ...) do { } while (0)
+#endif
 
 typedef struct LPC47M157State {
     ISADevice dev;
@@ -72,13 +77,12 @@ typedef struct LPC47M157State {
 static void update_devices(LPC47M157State *s)
 {
     ISADevice *isadev = ISA_DEVICE(s);
+    int i;
 
     /* init serial devices */
-    int i;
     for (i = 0; i < 2; i++) {
         uint8_t *dev = s->device_regs[DEVICE_SERIAL_PORT_1 + i];
         if (dev[CONFIG_DEVICE_ACTIVATE] && !s->serial[i].active) {
-
             uint32_t iobase = (dev[CONFIG_DEVICE_BASE_ADDRESS_HIGH] << 8)
                                 | dev[CONFIG_DEVICE_BASE_ADDRESS_LOW];
             uint32_t irq = dev[CONFIG_DEVICE_INETRRUPT];
@@ -99,9 +103,7 @@ static void lpc47m157_io_write(void *opaque, hwaddr addr, uint64_t val,
 {
     LPC47M157State *s = opaque;
 
-#ifdef DEBUG_LPC47M157
-    printf("lpc47m157 io write 0x%"HWADDR_PRIx" = 0x%"PRIx64"\n", addr, val);
-#endif
+    DPRINTF("lpc47m157 io write 0x%" HWADDR_PRIx " = 0x%" PRIx64 "\n", addr, val);
 
     if (addr == 0) {
         /* INDEX_PORT */
@@ -126,11 +128,9 @@ static void lpc47m157_io_write(void *opaque, hwaddr addr, uint64_t val,
             assert(s->config_regs[CONFIG_DEVICE_NUMBER] < MAX_DEVICE);
             uint8_t *dev = s->device_regs[s->config_regs[CONFIG_DEVICE_NUMBER]];
             dev[s->selected_reg] = val;
-#ifdef DEBUG_LPC47M157
-            printf("lpc47m157 dev %x . %x = %"PRIx64"\n",
+            DPRINTF("lpc47m157 dev %x . %x = %"PRIx64"\n",
                 s->config_regs[CONFIG_DEVICE_NUMBER],
                 s->selected_reg, val);
-#endif
         }
     } else {
         assert(false);
@@ -157,10 +157,7 @@ static uint64_t lpc47m157_io_read(void *opaque, hwaddr addr, unsigned int size)
         assert(false);
     }
 
-#ifdef DEBUG_LPC47M157
-    printf("lpc47m157 io read 0x%"HWADDR_PRIx" -> 0x%x\n", addr, val);
-#endif
-
+    DPRINTF("lpc47m157 io read 0x%"HWADDR_PRIx" -> 0x%x\n", addr, val);
     return val;
 }
 
@@ -183,6 +180,7 @@ static void lpc47m157_realize(DeviceState *dev, Error **errp)
 {
     LPC47M157State *s = LPC47M157_DEVICE(dev);
     ISADevice *isa = ISA_DEVICE(dev);
+    int i;
 
     const uint32_t iobase = 0x2e; //0x4e if SYSOPT pin, make it a property
     s->config_regs[CONFIG_PORT_LOW] = iobase & 0xFF;
@@ -193,7 +191,6 @@ static void lpc47m157_realize(DeviceState *dev, Error **errp)
     isa_register_ioport(isa, &s->io, iobase);
 
     /* init serial cores */
-    int i;
     for (i = 0; i < 2; i++) {
         Chardev *chr = serial_hds[i];
         if (chr == NULL) {
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -20,12 +20,11 @@
  */
 
 #include "qemu/osdep.h"
+#include "hw/display/vga_regs.h"
 #include "qapi/error.h"
 #include "qemu/error-report.h"
-#include "qemu/error-report.h"
-#include <assert.h>
 #include "nv2a.h"
-#include "hw/display/vga_regs.h"
+#include <assert.h>
 
 #ifdef __WINNT__
 // HACK: mingw-w64 doesn't provide ffs, for now we just shove it here
@@ -34,19 +33,17 @@ int ffs(register int valu)
 {
     register int bit;
 
-    if (valu == 0)
+    if (valu == 0) {
         return 0;
+    }
 
-    for (bit = 1; !(valu & 1); bit++)
+    for (bit = 1; !(valu & 1); bit++) {
         valu >>= 1;
+    }
 
     return bit;
 }
 #endif
-
-DMAObject nv_dma_load(NV2AState *d, hwaddr dma_obj_address);
-void *nv_dma_map(NV2AState *d, hwaddr dma_obj_address, hwaddr *len);
-void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
 
 void update_irq(NV2AState *d)
 {
@@ -83,16 +80,16 @@ DMAObject nv_dma_load(NV2AState *d, hwaddr dma_obj_address)
 {
     assert(dma_obj_address < memory_region_size(&d->ramin));
 
-    uint32_t *dma_obj = (uint32_t*)(d->ramin_ptr + dma_obj_address);
+    uint32_t *dma_obj = (uint32_t *)(d->ramin_ptr + dma_obj_address);
     uint32_t flags = ldl_le_p(dma_obj);
     uint32_t limit = ldl_le_p(dma_obj + 1);
     uint32_t frame = ldl_le_p(dma_obj + 2);
 
     return (DMAObject){
-        .dma_class = GET_MASK(flags, NV_DMA_CLASS),
+        .dma_class  = GET_MASK(flags, NV_DMA_CLASS),
         .dma_target = GET_MASK(flags, NV_DMA_TARGET),
-        .address = (frame & NV_DMA_ADDRESS) | GET_MASK(flags, NV_DMA_ADJUST),
-        .limit = limit,
+        .address    = (frame & NV_DMA_ADDRESS) | GET_MASK(flags, NV_DMA_ADJUST),
+        .limit      = limit,
     };
 }
 
@@ -113,38 +110,11 @@ void *nv_dma_map(NV2AState *d, hwaddr dma_obj_address, hwaddr *len)
     return d->vram_ptr + dma.address;
 }
 
-#define DEFINE_PROTO(prefix) \
-    uint64_t prefix ## _read(void *opaque, hwaddr addr, unsigned int size); \
-    void prefix ## _write(void *opaque, hwaddr addr, uint64_t val, unsigned int size);
-
-DEFINE_PROTO(pmc)
-DEFINE_PROTO(pbus)
-DEFINE_PROTO(pfifo)
-DEFINE_PROTO(prma)
-DEFINE_PROTO(pvideo)
-DEFINE_PROTO(ptimer)
-DEFINE_PROTO(pcounter)
-DEFINE_PROTO(pvpe)
-DEFINE_PROTO(ptv)
-DEFINE_PROTO(prmfb)
-DEFINE_PROTO(prmvio)
-DEFINE_PROTO(pfb)
-DEFINE_PROTO(pstraps)
-DEFINE_PROTO(pgraph)
-DEFINE_PROTO(pcrtc)
-DEFINE_PROTO(prmcio)
-DEFINE_PROTO(pramdac)
-DEFINE_PROTO(prmdio)
-// DEFINE_PROTO(pramin)
-DEFINE_PROTO(user)
-
-#undef DEFINE_PROTO
-
 #include "nv2a_pbus.c"
 #include "nv2a_pcrtc.c"
 #include "nv2a_pfb.c"
-#include "nv2a_pgraph.c"
 #include "nv2a_pfifo.c"
+#include "nv2a_pgraph.c"
 #include "nv2a_pmc.c"
 #include "nv2a_pramdac.c"
 #include "nv2a_prmcio.c"
@@ -154,42 +124,46 @@ DEFINE_PROTO(user)
 #include "nv2a_stubs.c"
 #include "nv2a_user.c"
 
+#define ENTRY(NAME, OFFSET, SIZE, RDFUNC, WRFUNC)      \
+    [NV_##NAME] = {                                    \
+        .name   = #NAME,                               \
+        .offset = OFFSET,                              \
+        .size   = SIZE,                                \
+        .ops    = { .read = RDFUNC, .write = WRFUNC }, \
+    }
+
 const struct NV2ABlockInfo blocktable[] = {
-
-#define ENTRY(NAME, OFFSET, SIZE, RDFUNC, WRFUNC) \
-    [ NV_ ## NAME ]  = { \
-        .name = #NAME, .offset = OFFSET, .size = SIZE, \
-        .ops = { .read = RDFUNC, .write = WRFUNC }, \
-    }, \
-
-    ENTRY(PMC,      0x000000, 0x001000, pmc_read,      pmc_write)
-    ENTRY(PBUS,     0x001000, 0x001000, pbus_read,     pbus_write)
-    ENTRY(PFIFO,    0x002000, 0x002000, pfifo_read,    pfifo_write)
-    ENTRY(PRMA,     0x007000, 0x001000, prma_read,     prma_write)
-    ENTRY(PVIDEO,   0x008000, 0x001000, pvideo_read,   pvideo_write)
-    ENTRY(PTIMER,   0x009000, 0x001000, ptimer_read,   ptimer_write)
-    ENTRY(PCOUNTER, 0x00a000, 0x001000, pcounter_read, pcounter_write)
-    ENTRY(PVPE,     0x00b000, 0x001000, pvpe_read,     pvpe_write)
-    ENTRY(PTV,      0x00d000, 0x001000, ptv_read,      ptv_write)
-    ENTRY(PRMFB,    0x0a0000, 0x020000, prmfb_read,    prmfb_write)
-    ENTRY(PRMVIO,   0x0c0000, 0x001000, prmvio_read,   prmvio_write)
-    ENTRY(PFB,      0x100000, 0x001000, pfb_read,      pfb_write)
-    ENTRY(PSTRAPS,  0x101000, 0x001000, pstraps_read,  pstraps_write)
-    ENTRY(PGRAPH,   0x400000, 0x002000, pgraph_read,   pgraph_write)
-    ENTRY(PCRTC,    0x600000, 0x001000, pcrtc_read,    pcrtc_write)
-    ENTRY(PRMCIO,   0x601000, 0x001000, prmcio_read,   prmcio_write)
-    ENTRY(PRAMDAC,  0x680000, 0x001000, pramdac_read,  pramdac_write)
-    ENTRY(PRMDIO,   0x681000, 0x001000, prmdio_read,   prmdio_write)
-    // ENTRY(PRAMIN,   0x700000, 0x100000, pramin_read,   pramin_write)
-    ENTRY(USER,     0x800000, 0x800000, user_read,     user_write)
-#undef ENTRY
+    ENTRY(PMC,      0x000000, 0x001000, pmc_read,      pmc_write),
+    ENTRY(PBUS,     0x001000, 0x001000, pbus_read,     pbus_write),
+    ENTRY(PFIFO,    0x002000, 0x002000, pfifo_read,    pfifo_write),
+    ENTRY(PRMA,     0x007000, 0x001000, prma_read,     prma_write),
+    ENTRY(PVIDEO,   0x008000, 0x001000, pvideo_read,   pvideo_write),
+    ENTRY(PTIMER,   0x009000, 0x001000, ptimer_read,   ptimer_write),
+    ENTRY(PCOUNTER, 0x00a000, 0x001000, pcounter_read, pcounter_write),
+    ENTRY(PVPE,     0x00b000, 0x001000, pvpe_read,     pvpe_write),
+    ENTRY(PTV,      0x00d000, 0x001000, ptv_read,      ptv_write),
+    ENTRY(PRMFB,    0x0a0000, 0x020000, prmfb_read,    prmfb_write),
+    ENTRY(PRMVIO,   0x0c0000, 0x001000, prmvio_read,   prmvio_write),
+    ENTRY(PFB,      0x100000, 0x001000, pfb_read,      pfb_write),
+    ENTRY(PSTRAPS,  0x101000, 0x001000, pstraps_read,  pstraps_write),
+    ENTRY(PGRAPH,   0x400000, 0x002000, pgraph_read,   pgraph_write),
+    ENTRY(PCRTC,    0x600000, 0x001000, pcrtc_read,    pcrtc_write),
+    ENTRY(PRMCIO,   0x601000, 0x001000, prmcio_read,   prmcio_write),
+    ENTRY(PRAMDAC,  0x680000, 0x001000, pramdac_read,  pramdac_write),
+    ENTRY(PRMDIO,   0x681000, 0x001000, prmdio_read,   prmdio_write),
+    // ENTRY(PRAMIN,   0x700000, 0x100000, pramin_read,   pramin_write),
+    ENTRY(USER,     0x800000, 0x800000, user_read,     user_write),
 };
+
+#undef ENTRY
 
 const int blocktable_len = ARRAY_SIZE(blocktable);
 
+// FIXME: Add nv2a_reg_names or remove this code
 // static const char* nv2a_reg_names[] = {};
 
-void reg_log_read(int block, hwaddr addr, uint64_t val) {
+void reg_log_read(int block, hwaddr addr, uint64_t val)
+{
     if (blocktable[block].name) {
         // hwaddr naddr = blocktable[block].offset + addr;
         // if (naddr < ARRAY_SIZE(nv2a_reg_names) && nv2a_reg_names[naddr]) {
@@ -197,15 +171,16 @@ void reg_log_read(int block, hwaddr addr, uint64_t val) {
         //             blocktable[block].name, nv2a_reg_names[naddr], val);
         // } else {
             NV2A_DPRINTF("%s: read [%" HWADDR_PRIx "] -> 0x%" PRIx64 "\n",
-                    blocktable[block].name, addr, val);
+                         blocktable[block].name, addr, val);
         // }
     } else {
         NV2A_DPRINTF("(%d?): read [%" HWADDR_PRIx "] -> 0x%" PRIx64 "\n",
-                block, addr, val);
+                     block, addr, val);
     }
 }
 
-void reg_log_write(int block, hwaddr addr, uint64_t val) {
+void reg_log_write(int block, hwaddr addr, uint64_t val)
+{
     if (blocktable[block].name) {
         // hwaddr naddr = blocktable[block].offset + addr;
         // if (naddr < ARRAY_SIZE(nv2a_reg_names) && nv2a_reg_names[naddr]) {
@@ -213,11 +188,11 @@ void reg_log_write(int block, hwaddr addr, uint64_t val) {
         //             blocktable[block].name, nv2a_reg_names[naddr], val);
         // } else {
             NV2A_DPRINTF("%s: [%" HWADDR_PRIx "] = 0x%" PRIx64 "\n",
-                    blocktable[block].name, addr, val);
+                         blocktable[block].name, addr, val);
         // }
     } else {
         NV2A_DPRINTF("(%d?): [%" HWADDR_PRIx "] = 0x%" PRIx64 "\n",
-                block, addr, val);
+                     block, addr, val);
     }
 }
 
@@ -346,8 +321,8 @@ static void nv2a_get_offsets(VGACommonState *s,
     *pstart_addr = start_addr;
 
     line_compare = s->cr[VGA_CRTC_LINE_COMPARE] |
-        ((s->cr[VGA_CRTC_OVERFLOW] & 0x10) << 4) |
-        ((s->cr[VGA_CRTC_MAX_SCAN] & 0x40) << 3);
+                   ((s->cr[VGA_CRTC_OVERFLOW] & 0x10) << 4) |
+                   ((s->cr[VGA_CRTC_MAX_SCAN] & 0x40) << 3);
     *pline_compare = line_compare;
 }
 
@@ -394,7 +369,6 @@ static void nv2a_init_memory(NV2AState *d, MemoryRegion *ram)
                              d->vram, 0, memory_region_size(d->vram));
     d->vga.vram_ptr = memory_region_get_ram_ptr(&d->vga.vram);
     vga_dirty_log_start(&d->vga);
-
 
     pgraph_init(d);
 
@@ -460,7 +434,7 @@ static void nv2a_realize(PCIDevice *dev, Error **errp)
     QSIMPLEQ_INIT(&d->pfifo.cache1.retired_entries);
 
     /* Pre-allocate memory for CacheEntry objects */
-    for (i=0; i < 100000; i++) {
+    for (i = 0; i < 100000; i++) {
         CacheEntry *command = g_malloc0(sizeof(CacheEntry));
         assert(command != NULL);
         QSIMPLEQ_INSERT_TAIL(&d->pfifo.cache1.available_entries,
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -135,7 +135,7 @@ DEFINE_PROTO(pcrtc)
 DEFINE_PROTO(prmcio)
 DEFINE_PROTO(pramdac)
 DEFINE_PROTO(prmdio)
-DEFINE_PROTO(pramin)
+// DEFINE_PROTO(pramin)
 DEFINE_PROTO(user)
 
 #undef DEFINE_PROTO
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -113,34 +113,6 @@ void *nv_dma_map(NV2AState *d, hwaddr dma_obj_address, hwaddr *len)
     return d->vram_ptr + dma.address;
 }
 
-#define STUB 0
-
-#if STUB
-void *pfifo_puller_thread(void *opaque) { return NULL; }
-void pgraph_init(NV2AState *d){}
-static void pfifo_run_pusher(NV2AState *d){}
-void pgraph_destroy(PGRAPHState *pg){}
-static uint8_t cliptobyte(int x)
-{
-    return (uint8_t)((x < 0) ? 0 : ((x > 255) ? 255 : x));
-}
-static void convert_yuy2_to_rgb(const uint8_t *line, unsigned int ix,
-                                uint8_t *r, uint8_t *g, uint8_t* b) {
-    int c, d, e;
-    c = (int)line[ix * 2] - 16;
-    if (ix % 2) {
-        d = (int)line[ix * 2 - 1] - 128;
-        e = (int)line[ix * 2 + 1] - 128;
-    } else {
-        d = (int)line[ix * 2 + 1] - 128;
-        e = (int)line[ix * 2 + 3] - 128;
-    }
-    *r = cliptobyte((298 * c + 409 * e + 128) >> 8);
-    *g = cliptobyte((298 * c - 100 * d - 208 * e + 128) >> 8);
-    *b = cliptobyte((298 * c + 516 * d + 128) >> 8);
-}
-#endif
-
 #define DEFINE_PROTO(prefix) \
     uint64_t prefix ## _read(void *opaque, hwaddr addr, unsigned int size); \
     void prefix ## _write(void *opaque, hwaddr addr, uint64_t val, unsigned int size);
@@ -171,10 +143,8 @@ DEFINE_PROTO(user)
 #include "nv2a_pbus.c"
 #include "nv2a_pcrtc.c"
 #include "nv2a_pfb.c"
-#if !STUB
 #include "nv2a_pgraph.c"
 #include "nv2a_pfifo.c"
-#endif
 #include "nv2a_pmc.c"
 #include "nv2a_pramdac.c"
 #include "nv2a_prmcio.c"
@@ -183,32 +153,6 @@ DEFINE_PROTO(user)
 #include "nv2a_pvideo.c"
 #include "nv2a_stubs.c"
 #include "nv2a_user.c"
-
-#if STUB
-void pgraph_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size)
-{
-    reg_log_write(NV_PGRAPH, addr, val);
-}
-
-uint64_t pgraph_read(void *opaque, 
-                                   hwaddr addr, unsigned int size)
-{
-    reg_log_read(NV_PGRAPH, addr, 0);
-    return 0;
-}
-
-void pfifo_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size)
-{
-    reg_log_write(NV_PFIFO, addr, val);
-}
-
-uint64_t pfifo_read(void *opaque, 
-                                   hwaddr addr, unsigned int size)
-{
-    reg_log_read(NV_PFIFO, addr, 0);
-    return 0;
-}
-#endif
 
 const struct NV2ABlockInfo blocktable[] = {
 
--- a/hw/xbox/nv2a/nv2a.h
+++ b/hw/xbox/nv2a/nv2a.h
@@ -23,44 +23,41 @@
 #define HW_NV2A_H
 
 #include "hw/hw.h"
-#include "hw/i386/pc.h"
-#include "ui/console.h"
-#include "hw/pci/pci.h"
-#include "ui/console.h"
 #include "hw/display/vga.h"
 #include "hw/display/vga_int.h"
-#include "qemu/thread.h"
+#include "hw/i386/pc.h"
+#include "hw/pci/pci.h"
 #include "qapi/qmp/qstring.h"
+#include "qemu/thread.h"
 #include "cpu.h"
 
 #include "g-lru-cache.h"
-#include "swizzle.h"
-#include "nv2a_shaders.h"
 #include "nv2a_debug.h"
 #include "nv2a_int.h"
-
+#include "nv2a_shaders.h"
+#include "swizzle.h"
 #include "gl/gloffscreen.h"
 
 #define USE_TEXTURE_CACHE
 
-#define GET_MASK(v, mask) (((v) & (mask)) >> (ffs(mask)-1))
+#define GET_MASK(v, mask) (((v) & (mask)) >> (ffs(mask) - 1))
 
-#define SET_MASK(v, mask, val) ({                                    \
-        const unsigned int __val = (val);                             \
-        const unsigned int __mask = (mask);                          \
-        (v) &= ~(__mask);                                            \
-        (v) |= ((__val) << (ffs(__mask)-1)) & (__mask);              \
+#define SET_MASK(v, mask, val)                            \
+    ({                                                    \
+        const unsigned int __val = (val);                 \
+        const unsigned int __mask = (mask);               \
+        (v) &= ~(__mask);                                 \
+        (v) |= ((__val) << (ffs(__mask) - 1)) & (__mask); \
     })
 
-#define CASE_4(v, step)                                              \
-    case (v):                                                        \
-    case (v)+(step):                                                 \
-    case (v)+(step)*2:                                               \
-    case (v)+(step)*3
+#define CASE_4(v, step)      \
+    case (v):                \
+    case ((v) + (step)):     \
+    case ((v) + (step) * 2): \
+    case ((v) + (step) * 3)
 
 
-#define NV2A_DEVICE(obj) \
-    OBJECT_CHECK(NV2AState, (obj), "nv2a")
+#define NV2A_DEVICE(obj) OBJECT_CHECK(NV2AState, (obj), "nv2a")
 
 void reg_log_read(int block, hwaddr addr, uint64_t val);
 void reg_log_write(int block, hwaddr addr, uint64_t val);
@@ -147,8 +144,8 @@ typedef struct TextureShape {
 typedef struct TextureKey {
     TextureShape state;
     uint64_t data_hash;
-    uint8_t* texture_data;
-    uint8_t* palette_data;
+    uint8_t *texture_data;
+    uint8_t *palette_data;
 } TextureKey;
 
 typedef struct TextureBinding {
@@ -248,7 +245,7 @@ typedef struct PGRAPHState {
     bool texture_matrix_enable[NV2A_MAX_TEXTURES];
 
     /* FIXME: Move to NV_PGRAPH_BUMPMAT... */
-    float bump_env_matrix[NV2A_MAX_TEXTURES-1][4]; /* 3 allowed stages with 2x2 matrix each */
+    float bump_env_matrix[NV2A_MAX_TEXTURES - 1][4]; /* 3 allowed stages with 2x2 matrix each */
 
     GloContext *gl_context;
     GLuint gl_framebuffer;
@@ -260,7 +257,7 @@ typedef struct PGRAPHState {
     bool zpass_pixel_count_enable;
     unsigned int zpass_pixel_count_result;
     unsigned int gl_zpass_pixel_count_query_count;
-    GLuint* gl_zpass_pixel_count_queries;
+    GLuint *gl_zpass_pixel_count_queries;
 
     hwaddr dma_vertex_a, dma_vertex_b;
 
@@ -310,7 +307,6 @@ typedef struct PGRAPHState {
 
     uint32_t regs[0x2000];
 } PGRAPHState;
-
 
 typedef struct CacheEntry {
     QSIMPLEQ_ENTRY(CacheEntry) entry;
@@ -431,7 +427,7 @@ typedef struct NV2AState {
 } NV2AState;
 
 typedef struct NV2ABlockInfo {
-    const char* name;
+    const char *name;
     hwaddr offset;
     uint64_t size;
     MemoryRegionOps ops;
@@ -440,9 +436,45 @@ typedef struct NV2ABlockInfo {
 extern const struct NV2ABlockInfo blocktable[];
 extern const int blocktable_len;
 
+DMAObject nv_dma_load(NV2AState *d, hwaddr dma_obj_address);
+void *nv_dma_map(NV2AState *d, hwaddr dma_obj_address, hwaddr *len);
+void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
 void pgraph_init(NV2AState *d);
 void *pfifo_puller_thread(void *opaque);
 void pgraph_destroy(PGRAPHState *pg);
 void update_irq(NV2AState *d);
+void pgraph_context_switch(NV2AState *d, unsigned int channel_id);
+void pgraph_wait_fifo_access(NV2AState *d);
+void pgraph_method(NV2AState *d,
+                   unsigned int subchannel,
+                   unsigned int method,
+                   uint32_t parameter);
+
+#define DEFINE_PROTO(n)                                              \
+    uint64_t n##_read(void *opaque, hwaddr addr, unsigned int size); \
+    void n##_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size);
+
+DEFINE_PROTO(pmc)
+DEFINE_PROTO(pbus)
+DEFINE_PROTO(pfifo)
+DEFINE_PROTO(prma)
+DEFINE_PROTO(pvideo)
+DEFINE_PROTO(ptimer)
+DEFINE_PROTO(pcounter)
+DEFINE_PROTO(pvpe)
+DEFINE_PROTO(ptv)
+DEFINE_PROTO(prmfb)
+DEFINE_PROTO(prmvio)
+DEFINE_PROTO(pfb)
+DEFINE_PROTO(pstraps)
+DEFINE_PROTO(pgraph)
+DEFINE_PROTO(pcrtc)
+DEFINE_PROTO(prmcio)
+DEFINE_PROTO(pramdac)
+DEFINE_PROTO(prmdio)
+// DEFINE_PROTO(pramin)
+DEFINE_PROTO(user)
+
+#undef DEFINE_PROTO
 
 #endif
--- a/hw/xbox/nv2a/nv2a_stubs.c
+++ b/hw/xbox/nv2a/nv2a_stubs.c
@@ -19,100 +19,29 @@
  * along with this program; if not, see <http://www.gnu.org/licenses/>.
  */
 
-uint64_t prma_read(void *opaque,
-                                  hwaddr addr, unsigned int size)
-{
-    reg_log_read(NV_PRMA, addr, 0);
-    return 0;
-}
-void prma_write(void *opaque, hwaddr addr,
-                               uint64_t val, unsigned int size)
-{
-    reg_log_write(NV_PRMA, addr, val);
-}
+#define DEFINE_STUB(name, region_id) \
+    uint64_t name ## _read(void *opaque, \
+                           hwaddr addr, \
+                           unsigned int size) \
+    { \
+        reg_log_read(region_id, addr, 0); \
+        return 0; \
+    } \
+    void name ## _write(void *opaque, \
+                        hwaddr addr, \
+                        uint64_t val, \
+                        unsigned int size) \
+    { \
+        reg_log_write(region_id, addr, val); \
+    } \
 
-uint64_t pcounter_read(void *opaque,
-                                  hwaddr addr, unsigned int size)
-{
-    reg_log_read(NV_PCOUNTER, addr, 0);
-    return 0;
-}
-void pcounter_write(void *opaque, hwaddr addr,
-                               uint64_t val, unsigned int size)
-{
-    reg_log_write(NV_PCOUNTER, addr, val);
-}
+DEFINE_STUB(prma, NV_PRMA)
+DEFINE_STUB(pcounter, NV_PCOUNTER)
+DEFINE_STUB(pvpe, NV_PVPE)
+DEFINE_STUB(ptv, NV_PTV)
+DEFINE_STUB(prmfb, NV_PRMFB)
+DEFINE_STUB(prmdio, NV_PRMDIO)
+DEFINE_STUB(pstraps, NV_PSTRAPS)
+// DEFINE_STUB(pramin, NV_PRAMIN)
 
-uint64_t pvpe_read(void *opaque,
-                                  hwaddr addr, unsigned int size)
-{
-    reg_log_read(NV_PVPE, addr, 0);
-    return 0;
-}
-void pvpe_write(void *opaque, hwaddr addr,
-                               uint64_t val, unsigned int size)
-{
-    reg_log_write(NV_PVPE, addr, val);
-}
-
-uint64_t ptv_read(void *opaque,
-                                  hwaddr addr, unsigned int size)
-{
-    reg_log_read(NV_PTV, addr, 0);
-    return 0;
-}
-void ptv_write(void *opaque, hwaddr addr,
-                               uint64_t val, unsigned int size)
-{
-    reg_log_write(NV_PTV, addr, val);
-}
-
-uint64_t prmfb_read(void *opaque,
-                                  hwaddr addr, unsigned int size)
-{
-    reg_log_read(NV_PRMFB, addr, 0);
-    return 0;
-}
-void prmfb_write(void *opaque, hwaddr addr,
-                               uint64_t val, unsigned int size)
-{
-    reg_log_write(NV_PRMFB, addr, val);
-}
-
-uint64_t prmdio_read(void *opaque,
-                                  hwaddr addr, unsigned int size)
-{
-    reg_log_read(NV_PRMDIO, addr, 0);
-    return 0;
-}
-void prmdio_write(void *opaque, hwaddr addr,
-                               uint64_t val, unsigned int size)
-{
-    reg_log_write(NV_PRMDIO, addr, val);
-}
-
-uint64_t pstraps_read(void *opaque,
-                                  hwaddr addr, unsigned int size)
-{
-    reg_log_read(NV_PSTRAPS, addr, 0);
-    return 0;
-}
-void pstraps_write(void *opaque, hwaddr addr,
-                               uint64_t val, unsigned int size)
-{
-    reg_log_write(NV_PSTRAPS, addr, val);
-}
-
-/* PRAMIN - RAMIN access */
-/*
-uint64_t pramin_read(void *opaque,
-                                 hwaddr addr, unsigned int size)
-{
-    NV2A_DPRINTF("nv2a PRAMIN: read [0x%" HWADDR_PRIx "] -> 0x%" HWADDR_PRIx "\n", addr, r);
-    return 0;
-}
-void pramin_write(void *opaque, hwaddr addr,
-                              uint64_t val, unsigned int size)
-{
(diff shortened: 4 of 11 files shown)


hw/xbox/nv2a/nv2a.c
```diff
@@ -163,18 +135,16 @@ DEFINE_PROTO(pcrtc)
 DEFINE_PROTO(prmcio)
 DEFINE_PROTO(pramdac)
 DEFINE_PROTO(prmdio)
-DEFINE_PROTO(pramin)
+// DEFINE_PROTO(pramin)
```
JayFoxRox: Why is pramin commented out?

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -49,25 +49,25 @@
...
-    PS_TEXTUREMODES_DOT_ZW = 0x0aL, // - - * *
-    PS_TEXTUREMODES_DOT_RFLCT_DIFF = 0x0bL, // - - * -
-    PS_TEXTUREMODES_DOT_RFLCT_SPEC = 0x0cL, // - - - *
-    PS_TEXTUREMODES_DOT_STR_3D = 0x0dL, // - - - *
-    PS_TEXTUREMODES_DOT_STR_CUBE = 0x0eL, // - - - *
-    PS_TEXTUREMODES_DPNDNT_AR = 0x0fL, // - * * *
-    PS_TEXTUREMODES_DPNDNT_GB = 0x10L, // - * * *
-    PS_TEXTUREMODES_DOTPRODUCT = 0x11L, // - * * -
+enum PS_TEXTUREMODES {                            // valid in stage 0 1 2 3
+    PS_TEXTUREMODES_NONE                 = 0x00L, // * * * *
```
JayFoxRox: This should be aligned, so that each `*` is under `0`, `1`, `2` and `3` respectively.

hw/xbox/xbox.c
```diff
@@ -204,7 +204,8 @@ static void xbox_flash_init(MemoryRegion *rom_memory)
          map_loc >= 0xff000000;
          map_loc -= bios_size) {
         map_bios = g_malloc(sizeof(*map_bios));
-        memory_region_init_alias(map_bios, NULL, "pci-bios", bios, 0, bios_size);
+        memory_region_init_alias(map_bios, NULL, "pci-bios",
+                                 bios, 0, bios_size);
```
JayFoxRox: yuk.. I liked the old one better - that's also why I hate the 80 column limit (when almost nobody uses 80 columns anymore)

*Just wanted to rant, change is acceptable - it's a good implementation of a stupid rule.*

hw/xbox/nv2a/gl/gloffscreen_cgl.c
```diff
@@ -21,21 +21,21 @@
...
 #include <OpenGL/CGLCurrent.h>
+#include <OpenGL/CGLTypes.h>
 
 struct _GloContext {
-  CGLContextObj     cglContext;
+    CGLContextObj cglContext;
 };
 
-/* Create an OpenGL context for a certain pixel format. formatflags are from 
+/* Create an OpenGL context for a certain pixel format. Format flags are from
```
JayFoxRox: Why did you touch this?
What does the comment even refer to? I only see "format attributes" below.

Also nit while we are at it: Some comments start lowercase (like the next one below).

hw/xbox/nv2a/gl/gloffscreen_glx.c
```diff
@@ -22,31 +22,33 @@
...
 };
 
-static Display* x_display;
-
+static Display *x_display;
 
 /* Create an OpenGL context */
 GloContext *glo_context_create(void)
 {
+    static bool initialized;
```
JayFoxRox: Isn't this broken now?

It used to be `static bool initialized = false;` to explicitly show that it has not been initialized.
The `false` is gone now which is less descriptive (it probably still means `0` / `false` on most platforms, but is it even guaranteed?)

hw/xbox/nv2a/gl/gloffscreen_wgl.c
```diff
@@ -43,29 +43,30 @@
...
-    HGLRC                 hContext;
+    HINSTANCE hInstance;
+    HDC hDC;
+    HWND hWnd; /* Our hidden window */
+    HGLRC hContext;
 };
 
 struct GloMain glo;
-int glo_inited = 0;
+int glo_inited;
```
JayFoxRox: Initialization removed here too.
This seems to be a recurring issue. Was this intended?

While we are at it, this could also become a `bool`, possibly `static bool` like our existing code in GLX.

hw/xbox/nv2a/nv2a.c
```diff
@@ -104,7 +101,10 @@ void *nv_dma_map(NV2AState *d, hwaddr dma_obj_address, hwaddr *len)
 
     /* TODO: Handle targets and classes properly */
     NV2A_DPRINTF("dma_map %x, %x, %" HWADDR_PRIx " %" HWADDR_PRIx "\n",
-                 dma.dma_class, dma.dma_target, dma.address, dma.limit);
+                 dma.dma_class,
+                 dma.dma_target,
+                 dma.address,
+                 dma.limit);
```
JayFoxRox: Not sure if I like this change. I prefer to keep `NV2A_DPRINTF` short and use newlines to highlight specific semantics (`x, y, z,` of a vector on the same line for example)

hw/xbox/nv2a/nv2a.c
```diff
@@ -497,23 +403,24 @@ static void nv2a_class_init(ObjectClass *klass, void *data)
...
-    .interfaces          = (InterfaceInfo[]) {
-        { INTERFACE_CONVENTIONAL_PCI_DEVICE },
-        { },
-    },
+    .class_init = nv2a_class_init,
+    .interfaces =
+        (InterfaceInfo[]){
+            { INTERFACE_CONVENTIONAL_PCI_DEVICE },
+            {},
+        },
```
JayFoxRox: These seem to be aligned in other QEMU modules I looked at.

hw/xbox/nv2a/nv2a.h
```diff
@@ -22,45 +22,44 @@
...
-#define CASE_4(v, step)                                              \
-    case (v):                                                        \
-    case (v)+(step):                                                 \
-    case (v)+(step)*2:                                               \
-    case (v)+(step)*3
+#define CASE_4(v, step)      \
+    case (v):                \
+    case ((v) + (step)):     \
+    case ((v) + (step) * 2): \
+    case ((v) + (step) * 3)
```
JayFoxRox: What's the point of the extra parentheses?

hw/xbox/nv2a/nv2a.h
```diff
@@ -248,7 +247,8 @@ typedef struct PGRAPHState {
     bool texture_matrix_enable[NV2A_MAX_TEXTURES];
 
     /* FIXME: Move to NV_PGRAPH_BUMPMAT... */
-    float bump_env_matrix[NV2A_MAX_TEXTURES-1][4]; /* 3 allowed stages with 2x2 matrix each */
+    float bump_env_matrix[NV2A_MAX_TEXTURES - 1][4]; /* 3 allowed stages with
+                                                        2x2 matrix each */
```
JayFoxRox: I dislike how the comment is split so weirdly. Even if I was on a 80 column editor, I'd rather have my terminal split this.

hw/xbox/nv2a/nv2a.h
```diff
@@ -314,8 +314,8 @@ typedef struct PGRAPHState {
 
 typedef struct CacheEntry {
     QSIMPLEQ_ENTRY(CacheEntry) entry;
-    unsigned int method : 14;
-    unsigned int subchannel : 3;
+    unsigned int method:14;
+    unsigned int subchannel:3;
```
JayFoxRox: Is this how QEMU expects bitfields? I *strongly* dislike this. It makes it way harder to see what's going on:

```c
int pgraph_cache1;
int pgraph_cache:1;
```

I prefer the more obvious:

```c
int pgraph_cache1;
int pgraph_cache : 1;
```

hw/xbox/nv2a/nv2a.h
```diff
@@ -440,9 +440,42 @@ typedef struct NV2ABlockInfo {
...
+DEFINE_PROTO(pgraph)
+DEFINE_PROTO(pcrtc)
+DEFINE_PROTO(prmcio)
+DEFINE_PROTO(pramdac)
+DEFINE_PROTO(prmdio)
+// DEFINE_PROTO(pramin)
+DEFINE_PROTO(user)
+
+#undef DEFINE_PROTO
+
```
JayFoxRox: Is there any reason why we'd need these prototypes in the header?
I think hiding them in the source code was a better choice.

hw/xbox/nv2a/nv2a.h
```diff
@@ -440,9 +440,42 @@ typedef struct NV2ABlockInfo {
 extern const struct NV2ABlockInfo blocktable[];
 extern const int blocktable_len;
 
+DMAObject nv_dma_load(NV2AState *d, hwaddr dma_obj_address);
+void *nv_dma_map(NV2AState *d, hwaddr dma_obj_address, hwaddr *len);
```
JayFoxRox: Why would this need to be public?
Are there any uses outside nv2a.c? Should there be?

*Possibly other similar instances too.*

hw/xbox/nv2a/nv2a.c
```diff
@@ -34,20 +33,18 @@ int ffs(register int valu)
 {
     register int bit;
 
-    if (valu == 0)
+    if (valu == 0) {
```
JayFoxRox: Doesn't `ERROR: use ctz32() instead of ffs()` from PR description mean we can remove this function entirely?

If not, while we are at it: `value`?

hw/xbox/nv2a/nv2a_debug.c
```diff
@@ -22,18 +22,19 @@
...
-#include <stdarg.h>
 #include <assert.h>
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdio.h>
 
-static bool has_GL_GREMEDY_frame_terminator = false;
-static bool has_GL_KHR_debug = false;
+static bool has_GL_GREMEDY_frame_terminator;
+static bool has_GL_KHR_debug;
```
JayFoxRox: I think this change is fine. Although I did not check for accidental use of debug functions before initialization.

hw/xbox/nv2a/nv2a_debug.c
```diff
@@ -22,18 +22,19 @@
...
 
 void gl_debug_initialize(void)
 {
     has_GL_KHR_debug = glo_check_extension("GL_KHR_debug");
-    has_GL_GREMEDY_frame_terminator = glo_check_extension("GL_GREMEDY_frame_terminator");
+    has_GL_GREMEDY_frame_terminator = glo_check_extension(
+        "GL_GREMEDY_frame_terminator");
```
JayFoxRox: Looks weird, having a string on same line is not bad in my opinion.

hw/xbox/nv2a/nv2a_debug.h
```diff
@@ -41,25 +42,26 @@ void gl_debug_group_end(void);
...
+        }                                             \
     } while (0)
-# define NV2A_GL_DGROUP_BEGIN(format, ...)         do { } while (0)
-# define NV2A_GL_DGROUP_END()                      do { } while (0)
-# define NV2A_GL_DLABEL(target, name, format, ...) do { } while (0)
-# define NV2A_GL_DFRAME_TERMINATOR()               do { } while (0)
+#define NV2A_GL_DGROUP_BEGIN(format, ...)         do { } while (0)
+#define NV2A_GL_DGROUP_END()                      do { } while (0)
+#define NV2A_GL_DLABEL(target, name, format, ...) do { } while (0)
+#define NV2A_GL_DFRAME_TERMINATOR()               do { } while (0)
```
JayFoxRox: These we indented to make it easier to see the `#if` / `#else` block.

hw/xbox/nv2a/nv2a.h
```diff
@@ -447,6 +447,14 @@ void *pfifo_puller_thread(void *opaque);
...
 void update_irq(NV2AState *d);
 
+
+void pgraph_context_switch(NV2AState *d, unsigned int channel_id);
+void pgraph_wait_fifo_access(NV2AState *d);
+void pgraph_method(NV2AState *d,
+                   unsigned int subchannel,
+                   unsigned int method,
+                   uint32_t parameter);
+
```
JayFoxRox: I don't think these have to be public?

hw/xbox/nv2a/nv2a_pcrtc.c
```diff
@@ -63,8 +63,10 @@ void pcrtc_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size)
         d->pcrtc.start = val;
 
         NV2A_DPRINTF("PCRTC_START - %x %x %x %x\n",
-                d->vram_ptr[val+64], d->vram_ptr[val+64+1],
-                d->vram_ptr[val+64+2], d->vram_ptr[val+64+3]);
+                     d->vram_ptr[val + 64],
```
JayFoxRox: nit: If it's formated like this, I'd probably add `+ 0`

hw/xbox/nv2a/swizzle.c
```diff
@@ -36,21 +36,33 @@
...
+        if (bit < height) {
+            y |= mask_bit;
+            mask_bit <<= 1;
+            done = false;
+        }
+        if (bit < depth) {
+            z |= mask_bit;
+            mask_bit <<= 1;
+            done = false;
+        }
```
JayFoxRox: Harder to read now in my opinion.

hw/xbox/nv2a/swizzle.c
```diff
@@ -102,27 +116,26 @@ void swizzle_box(
...
-                const uint8_t *src = src_buf
-                                         + y * row_pitch + x * bytes_per_pixel;
-                uint8_t *dst = dst_buf + get_swizzled_offset(x, y, 0,
-                                                             mask_x, mask_y, 0,
-                                                             bytes_per_pixel);
+                const uint8_t *src = src_buf + y * row_pitch +
+                                     x * bytes_per_pixel;
+                uint8_t *dst = dst_buf +
+                               get_swizzled_offset(
+                                   x, y, 0, mask_x, mask_y, 0, bytes_per_pixel);
```
JayFoxRox: Also harder to read now in my opinion.

hw/xbox/nv2a/swizzle.c
```diff
@@ -131,9 +144,14 @@ void unswizzle_box(
...
-                    + get_swizzled_offset(x, y, z, mask_x, mask_y, mask_z,
-                                          bytes_per_pixel);
+                const uint8_t *src = src_buf +
+                                     get_swizzled_offset(x,
+                                                         y,
+                                                         z,
+                                                         mask_x,
+                                                         mask_y,
+                                                         mask_z,
+                                                         bytes_per_pixel);
```
JayFoxRox: Harder to read now.

hw/xbox/nv2a/swizzle.c
```diff
@@ -142,24 +160,23 @@ void unswizzle_box(
...
+void unswizzle_rect(const uint8_t *src_buf,
+                    unsigned int width,
+                    unsigned int height,
+                    uint8_t *dst_buf,
+                    unsigned int pitch,
+                    unsigned int bytes_per_pixel)
 {
-    unswizzle_box(src_buf, width, height, 1, dst_buf, pitch, 0, bytes_per_pixel);
+    unswizzle_box(
+        src_buf, width, height, 1, dst_buf, pitch, 0, bytes_per_pixel);
```
JayFoxRox: I'd rather see the list of arguments split halfway through than moving them into the next line.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -27,16 +27,16 @@
 
 #include "qemu/osdep.h"
 
+#include <stdbool.h>
+#include <stdint.h>
```
JayFoxRox: Note to self: Pausing review at this point (done commit-by-commit, so please don't force-push). I feel this needs a lot of concentration. Will continue later.

hw/xbox/nv2a/nv2a_debug.c
mborgerson: Agree, but it got wrapped by clang-format because of the 80 character line limit rule.

hw/xbox/nv2a/nv2a.h
mborgerson: Due to splitting up nv2a.c into nv2a_pgraph.c and nv2a_pfifo.c

hw/xbox/nv2a/swizzle.c
mborgerson: I think it's fine either way, we can revert this change

hw/xbox/nv2a/swizzle.c
mborgerson: Got wrapped for 80 column limit, I think it looks okay though?

hw/xbox/nv2a/swizzle.c
mborgerson: Idk why clang-format does this sometimes but not others, but agree

hw/xbox/nv2a/gl/gloffscreen_cgl.c
mborgerson: Because it looks like a typo, is it not?

hw/xbox/nv2a/gl/gloffscreen_glx.c
mborgerson: As we discussed via Discord, according to checkpatch.pl static/global variables should not be initialized to `0`, `false`, or `NULL` values (as they should already have those values upon program startup). I made this change to comply with the script, but I've since changed my mind and I think XQEMU can deviate from this QEMU guideline in favor of readability.

hw/xbox/nv2a/gl/gloffscreen_wgl.c
mborgerson: See above. Agree to change to `bool` type.

hw/xbox/nv2a/nv2a.c
mborgerson: Agree to change `valu` to `value`.

Think s/`ffs`/`ctz32`/ should happen in another PR.

hw/xbox/nv2a/nv2a.c
mborgerson: Still not sure why clang-format packs them sometimes but not others. I'm okay to have them on one line as well.

hw/xbox/nv2a/gl/gloffscreen_cgl.c
JayFoxRox: I literally have no idea what the comment even wants to express to be honest. I can't find the respective code to match this comment.

But to me it read like `formatflags` was actually a variable / field or type of some sort (implying it's not a typo).
The comment below also doesn't even use "format flags". It only refers to "pixel format attributes".

hw/xbox/nv2a/nv2a.h
mborgerson: Yes, IIRC this is a suggestion by checkpatch. I'm indifferent here.

hw/xbox/nv2a/nv2a.h
mborgerson: Yes, see nv2a_[pgraph|pfifo].c

hw/xbox/nv2a/nv2a.c
JayFoxRox: Sounds good to me.
I'll create an issue about this one so we don't forget about it.

hw/xbox/nv2a/nv2a_debug.h
mborgerson: Indifferent here, clang-format auto cleaned these

hw/xbox/nv2a/swizzle.c
mborgerson: Need to double check line limit, if it's < 80 then I'm ok to put them back on one line

hw/xbox/nv2a/nv2a_debug.h
JayFoxRox: I think we should either indent them, or at least have an empty line around the `#if` `#else` `#endif` so it's more obvious what's going on.

mborgerson: I'll go w/ blank line.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -49,120 +49,128 @@
...
-    PS_TEXTUREMODES_DOT_RFLCT_DIFF=       0x0bL, // - - * -
-    PS_TEXTUREMODES_DOT_RFLCT_SPEC=       0x0cL, // - - - *
-    PS_TEXTUREMODES_DOT_STR_3D=           0x0dL, // - - - *
-    PS_TEXTUREMODES_DOT_STR_CUBE=         0x0eL, // - - - *
-    PS_TEXTUREMODES_DPNDNT_AR=            0x0fL, // - * * *
-    PS_TEXTUREMODES_DPNDNT_GB=            0x10L, // - * * *
-    PS_TEXTUREMODES_DOTPRODUCT=           0x11L, // - * * -
-    PS_TEXTUREMODES_DOT_RFLCT_SPEC_CONST= 0x12L, // - - - *
-    // 0x13-0x1f reserved
+enum PS_TEXTUREMODES { // valid in stage 0 1 2 3
```
JayFoxRox: Bad alignment

*(This was brought up before, so I assume this comment is outdated I'll review a fixup in a few minutes - reviewing commit-by-commit right now)*

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -49,120 +49,128 @@
...
+    PS_INPUTMAPPING_EXPAND_NEGATE =
+        0x60L, // 1 - 2*max(0,x)   invalid for final combiner
+    PS_INPUTMAPPING_HALFBIAS_NORMAL =
+        0x80L, // max(0,x) - 1/2   invalid for final combiner
+    PS_INPUTMAPPING_HALFBIAS_NEGATE =
+        0xa0L, // 1/2 - max(0,x)   invalid for final combiner
+    PS_INPUTMAPPING_SIGNED_IDENTITY =
+        0xc0L, // x                invalid for final combiner
+    PS_INPUTMAPPING_SIGNED_NEGATE =
+        0xe0L, // -x               invalid for final combiner
```
JayFoxRox: Formatting of this entire bog is bad in my opinion. The table format was much better (even if it exceeds 80 cols).

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -49,120 +49,128 @@
...
+    PS_REGISTER_V0 = 0x04L, // r/w
+    PS_REGISTER_V1 = 0x05L, // r/w
+    PS_REGISTER_T0 = 0x08L, // r/w
+    PS_REGISTER_T1 = 0x09L, // r/w
+    PS_REGISTER_T2 = 0x0aL, // r/w
+    PS_REGISTER_T3 = 0x0bL, // r/w
+    PS_REGISTER_R0 = 0x0cL, // r/w
+    PS_REGISTER_R1 = 0x0dL, // r/w
+    PS_REGISTER_V1R0_SUM = 0x0eL, // r
+    PS_REGISTER_EF_PROD = 0x0fL, // r
```
JayFoxRox: I prefered the table-like formatting from before.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -49,120 +49,128 @@
...
+                      PS_INPUTMAPPING_UNSIGNED_INVERT, // OK for final combiner
+    PS_REGISTER_NEGATIVE_ONE =
+        PS_REGISTER_ZERO |
+        PS_INPUTMAPPING_EXPAND_NORMAL, // invalid for final combiner
+    PS_REGISTER_ONE_HALF =
+        PS_REGISTER_ZERO |
+        PS_INPUTMAPPING_HALFBIAS_NEGATE, // invalid for final combiner
+    PS_REGISTER_NEGATIVE_ONE_HALF =
+        PS_REGISTER_ZERO |
+        PS_INPUTMAPPING_HALFBIAS_NORMAL, // invalid for final combiner
```
JayFoxRox: Liked the old formatting better on this block. At least we should attempt to *always* have the `_ZERO` in either the first or second line, but not mixing it.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -49,120 +49,128 @@
...
 
-    PS_COMBINERCOUNT_SAME_C0=     0x0000L, // c0 same in each stage
-    PS_COMBINERCOUNT_UNIQUE_C0=   0x0010L, // c0 unique in each stage
+    PS_COMBINERCOUNT_SAME_C0 = 0x0000L, // c0 same in each stage
+    PS_COMBINERCOUNT_UNIQUE_C0 = 0x0010L, // c0 unique in each stage
 
-    PS_COMBINERCOUNT_SAME_C1=     0x0000L, // c1 same in each stage
-    PS_COMBINERCOUNT_UNIQUE_C1=   0x0100L  // c1 unique in each stage
+    PS_COMBINERCOUNT_SAME_C1 = 0x0000L, // c1 same in each stage
+    PS_COMBINERCOUNT_UNIQUE_C1 = 0x0100L // c1 unique in each stage
```
JayFoxRox: Please re-align as it suggests these belong together (I'd re-align each pair individually).

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -49,120 +49,128 @@
...
-    PS_COMBINEROUTPUT_CD_MULTIPLY=         0x00L,
-    PS_COMBINEROUTPUT_CD_DOT_PRODUCT=      0x01L, // RGB only
+    PS_COMBINEROUTPUT_CD_MULTIPLY = 0x00L,
+    PS_COMBINEROUTPUT_CD_DOT_PRODUCT = 0x01L, // RGB only
 
-    PS_COMBINEROUTPUT_AB_CD_SUM=           0x00L, // 3rd output is AB+CD
-    PS_COMBINEROUTPUT_AB_CD_MUX=           0x04L, // 3rd output is MUX(AB,CD) based on R0.a
+    PS_COMBINEROUTPUT_AB_CD_SUM = 0x00L, // 3rd output is AB+CD
+    PS_COMBINEROUTPUT_AB_CD_MUX =
+        0x04L, // 3rd output is MUX(AB,CD) based on R0.a
```
JayFoxRox: Also re-align.

With so many aligned stuff (these are literally tables translated into code-form), we should consider keeping the old enums in my opinion.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -49,120 +49,128 @@
...
-{
-    PS_FINALCOMBINERSETTING_CLAMP_SUM=     0x80, // V1+R0 sum clamped to [0,1]
+enum PS_FINALCOMBINERSETTING {
+    PS_FINALCOMBINERSETTING_CLAMP_SUM = 0x80, // V1+R0 sum clamped to [0,1]
 
-    PS_FINALCOMBINERSETTING_COMPLEMENT_V1= 0x40, // unsigned invert mapping
+    PS_FINALCOMBINERSETTING_COMPLEMENT_V1 = 0x40, // unsigned invert mapping
 
-    PS_FINALCOMBINERSETTING_COMPLEMENT_R0= 0x20, // unsigned invert mapping
+    PS_FINALCOMBINERSETTING_COMPLEMENT_R0 = 0x20, // unsigned invert mapping
```
JayFoxRox: Good catch! Also don't care about alignment here.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -177,21 +185,21 @@ struct InputVarInfo {
 struct FCInputInfo {
     struct InputInfo a, b, c, d, e, f, g;
     int c0, c1;
-    //uint32_t c0_value, c1_value;
+    // uint32_t c0_value, c1_value;
```
JayFoxRox: Probably doesn't belong in this PR, but we should probably remove this, or add a comment instead as to why these are not here (I actually don't know from the top of my head).

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -177,21 +185,21 @@ struct InputVarInfo {
...
     bool c0_used, c1_used;
     bool v1r0_sum, clamp_sum, inv_v1, inv_r0, enabled;
 };
 
 struct OutputInfo {
-    int ab, cd, muxsum, flags, ab_op, cd_op, muxsum_op,
-        mapping, ab_alphablue, cd_alphablue;
+    int ab, cd, muxsum, flags, ab_op, cd_op, muxsum_op, mapping, ab_alphablue,
+        cd_alphablue;
```
JayFoxRox: Why did we ever decide to put these all in one line anyway?
In my opinion we should probably split these into multiple lines anyway (maybe grouped slightly like in `FCInputInfo`)

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -177,21 +185,21 @@ struct InputVarInfo {
...
 };
 
 struct PSStageInfo {
     struct InputVarInfo rgb_input, alpha_input;
     struct OutputInfo rgb_output, alpha_output;
     int c0, c1;
-    //uint32_t c0_value, c1_value;
+    // uint32_t c0_value, c1_value;
```
JayFoxRox: See other comment about `c0_value` and `c1_value`.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -203,7 +211,7 @@ struct PixelShader {
     struct FCInputInfo final_input;
     int tex_modes[4], input_tex[4];
 
-    //uint32_t dot_mapping, input_texture;
+    // uint32_t dot_mapping, input_texture;
```
JayFoxRox: Should probably add a comment that we don't support dotmapping yet (or do we store this thing elsewhere?)
Not even sure what `input_texture` was (or supposed to be) used for.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -218,23 +226,27 @@ struct PixelShader {
 static void add_var_ref(struct PixelShader *ps, const char *var)
 {
     int i;
-    for (i=0; i<ps->num_var_refs; i++) {
-        if (strcmp((char*)ps->var_refs[i], var) == 0) return;
+    for (i = 0; i < ps->num_var_refs; i++) {
+        if (strcmp((char *)ps->var_refs[i], var) == 0) {
```
JayFoxRox: I prefer `!strcmp`.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -218,23 +226,27 @@ struct PixelShader {
...
+            return;
+        }
     }
-    strcpy((char*)ps->var_refs[ps->num_var_refs++], var);
+    strcpy((char *)ps->var_refs[ps->num_var_refs++], var);
```
JayFoxRox: I'm a bit surprised, what type is `var_refs`? Why do we even need a cast? This looks sketchy.

*Not your change, but still had me wondering.*

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -218,23 +226,27 @@ struct PixelShader {
...
     int i;
-    for (i=0; i<ps->num_const_refs; i++) {
-        if (strcmp((char*)ps->const_refs[i], var) == 0) return;
+    for (i = 0; i < ps->num_const_refs; i++) {
+        if (strcmp((char *)ps->const_refs[i], var) == 0) {
+            return;
+        }
     }
-    strcpy((char*)ps->const_refs[ps->num_const_refs++], var);
+    strcpy((char *)ps->const_refs[ps->num_const_refs++], var);
```
JayFoxRox: *(Same comment as with previous block)*

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -310,13 +323,14 @@ static QString* get_var(struct PixelShader *ps, int reg, bool is_dest)
 }
 
 // Get input variable code
-static QString* get_input_var(struct PixelShader *ps, struct InputInfo in, bool is_alpha)
+static QString *
+get_input_var(struct PixelShader *ps, struct InputInfo in, bool is_alpha)
```
JayFoxRox: Not a big fan of linebreaks like this. I'd rather break the parameter list.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -417,19 +433,23 @@ static void add_stage_code(struct PixelShader *ps,
     QString *ab;
     if (output.ab_op == PS_COMBINEROUTPUT_AB_DOT_PRODUCT) {
         ab = qstring_from_fmt("dot(%s, %s)",
-                              qstring_get_str(a), qstring_get_str(b));
+                              qstring_get_str(a),
+                              qstring_get_str(b));
```
JayFoxRox: Indifferent about these changes.

*Other instances too.*

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -439,44 +459,63 @@ static void add_stage_code(struct PixelShader *ps,
     QString *sum_dest = get_var(ps, output.muxsum, true);
 
     if (qstring_get_length(ab_dest)) {
-        qstring_append_fmt(ps->code, "%s.%s = %s(%s);\n",
-                           qstring_get_str(ab_dest), write_mask, caster, qstring_get_str(ab_mapping));
+        qstring_append_fmt(ps->code,
+                           "%s.%s = %s(%s);\n",
+                           qstring_get_str(ab_dest),
+                           write_mask,
```
JayFoxRox: I'd like to have the writemask on the same line as `ab_dest` because they have a *very* strong relation.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -542,30 +584,38 @@ static QString* psh_convert(struct PixelShader *ps)
...
+        qstring_append_fmt(preflight,
+                           "uniform ivec4 clipRegion[%d];\n",
                            ps->state.window_clip_count);
-        qstring_append_fmt(clip, "/*  Window-clip (%s) */\n",
-                           ps->state.window_clip_exclusive ?
-                               "Exclusive" : "Inclusive");
+        qstring_append_fmt(clip,
+                           "/*  Window-clip (%s) */\n",
+                           ps->state.window_clip_exclusive ? "Exclusive" :
+                                                             "Inclusive");
```
JayFoxRox: Here, (and in surrounding code) I feel the old code benefitted readability, as it was easier to spot the GLSL.

I'm not feeling very strongly about this either. So I'd tolerate this.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -542,30 +584,38 @@ static QString* psh_convert(struct PixelShader *ps)
...
-        qstring_append_fmt(clip, "for (int i = 0; i < %d; i++) {\n",
+        qstring_append_fmt(clip,
+                           "for (int i = 0; i < %d; i++) {\n",
                            ps->state.window_clip_count);
-        qstring_append(clip, "  bvec4 clipTest = bvec4(lessThan(gl_FragCoord.xy, clipRegion[i].xy),\n"
-                             "                         greaterThan(gl_FragCoord.xy, clipRegion[i].zw));\n"
-                             "  if (!any(clipTest)) {\n");
+        qstring_append(clip,
+                       "  bvec4 clipTest = bvec4(lessThan(gl_FragCoord.xy, "
+                       "clipRegion[i].xy),\n"
```
JayFoxRox: Yuck! Cutting GLSL into multiline here is a bad idea in my opinion.
I think it's a good case where ignoring the 80 col limit is much more beneficial.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -581,7 +631,9 @@ static QString* psh_convert(struct PixelShader *ps)
     qstring_append(vars, "vec4 pD1 = vtx.D1 / vtx.inv_w;\n");
     qstring_append(vars, "vec4 pB0 = vtx.B0 / vtx.inv_w;\n");
     qstring_append(vars, "vec4 pB1 = vtx.B1 / vtx.inv_w;\n");
-    qstring_append(vars, "vec4 pFog = vec4(fogColor.rgb, clamp(vtx.Fog / vtx.inv_w, 0.0, 1.0));\n");
+    qstring_append(vars,
+                   "vec4 pFog = vec4(fogColor.rgb, clamp(vtx.Fog / vtx.inv_w, "
+                   "0.0, 1.0));\n");
```
JayFoxRox: I'd undo this regardless of 80 col limit, simply because it's in the same block with many similar lines, so there's no reason why this should stand out.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -593,43 +645,58 @@ static QString* psh_convert(struct PixelShader *ps)
...
             } else {
                 sampler_type = "sampler2D";
             }
-            qstring_append_fmt(vars, "vec4 t%d = textureProj(texSamp%d, pT%d.xyw);\n",
-                               i, i, i);
+            qstring_append_fmt(vars,
+                               "vec4 t%d = textureProj(texSamp%d, pT%d.xyw);\n",
+                               i,
+                               i,
+                               i);
```
JayFoxRox: I think the old one was better.

*Other instances too.*

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -593,43 +645,58 @@ static QString* psh_convert(struct PixelShader *ps)
...
+                vars,
+                "vec4 t%d = vec4(0.0); /* PS_TEXTUREMODES_CLIPPLANE */\n",
+                i);
             for (j = 0; j < 4; j++) {
-                qstring_append_fmt(vars, "  if(pT%d.%c %s 0.0) { discard; };\n",
-                                   i, "xyzw"[j],
+                qstring_append_fmt(vars,
+                                   "  if(pT%d.%c %s 0.0) { discard; };\n",
+                                   i,
+                                   "xyzw"[j],
```
JayFoxRox: swizzlemask (and writemasks) should always be on same line as register / variable they belong to in my opinion.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -732,8 +848,10 @@ static QString* psh_convert(struct PixelShader *ps)
     for (i = 0; i < ps->num_stages; i++) {
         ps->cur_stage = i;
         qstring_append_fmt(ps->code, "// Stage %d\n", i);
-        add_stage_code(ps, ps->stage[i].rgb_input, ps->stage[i].rgb_output, "rgb", false);
-        add_stage_code(ps, ps->stage[i].alpha_input, ps->stage[i].alpha_output, "a", true);
+        add_stage_code(
+            ps, ps->stage[i].rgb_input, ps->stage[i].rgb_output, "rgb", false);
+        add_stage_code(
+            ps, ps->stage[i].alpha_input, ps->stage[i].alpha_output, "a", true);
```
JayFoxRox: I'd break the `"rgb"` and `"a"`

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -761,19 +879,32 @@ static QString* psh_convert(struct PixelShader *ps)
...
+                break;
+            case ALPHA_FUNC_GREATER:
+                alpha_op = ">";
+                break;
+            case ALPHA_FUNC_NOTEQUAL:
+                alpha_op = "!=";
+                break;
+            case ALPHA_FUNC_GEQUAL:
+                alpha_op = ">=";
+                break;
```
JayFoxRox: We should eventually move this into a seperate function probably.
I'm not sure if I like this longer version. I don't feel strongly about it, but I think the previous version with some alignment (table-style) would be the most readable, and avoid most accidental errors.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -847,40 +980,48 @@ QString *psh_translate(const PshState state)
...
-        //ps.stage[i].c0 = (pDef->PSC0Mapping >> (i * 4)) & 0xF;
-        //ps.stage[i].c1 = (pDef->PSC1Mapping >> (i * 4)) & 0xF;
-        //ps.stage[i].c0_value = constant_0[i];
-        //ps.stage[i].c1_value = constant_1[i];
+        parse_combiner_output(state.alpha_outputs[i],
+                              &ps.stage[i].alpha_output);
+        // ps.stage[i].c0 = (pDef->PSC0Mapping >> (i * 4)) & 0xF;
+        // ps.stage[i].c1 = (pDef->PSC1Mapping >> (i * 4)) & 0xF;
+        // ps.stage[i].c0_value = constant_0[i];
+        // ps.stage[i].c1_value = constant_1[i];
```
JayFoxRox: I wonder if we should use this instead:

```C
#if 0
// This is dead because <reason>
<dead code>
#endif
```

As for reason here: no clue.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -847,40 +980,48 @@ QString *psh_translate(const PshState state)
...
         ps.final_input.inv_v1 = flags & PS_FINALCOMBINERSETTING_COMPLEMENT_V1;
         ps.final_input.inv_r0 = flags & PS_FINALCOMBINERSETTING_COMPLEMENT_R0;
-        //ps.final_input.c0 = (pDef->PSFinalCombinerConstants >> 0) & 0xF;
-        //ps.final_input.c1 = (pDef->PSFinalCombinerConstants >> 4) & 0xF;
-        //ps.final_input.c0_value = final_constant_0;
-        //ps.final_input.c1_value = final_constant_1;
+        // ps.final_input.c0 = (pDef->PSFinalCombinerConstants >> 0) & 0xF;
+        // ps.final_input.c1 = (pDef->PSFinalCombinerConstants >> 4) & 0xF;
+        // ps.final_input.c0_value = final_constant_0;
+        // ps.final_input.c1_value = final_constant_1;
```
JayFoxRox: *Same thing about dead code, see note above.*

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -307,28 +305,30 @@ static QString* decode_swizzle(const uint32_t *shader_token,
...
-    if (x == SWIZZLE_X && y == SWIZZLE_Y
-        && z == SWIZZLE_Z && w == SWIZZLE_W) {
+    if (x == SWIZZLE_X && y == SWIZZLE_Y && z == SWIZZLE_Z && w == SWIZZLE_W) {
         /* Don't print the swizzle if it's .xyzw */
         return qstring_from_str(""); // Will turn ".xyzw" into "."
+
     /* Don't print duplicates */
     } else if (x == y && y == z && z == w) {
-        return qstring_from_str((char[]){'.', swizzle_str[x], '\0'});
+        return qstring_from_str((char[]){ '.', swizzle_str[x], '\0' });
```
JayFoxRox: I'd break on `swizzle_str[x]`

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -307,28 +305,30 @@ static QString* decode_swizzle(const uint32_t *shader_token,
...
     } else if (y == z && z == w) {
-        return qstring_from_str((char[]){'.',
-            swizzle_str[x], swizzle_str[y], '\0'});
+        return qstring_from_str(
+            (char[]){ '.', swizzle_str[x], swizzle_str[y], '\0' });
```
JayFoxRox: I'd undo the break-change. This and the previous change would make the repeating code pattern more obvious.

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -358,7 +358,8 @@ static QString* decode_opcode_input(const uint32_t *shader_token,
     case PARAM_C:
         reg_num = convert_c_register(vsh_get_field(shader_token, FLD_CONST));
         if (vsh_get_field(shader_token, FLD_A0X) > 0) {
-            //FIXME: does this really require the "correction" doe in convert_c_register?!
+            // FIXME: does this really require the "correction" doe in
+            // convert_c_register?!
```
JayFoxRox: If we break this anyway, we could also just use C89 style comment here.
There's other `/* */ ` comments anyway.

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -423,16 +422,17 @@ static QString* decode_opcode(const uint32_t *shader_token,
             /* TODO : Emulate writeable const registers */
             qstring_append(ret, "c");
             qstring_append_int(ret,
-                convert_c_register(
-                    vsh_get_field(shader_token, FLD_OUT_ADDRESS)));
+                               convert_c_register(
+                                   vsh_get_field(shader_token,
+                                                 FLD_OUT_ADDRESS)));
```
JayFoxRox: I prefer the old version. Not feeling too strongly about this one.

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -423,16 +422,17 @@ static QString* decode_opcode(const uint32_t *shader_token,
...
         } else {
-            qstring_append(ret,
-                out_reg_name[
-                    vsh_get_field(shader_token, FLD_OUT_ADDRESS) & 0xF]);
+            qstring_append(
+                ret,
+                out_reg_name[vsh_get_field(shader_token, FLD_OUT_ADDRESS) &
+                             0xF]);
```
JayFoxRox: I *strongly* prefer the old version. The new one manages to hide the mask or what it is applied to (or just makes it less easy to read).

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -482,7 +482,8 @@ static QString* decode_token(const uint32_t *shader_token)
             qstring_append(inputs_mac, qstring_get_str(input_c));
         }
 
-        /* Then prepend these inputs with the actual opcode, mask, and input : */
+        /* Then prepend these inputs with the actual opcode, mask, and input :
+         */
```
JayFoxRox: Wat. Remove the ` : ` and we should be good?

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -751,7 +766,8 @@ void vsh_translate(uint16_t version,
          * on OS X.
          */
         "  if (clipRange.y != clipRange.x) {\n"
-        "    oPos.z = (oPos.z - 0.5 * (clipRange.x + clipRange.y)) / (0.5 * (clipRange.y - clipRange.x));\n"
+        "    oPos.z = (oPos.z - 0.5 * (clipRange.x + clipRange.y)) / (0.5 * "
+        "(clipRange.y - clipRange.x));\n"
```
JayFoxRox: Again: Linebreak in GLSL is extremly dangerous. Very often bug source.

hw/xbox/nv2a/nv2a_vsh.h
```diff
@@ -73,16 +73,16 @@ enum VshSkinning {
...
-#define VSH_VERSION_XVS                    0x2078
+#define VSH_VERSION_XVS 0x2078
 
 // Xbox vertex state shader
-#define VSH_VERSION_XVSS                   0x7378
+#define VSH_VERSION_XVSS 0x7378
 
 // Xbox vertex read/write shader
-#define VSH_VERSION_XVSW                   0x7778
+#define VSH_VERSION_XVSW 0x7778
```
JayFoxRox: Do we even need these defines? I believe they are for HLE only.

hw/xbox/nv2a/nv2a_vsh.h
```diff
@@ -136,7 +136,8 @@ void vsh_translate(uint16_t version,
                    const uint32_t *tokens,
                    unsigned int length,
                    bool z_perspective,
-                   QString *header, QString *body);
+                   QString *header,
+                   QString *body);
```
JayFoxRox: *Reminder: Depends on the change in the C file.*

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -697,7 +698,7 @@ static QString *psh_convert(struct PixelShader *ps)
                                    "  if(pT%d.%c %s 0.0) { discard; };\n",
                                    i,
                                    "xyzw"[j],
-                                   ps->state.compare_mode[i][j] ? ">=" : "<");
+                                   ps -> state.compare_mode[i][j] ? ">=" : "<");
```
JayFoxRox: huh? Do we *really* want to leave spaces here?

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -947,15 +948,15 @@ static void parse_combiner_inputs(uint32_t value,
...
-    out->muxsum_op = flags & 4;
-    out->mapping = flags & 0x38;
+    out->cd           = value & 0xF;
+    out->ab           = (value >> 4) & 0xF;
+    out->muxsum       = (value >> 8) & 0xF;
+    int flags         = value >> 12;
+    out->flags        = flags;
+    out->cd_op        = flags & 1;
+    out->ab_op        = flags & 2;
+    out->muxsum_op    = flags & 4;
```
JayFoxRox: Please make them hex too.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -969,7 +970,7 @@ QString *psh_translate(const PshState state)
     ps.state = state;
 
     ps.num_stages = state.combiner_control & 0xFF;
-    ps.flags = state.combiner_control >> 8;
+    ps.flags      = state.combiner_control >> 8;
```
JayFoxRox: nit: not necessary

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -1012,7 +1013,7 @@ QString *psh_translate(const PshState state)
                               &ps.final_input.f,
                               &ps.final_input.g,
                               &blank);
-        int flags = state.final_inputs_1 & 0xFF;
+        int flags                = state.final_inputs_1 & 0xFF;
```
JayFoxRox: nit: not necessary

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -965,12 +953,14 @@ QString *psh_translate(const PshState state)
 {
     int i;
     struct PixelShader ps;
-    memset(&ps, 0, sizeof(ps));
+    struct InputInfo blank;
```
JayFoxRox: nit: How is this even used? Looks like it only serves as a throw-away?
If so, I'd rather pass a `NULL` and let the function deal with it internally.

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -149,17 +147,17 @@ typedef struct VshOpcodeParams {
...
+//     /* ILU OP       ParamA ParamB ParamC */
+//     /* ILU_NOP */ { false, false, false }, // Dxbx note : Unused
+//     /* ILU_MOV */ { false, false, true  },
+//     /* ILU_RCP */ { false, false, true  },
+//     /* ILU_RCC */ { false, false, true  },
+//     /* ILU_RSQ */ { false, false, true  },
+//     /* ILU_EXP */ { false, false, true  },
+//     /* ILU_LOG */ { false, false, true  },
+//     /* ILU_LIT */ { false, false, true  },
+// };
```
JayFoxRox: `#if 0` or complete removal instead?

Why was this removed (and in such a weird way?)

hw/xbox/nv2a/nv2a.c
```diff
@@ -162,6 +162,7 @@ const struct NV2ABlockInfo blocktable[] = {
 
 const int blocktable_len = ARRAY_SIZE(blocktable);
 
+// FIXME: Add nv2a_reg_names or remove this code
```
JayFoxRox: I vote to remove this instead.

Adding a list would quickly lead to completionists who leak copyrighted lists.
So despite this being insanely useful, idiots would ruin the fun for others by doing illegal stuff on our behalf.
I'd be fine with this if we used envytool names (but unless the whole codebase uses them, the nvidia names would likely return in PRs).

hw/xbox/nv2a/nv2a.c
```diff
@@ -240,9 +241,9 @@ static void nv2a_get_offsets(VGACommonState *s,
 static void nv2a_vga_gfx_update(void *opaque)
 {
     VGACommonState *vga = opaque;
-    vga->hw_ops->gfx_update(vga);
-
     NV2AState *d = container_of(vga, NV2AState, vga);
+
+    vga->hw_ops->gfx_update(vga);
```
JayFoxRox: I liked the old code better as it split VGA and `NV2AState` more cleanly.

hw/xbox/nv2a/nv2a.c
```diff
@@ -274,7 +275,7 @@ static void nv2a_init_memory(NV2AState *d, MemoryRegion *ram)
 
     memory_region_add_subregion(&d->mmio, 0x700000, &d->ramin);
 
-    d->vram_ptr = memory_region_get_ram_ptr(d->vram);
+    d->vram_ptr  = memory_region_get_ram_ptr(d->vram);
```
JayFoxRox: Not necessary imo. We don't have to align *everything*

hw/xbox/nv2a/nv2a.c
```diff
@@ -312,12 +312,11 @@ static void nv2a_realize(PCIDevice *dev, Error **errp)
 
     dev->config[PCI_INTERRUPT_PIN] = 0x01;
 
-    d->pcrtc.start = 0;
-
-    d->pramdac.core_clock_coeff = 0x00011c01; /* 189MHz...? */
-    d->pramdac.core_clock_freq = 189000000;
+    d->pcrtc.start                = 0;
+    d->pramdac.core_clock_coeff   = 0x00011c01; /* 189MHz...? */
+    d->pramdac.core_clock_freq    = 189000000;
```
JayFoxRox: I dislike the alignment here, because it hides that we mix hex and dec here.

hw/xbox/nv2a/nv2a.c
```diff
@@ -328,11 +327,11 @@ static void nv2a_realize(PCIDevice *dev, Error **errp)
     vga->msr = VGA_MIS_COLOR;
 
     vga_common_init(vga, OBJECT(dev), false); // FIXME: true or false? idk
-    vga->get_bpp = nv2a_get_bpp;
+    vga->get_bpp     = nv2a_get_bpp;
```
JayFoxRox: This and next block also don't need alignment. It just groups unrelated things which does not benefit readibility.

hw/xbox/nv2a/nv2a.c
mborgerson: Huh? How does it hide that?

hw/xbox/nv2a/nv2a_vsh.c
mborgerson: This was unused by the code entirely, but I left it in for now. Vote to remove entirely if it's not actually needed (but I'm not entirely sure tbh).

hw/xbox/nv2a/nv2a.h
```diff
@@ -446,21 +444,16 @@ void pgraph_init(NV2AState *d);
...
 
-#define DEFINE_PROTO(prefix)                                              \
-    uint64_t prefix##_read(void *opaque, hwaddr addr, unsigned int size); \
-    void prefix##_write(void *opaque,                                     \
-                        hwaddr addr,                                      \
-                        uint64_t val,                                     \
-                        unsigned int size);
+#define DEFINE_PROTO(n)                                              \
+    uint64_t n##_read(void *opaque, hwaddr addr, unsigned int size); \
+    void n##_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size);
```
JayFoxRox: nit: This is shorter but also slightly less readable. I think the `prefix` name helped beginners to understand what goes on here. `n` is a lot more cryptic. I'd rather shorten or remove the argument names.

hw/xbox/amd_smbus.c
```diff
@@ -37,7 +37,7 @@
 #endif
 
 /* AMD756 SMBus address offsets */
-#define SMB_ADDR_OFFSET     0xE0
+#define SMB_ADDR_OFFSET 0xE0
 #define SMB_IOSIZE      16
```
JayFoxRox: nit: I'd probably remove alignment as it's mixing hex and dec.

hw/xbox/amd_smbus.c
```diff
@@ -209,7 +208,7 @@ void amd756_smb_ioport_writeb(void *opaque, uint32_t addr, uint32_t val)
 
 uint32_t amd756_smb_ioport_readb(void *opaque, uint32_t addr)
 {
-    AMD756SMBus *s = opaque;
+    AMD756SMBus *s = (AMD756SMBus *)opaque;
```
JayFoxRox: Shouldn't there be a macro which checks type here in QOM?

hw/xbox/chihiro-usb.c
```diff
@@ -293,17 +293,17 @@ static void chihiro_an2131sc_class_initfn(ObjectClass *klass, void *data)
...
+    uc->product_desc = "Chihiro an2131sc";
+    uc->usb_desc     = &desc_chihiro_an2131sc;
 
     uc->handle_reset   = handle_reset;
     uc->handle_control = handle_control;
     uc->handle_data    = handle_data;
     uc->handle_attach  = usb_desc_attach;
 
-    //dc->vmsd = &vmstate_usb_kbd;
+    // dc->vmsd = &vmstate_usb_kbd;
```
JayFoxRox: `#if 0` or full removal? What is this for?

hw/xbox/lpc47m157.c
```diff
@@ -47,7 +47,12 @@
 #define CONFIG_DEVICE_BASE_ADDRESS_LOW      0x61
 #define CONFIG_DEVICE_INETRRUPT             0x70
 
-#define DEBUG_LPC47M157
+// #define DEBUG
+#ifdef DEBUG
+# define XBOX_DPRINTF(format, ...)     printf(format, ## __VA_ARGS__)
+#else
+# define XBOX_DPRINTF(format, ...)     do { } while (0)
```
JayFoxRox: Do we really need the `XBOX_` prefix?

hw/xbox/lpc47m157.c
```diff
@@ -97,11 +101,11 @@ static void update_devices(LPC47M157State *s)
 static void lpc47m157_io_write(void *opaque, hwaddr addr, uint64_t val,
                                unsigned int size)
 {
-    LPC47M157State *s = opaque;
+    LPC47M157State *s = (LPC47M157State *)opaque;
```
JayFoxRox: Again: QOM has macros to do these casts I believe?

hw/xbox/lpc47m157.c
```diff
@@ -139,7 +141,7 @@ static void lpc47m157_io_write(void *opaque, hwaddr addr, uint64_t val,
 
 static uint64_t lpc47m157_io_read(void *opaque, hwaddr addr, unsigned int size)
 {
-    LPC47M157State *s = opaque;
+    LPC47M157State *s = (LPC47M157State *)opaque;
```
JayFoxRox: *QOM Typecast thing again*

hw/xbox/lpc47m157.c
```diff
@@ -235,10 +234,10 @@ static void lpc47m157_class_init(ObjectClass *klass, void *data)
 }
 
 static const TypeInfo lpc47m157_type_info = {
-    .name          = "lpc47m157",
-    .parent        = TYPE_ISA_DEVICE,
+    .name = "lpc47m157",
+    .parent = TYPE_ISA_DEVICE,
     .instance_size = sizeof(LPC47M157State),
-    .class_init    = lpc47m157_class_init,
+    .class_init = lpc47m157_class_init,
```
JayFoxRox: Didn't this PR also align these for other types?

hw/xbox/lpc47m157.c
mborgerson: No, this was a copy-paste from another file, we can remove prefix

hw/xbox/amd_smbus.c
mborgerson: I don't understand the argument?

hw/xbox/nv2a/nv2a_debug.c
```diff
@@ -57,13 +57,14 @@ void gl_debug_initialize(void)
 
 void gl_debug_message(bool cc, const char *fmt, ...)
 {
+    size_t n;
+    char buffer[1024];
+    va_list ap;
```
JayFoxRox: Personal opinion: I prefer variable declarations where they are used. This legacy "everything on top" is often leading to much scrolling to check types in my opinion. Anyhow, fine in this instance, I just wanted to whine a little.

hw/xbox/nv2a/nv2a_stubs.c
```diff
@@ -19,21 +19,19 @@
...
-    } \
-    void name ## _write(void *opaque, \
-                        hwaddr addr, \
-                        uint64_t val, \
-                        unsigned int size) \
-    { \
-        reg_log_write(region_id, addr, val); \
-    } \
+#define DEFINE_STUB(name, region_id)                                   \
+    uint64_t name##_read(void *opaque, hwaddr addr, unsigned int size) \
```
JayFoxRox: Should use the same name as `DEFINE_PROTO` for first argument (was `prefix`, is `n`).

hw/xbox/nv2a/nv2a_user.c
```diff
@@ -59,21 +61,23 @@ uint64_t user_read(void *opaque, hwaddr addr, unsigned int size)
 void user_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size)
 {
     NV2AState *d = (NV2AState *)opaque;
+    unsigned int channel_id;
+    ChannelControl *control;
+    uint32_t channel_modes;
```
JayFoxRox: Same as above: Prefer declaration and initialization together for this function.

hw/xbox/smbus_adm1032.c
```diff
@@ -28,14 +28,19 @@
 #include "hw/i2c/smbus.h"
 #include "smbus.h"
 
-#define DEBUG
+// #define DEBUG
+#ifdef DEBUG
+# define XBOX_DPRINTF(format, ...)     printf(format, ## __VA_ARGS__)
+#else
+# define XBOX_DPRINTF(format, ...)     do { } while (0)
```
JayFoxRox: *Same question as before re `XBOX` prefix*

hw/xbox/smbus_cx25871.c
```diff
@@ -23,56 +23,54 @@
...
 
 static void cx_send_byte(SMBusDevice *dev, uint8_t val)
 {
-#ifdef DEBUG
-    printf("cx_send_byte: addr=0x%02x val=0x%02x\n",
-           dev->i2c.address, val);
-#endif
+    XBOX_DPRINTF("cx_send_byte: addr=0x%02x val=0x%02x\n",
+                 dev->i2c.address,
+                 val);
```
JayFoxRox: For the debug stuff in particular I'd like to keep it short. so on same line was better in my opinion.

hw/xbox/xbox.c
```diff
@@ -531,17 +529,17 @@ static void xbox_machine_class_init(ObjectClass *oc, void *data)
...
-    .class_init = xbox_machine_class_init,
-    .interfaces = (InterfaceInfo[]) {
-         // { TYPE_HOTPLUG_HANDLER },
-         // { TYPE_NMI },
-         { }
+    .class_size    = sizeof(XboxMachineClass),
+    .class_init    = xbox_machine_class_init,
+    .interfaces    = (InterfaceInfo[]){
+        // { TYPE_HOTPLUG_HANDLER },
+        // { TYPE_NMI },
```
JayFoxRox: Remove dead code?

hw/xbox/amd_smbus.c
JayFoxRox: If you align these, it's easier to accidentally use them in the wrong way (such as using 237384 instead of 0x237384 and vice versa):

```c
#define FOO 0x237384
#define ABC 10341238
#define BAR 0x569473
```

While this does not apply as much here, it can still be an issue.
So to avoid this, I typically avoid aligning different number bases altogether.
We *do* have this issue in the NV2A clock constants.

*(But I also like to keep offsets in hex, and most sizes in decimal by the way)*

hw/xbox/nv2a/nv2a.c
```diff
@@ -154,70 +124,75 @@ DEFINE_PROTO(user)
 #include "nv2a_stubs.c"
 #include "nv2a_user.c"
 
+#define ENTRY(NAME, OFFSET, SIZE, RDFUNC, WRFUNC)      \
+    [NV_##NAME] = {                                    \
+        .name   = #NAME,                               \
+        .offset = OFFSET,                              \
+        .size   = SIZE,                                \
+        .ops    = { .read = RDFUNC, .write = WRFUNC }, \
```
JayFoxRox: nit: could paste these from NAME in lowercase

hw/xbox/nv2a/nv2a_stubs.c
```diff
@@ -19,100 +19,29 @@
...
+        reg_log_read(region_id, addr, 0); \
+        return 0; \
+    } \
+    void name ## _write(void *opaque, \
+                        hwaddr addr, \
+                        uint64_t val, \
+                        unsigned int size) \
+    { \
+        reg_log_write(region_id, addr, val); \
+    } \
```
JayFoxRox: nit: Extra backslash slightly dangerous in my opinion

mborgerson:
This PR has now been open for two weeks, and I consider it ready to merge. Have you finished your review @JayFoxRox? I plan to merge this myself within 24 hours if there is no activity.

JayFoxRox:
I still plan to review it a final time and merge it if it's. 
But I have some things to deal with in real-life right now.
Please wait with merge.

hw/xbox/nv2a/nv2a.c
JayFoxRox: This was not done yet. But it's fine with me, given that it will be removed in the future anyway.

hw/xbox/nv2a/nv2a.c
JayFoxRox: This was not done. A fixup would be nice.
I *really* think we'll run into potential legal issues with this, so I'd rather see it removed before merge.

hw/xbox/nv2a/nv2a.h
JayFoxRox: Got no feedback here. Would accept as-is because it doesn't change behaviour.

hw/xbox/nv2a/nv2a_stubs.c
JayFoxRox: This was not done. I'd appreciate if it was done, as at least one other multi-line macro in this change doesn't seem to have it either.

JayFoxRox:
I have reviewed the current code and commit style.
I have also ran XQEMU with this patch and used a debug bios, and a retail bios and ran Broken Sword 3.
I have also enabled the new DEBUG macros for testing.

I have only had some small comments left:

- Some discussions were left unanswered, but these are minor formatting, so nothing preventing a merge.
- "sio: Replace debug printf statements with macro": I noticed this one also moves `int i` to loop start. I thought it was a bit odd; but nothing preventing merge. The commit title prefix is also not ideal as sio typically refers to Serial I/O. but the LPC is Super I/O which does a lot more.
- `nv2a_reg_names` is the only one I feel rather strongly about as I had *very* bad experience with similar placeholders / comments in the past. Anyhow, we can just reject bad PRs and contact support to have them removed if necesary.


So I consider this to be complete.

Merging.

PR #76, opened 2018-07-09
nv2a FIFO Refactoring

mborgerson:
This patch ports over the following commits by @espes from the XQEMU 1.x tree (available via tag archive-xqemu-1.x) to the refactored nv2a code:

- 4d9107e8 (HEAD -> xbox, upstream/xbox) Merge branch 'fix-fifo' into xbox
- c33f4ab2 cleanups
- d3f83d93 cleanups
- faca5dd0 fix
- 7c62d7c1 fixes
- 8ad239b0 wip
- 45ed3c31 wip
- c006d5e8 wip

### DIFF
--- a/hw/xbox/nv2a/nv2a.c
+++ b/hw/xbox/nv2a/nv2a.c
@@ -95,16 +95,16 @@ DMAObject nv_dma_load(NV2AState *d, hwaddr dma_obj_address)
 
 void *nv_dma_map(NV2AState *d, hwaddr dma_obj_address, hwaddr *len)
 {
-    assert(dma_obj_address < memory_region_size(&d->ramin));
-
     DMAObject dma = nv_dma_load(d, dma_obj_address);
 
     /* TODO: Handle targets and classes properly */
-    NV2A_DPRINTF("dma_map %x, %x, %" HWADDR_PRIx " %" HWADDR_PRIx "\n",
+    NV2A_DPRINTF("dma_map %" HWADDR_PRIx " - %x, %x, %" HWADDR_PRIx " %" HWADDR_PRIx "\n",
+                 dma_obj_address,
                  dma.dma_class, dma.dma_target, dma.address, dma.limit);
 
     dma.address &= 0x07FFFFFF;
 
+    assert(dma.address < memory_region_size(d->vram));
     // assert(dma.address + dma.limit < memory_region_size(d->vram));
     *len = dma.limit;
     return d->vram_ptr + dma.address;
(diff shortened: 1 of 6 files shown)


hw/xbox/nv2a/nv2a.c
```diff
@@ -506,22 +511,11 @@ static void nv2a_realize(PCIDevice *dev, Error **errp)
...
 
-    /* init fifo cache1 */
-    qemu_spin_init(&d->pfifo.cache1.alloc_lock);
-    qemu_mutex_init(&d->pfifo.cache1.cache_lock);
-    qemu_cond_init(&d->pfifo.cache1.cache_cond);
-    QSIMPLEQ_INIT(&d->pfifo.cache1.cache);
-    QSIMPLEQ_INIT(&d->pfifo.cache1.working_cache);
-    QSIMPLEQ_INIT(&d->pfifo.cache1.available_entries);
-    QSIMPLEQ_INIT(&d->pfifo.cache1.retired_entries);
-
```
PatrickvL: So you decided to do away with this optimization?

JayFoxRox: Yes, because it was a hack. Hardware doesn't work this way. The way the hardware works is exposed via registers and is simpler.

It's also necessary to implement the limited cache size as D3D is able to patch the FIFO while its still being processed. If you process the entire FIFO in one step, then you might run into issues as the GPU is not aware of those patches.

Doing it properly will hopefully also make it possible to [use nv2a-trace](https://github.com/XboxDev/nv2a-trace) in the future.

mborgerson: @PatrickvL Yeah, the `available_entries` and `retired_entries` queues is an optimization to the older code which was not HW accurate. That optimization did improve performance and is fine for now, but going forward we want to clean up the FIFO stuff altogether so it'll get thrown out with the rest of the cache entry stuff, and that's what this PR seeks to do.

gandalfthewhite19890404:
PR cause completely freezing in Lord of the rings Third Age:
image

Yellow thing appear on master too, but it much less (just around menu) than here.
How it looks on master:
image

mborgerson:
**Debug Notes:** I spent some time working on this last night, here are some notes. Early titles don't have an issue (Halo:CE gets in game for instance), but code compiled with later XDKs are broken with the update. This issue appears to be timing sensitive indicating a possible problem with locking in its current form.

Eventually the guest will enter into a tight loop waiting for PGRAPH register 0xb10 to be set, which will never happen. It might be that this specific loop involves waiting for some operation in the FIFO that is supposed to set this register, but it never happens, so the guest hangs.

Interestingly, if I break out of the tight loop it'll continue just fine to render a frame. I can then set a breakpoint and when continuing only a couple of times per second, the guest never enters this tight loop indicating possible timing/locking issue.

gandalfthewhite19890404:
Lord of the rings: Third age is fixed, now its not freezing in main menu and looks like in master:
image

PR #82, opened 2018-07-13
Fix copying of DLLs on Windows

### LINKED ISSUE #62: ./get_deps.py: "IOError: [Errno 2] No such file or directory: '/mingw64/bin/libgcc_s_seh-1.dll'"
I have not seen this happen on AppVeyor CI yet, which suggests this is environment specific.

It's quite bad because this happens at the end of compilation, so it looks like XQEMU compilation has failed, despite having worked fine.

**Environment**

I'm running 0523deaa938dd267a17bcc5102a49dd58ebffe79 with #56 and #61, but had the same issue happen without those 2 PRs added.
My system is Windows 10 Pro N 64-bit (10.0, Build 17133).
MSYS2 has been upgraded to latest version using `pacman -Syu`

**Package versions**

```
$ pacman -Qs mingw-w64-x86_64-toolchain && pacman -Qs python2 && pacman -Qs python3
local/mingw-w64-x86_64-binutils 2.30-4 (mingw-w64-x86_64-toolchain)
    A set of programs to assemble and manipulate binary and object files (mingw-w64)
local/mingw-w64-x86_64-crt-git 6.0.0.5144.45e00853-1 (mingw-w64-x86_64-toolchain)
    MinGW-w64 CRT for Windows
local/mingw-w64-x86_64-gcc 7.3.0-2 (mingw-w64-x86_64-toolchain)
    GNU Compiler Collection (C,C++,OpenMP) for MinGW-w64
local/mingw-w64-x86_64-gcc-libgfortran 7.3.0-2 (mingw-w64-x86_64-toolchain)
    GNU Compiler Collection (libgfortran) for MinGW-w64
local/mingw-w64-x86_64-gcc-libs 7.3.0-2 (mingw-w64-x86_64-toolchain)
    GNU Compiler Collection (libraries) for MinGW-w64
local/mingw-w64-x86_64-gdb 8.1-2 (mingw-w64-x86_64-toolchain)
    GNU Debugger (mingw-w64)
local/mingw-w64-x86_64-headers-git 6.0.0.5144.45e00853-1 (mingw-w64-x86_64-toolchain)
    MinGW-w64 headers for Windows
local/mingw-w64-x86_64-libmangle-git 6.0.0.5079.3b7a42fd-1 (mingw-w64-x86_64-toolchain)
    MinGW-w64 libmangle
local/mingw-w64-x86_64-libwinpthread-git 6.0.0.5134.2416de71-1 (mingw-w64-x86_64-toolchain)
    MinGW-w64 winpthreads library
local/mingw-w64-x86_64-make 4.2.1-2 (mingw-w64-x86_64-toolchain)
    GNU make utility to maintain groups of programs (mingw-w64)
local/mingw-w64-x86_64-pkg-config 0.29.2-1 (mingw-w64-x86_64-toolchain)
    A system for managing library compile/link flags (mingw-w64)
local/mingw-w64-x86_64-tools-git 6.0.0.5111.3bc5ab74-1 (mingw-w64-x86_64-toolchain)
    MinGW-w64 tools
local/mingw-w64-x86_64-winpthreads-git 6.0.0.5134.2416de71-1 (mingw-w64-x86_64-toolchain)
    MinGW-w64 winpthreads library
local/mingw-w64-x86_64-pyqt5-common 5.10.1-1
    Common PyQt files shared between python3-pyqt5 and python2-pyqt5
local/mingw-w64-x86_64-python2 2.7.15-1
    A high-level scripting language (mingw-w64)
local/mingw-w64-x86_64-python2-pyqt5 5.10.1-1
    A set of Python 2.x bindings for the Qt5 toolkit
local/mingw-w64-x86_64-python2-sip 4.19.9-1
    Python 2.x SIP bindings for C and C++ libraries (mingw-w64)
local/python2 2.7.15-2
    A high-level scripting language
local/mingw-w64-x86_64-pyqt5-common 5.10.1-1
    Common PyQt files shared between python3-pyqt5 and python2-pyqt5
local/mingw-w64-x86_64-python3 3.6.5-1
    A high-level scripting language (mingw-w64)
local/mingw-w64-x86_64-python3-pyqt5 5.10.1-1
    A set of Python 3.x bindings for the Qt5 toolkit
local/mingw-w64-x86_64-python3-sip 4.19.9-1
    Python 3.x SIP bindings for C and C++ libraries (mingw-w64)
local/python 3.6.6-1
    Next generation of the python high-level scripting language
local/python3-appdirs 1.4.3-2
    A small Python module for determining appropriate platform-specific dirs, e.g. a "user data dir".
local/python3-packaging 16.8-3
    Core utilities for Python packages
local/python3-pip 10.0.1-1
    The PyPA recommended tool for installing Python packages
local/python3-pyparsing 2.2.0-2
    General parsing module for Python
local/python3-setuptools 39.2.0-1
    Easily download, build, install, upgrade, and uninstall Python packages
local/python3-six 1.10.0-5
    Python 2 and 3 compatibility utilities
```

**Detailed log**

Running `./build_windows.sh` results in errors in dependency collection phase:

```
+ python2 ./get_deps.py dist/xqemu.exe dist
Skipping system DLL /c/WINDOWS/SYSTEM32/ntdll.dll
Skipping system DLL /c/WINDOWS/System32/KERNEL32.DLL
Skipping system DLL /c/WINDOWS/System32/KERNELBASE.dll
Skipping system DLL /c/WINDOWS/System32/ADVAPI32.dll
Skipping system DLL /c/WINDOWS/System32/msvcrt.dll
Skipping system DLL /c/WINDOWS/System32/sechost.dll
Skipping system DLL /c/WINDOWS/System32/RPCRT4.dll
Skipping system DLL /c/WINDOWS/System32/GDI32.dll
Skipping system DLL /c/WINDOWS/System32/gdi32full.dll
Skipping system DLL /c/WINDOWS/System32/msvcp_win.dll
Skipping system DLL /c/WINDOWS/System32/ucrtbase.dll
Skipping system DLL /c/WINDOWS/System32/USER32.dll
Skipping system DLL /c/WINDOWS/System32/win32u.dll
Skipping system DLL /c/WINDOWS/System32/ole32.dll
Skipping system DLL /c/WINDOWS/System32/combase.dll
Skipping system DLL /c/WINDOWS/SYSTEM32/IPHLPAPI.DLL
Skipping system DLL /c/WINDOWS/System32/bcryptPrimitives.dll
Skipping system DLL /c/WINDOWS/System32/SHELL32.dll
Skipping system DLL /c/WINDOWS/System32/cfgmgr32.dll
Skipping system DLL /c/WINDOWS/SYSTEM32/OPENGL32.dll
Skipping system DLL /c/WINDOWS/System32/shcore.dll
Skipping system DLL /c/WINDOWS/SYSTEM32/GLU32.dll
Skipping system DLL /c/WINDOWS/System32/windows.storage.dll
Skipping system DLL /c/WINDOWS/System32/shlwapi.dll
Skipping system DLL /c/WINDOWS/System32/kernel.appcore.dll
Skipping system DLL /c/WINDOWS/System32/profapi.dll
Skipping system DLL /c/WINDOWS/System32/powrprof.dll
Skipping system DLL /c/WINDOWS/System32/FLTLIB.DLL
Skipping system DLL /c/WINDOWS/System32/WS2_32.dll
Skipping system DLL /c/WINDOWS/SYSTEM32/WINMM.dll
Copying /mingw64/bin/libssp-0.dll...
Traceback (most recent call last):
  File "./get_deps.py", line 40, in <module>
    main()
  File "./get_deps.py", line 37, in main
    shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
  File "C:/msys64/mingw64/lib/python2.7/shutil.py", line 96, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: '/mingw64/bin/libssp-0.dll'
```

Re-running (`./build_windows.sh`) gives:

```
...
Skipping system DLL /c/WINDOWS/System32/powrprof.dll
Skipping system DLL /c/WINDOWS/System32/FLTLIB.DLL
Skipping system DLL /c/WINDOWS/System32/WS2_32.dll
Skipping system DLL /c/WINDOWS/SYSTEM32/WINMM.dll
Copying /mingw64/bin/libgcc_s_seh-1.dll...
Traceback (most recent call last):
  File "./get_deps.py", line 40, in <module>
    main()
  File "./get_deps.py", line 37, in main
    shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
  File "C:/msys64/mingw64/lib/python2.7/shutil.py", line 96, in copyfile
    with open(src, 'rb') as fsrc:
IOError: [Errno 2] No such file or directory: '/mingw64/bin/libgcc_s_seh-1.dll'
```

**Notes**

Running more often seems to get '/mingw64/bin/libgcc_s_seh-1.dll' again, although I'm not sure - I only ran this about 5 times as each build takes a considerable amount of time on this machine. Could be that it still randomly returns the other dll too (maybe others too).

muemart:
This fixes the Python script copying the DLLs by converting the Unix path to Windows paths. Apparently Python recognizes that it's running on Windows and thus interprets the paths wrong.
I was building on Drive D, and Python interpreted the path "/mingw64/bin/libgcc_s_seh-1.dll" as "D:/mingw64/bin/libgcc_s_seh-1.dll", which was wrong in my case.
This fixes #62.

### DIFF
--- a/get_deps.py
+++ b/get_deps.py
@@ -30,14 +30,23 @@ def main():
 		if dll_name.startswith('???'):
 			print('Unknown DLL?')
 			continue
-		if dll_path.lower().startswith('/c/windows') or dll_path.lower().startswith('c:/windows'):
+		try:
+			# ldd on msys2 gives Unix-style paths, but Python wants them Windows-style
+			# Try cygpath to convert them
+			dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
+		except:
+			# cygpath doesn't exist or failed. Carry on and hope for the best
+			pass
+		if dll_path.lower().startswith('c:\\windows'):
 			print('Skipping system DLL %s' % dll_path)
 			continue
 
 		print('Copying %s...' % dll_path)
-		# Python wants Windows-style paths, not Unix-style
-		dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
-		shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
+		try:
+			shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
+		except shutil.Error as err:
+			# copyfile produces an exception if the files are the same
+			print('Copying failed: %s' % err)
 
 if __name__ == '__main__':
 	main()
--- a/get_deps.py
+++ b/get_deps.py
@@ -22,6 +22,13 @@ def main():
 	elif not os.path.isdir(args.dest):
 		print('File exists with destination name')
 		sys.exit(1)
+		
+	has_cygpath = True
+	try:
+		subprocess.check_output(['cygpath', '--help'])
+	except OSError as err:
+		print("Couldn't execute cygpath (Reason: '%s'). Continuing without." % err)
+		has_cygpath = False
 
 	sout = subprocess.check_output(['ldd', args.prog])
 	for line in sout.splitlines():
@@ -30,24 +37,23 @@ def main():
 		if dll_name.startswith('???'):
 			print('Unknown DLL?')
 			continue
-		try:
+		if has_cygpath:
 			# ldd on msys2 gives Unix-style paths, but Python wants them Windows-style
-			# Try cygpath to convert them
+			# If we have cygpath, convert the paths
 			dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
-		except:
-			# cygpath doesn't exist or failed. Carry on and hope for the best
-			pass
 		if dll_path.lower().startswith('c:\\windows'):
 			print('Skipping system DLL %s' % dll_path)
 			continue
 
-		print('Copying %s...' % dll_path)
-		try:
-			shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
-		except shutil.Error as err:
-			# This exception only gets raised if both files are exactly the same, but
-			# that doesn't matter to us
-			print('Copying failed: %s' % err)
+		dest_path = os.path.join(args.dest, dll_name)
+		if os.path.normcase(os.path.realpath(dll_path)) == os.path.normcase(os.path.realpath(dest_path)):
+			# If the DLL is already in the same folder as the executable,
+			# ldd will return the path to that DLL and copyfile will raise an exception
+			# because we try to copy a file over itself
+			print('DLL %s is already next to executable. Skipping copy.' % dll_name)
+		else:
+			print('Copying %s...' % dll_path)
+			shutil.copyfile(dll_path, dest_path)
 
 if __name__ == '__main__':
 	main()
--- a/get_deps.py
+++ b/get_deps.py
@@ -23,12 +23,12 @@ def main():
 		print('File exists with destination name')
 		sys.exit(1)
 		
-	has_cygpath = True
 	try:
 		subprocess.check_output(['cygpath', '--help'])
 	except OSError as err:
-		print("Couldn't execute cygpath (Reason: '%s'). Continuing without." % err)
-		has_cygpath = False
+		print("Couldn't execute cygpath (Reason: '%s')." % err)
+		print("Make sure you're using a recent version of MSYS2 and that it works correctly.")
+		sys.exit(1)
 
 	sout = subprocess.check_output(['ldd', args.prog])
 	for line in sout.splitlines():
@@ -37,10 +37,9 @@ def main():
 		if dll_name.startswith('???'):
 			print('Unknown DLL?')
 			continue
-		if has_cygpath:
-			# ldd on msys2 gives Unix-style paths, but Python wants them Windows-style
-			# If we have cygpath, convert the paths
-			dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
+		# ldd on msys gives Unix-style paths, but mingw Python wants them Windows-style
+		# Use cygpath to convert the paths, because both mingw and msys Python can handle them
+		dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
 		if dll_path.lower().startswith('c:\\windows'):
 			print('Skipping system DLL %s' % dll_path)
 			continue
--- a/get_deps.py
+++ b/get_deps.py
@@ -45,7 +45,8 @@ def main():
 		try:
 			shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
 		except shutil.Error as err:
-			# copyfile produces an exception if the files are the same
+			# This exception only gets raised if both files are exactly the same, but
+			# that doesn't matter to us
 			print('Copying failed: %s' % err)
 
 if __name__ == '__main__':
--- a/get_deps.py
+++ b/get_deps.py
@@ -33,7 +33,10 @@ def main():
 		if dll_path.lower().startswith('/c/windows') or dll_path.lower().startswith('c:/windows'):
 			print('Skipping system DLL %s' % dll_path)
 			continue
+
 		print('Copying %s...' % dll_path)
+		# Python wants Windows-style paths, not Unix-style
+		dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
 		shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
 
 if __name__ == '__main__':


mborgerson:
Are you using native Python for Windows or are you using msys installed Python? The answer to that question may impact how paths are interpreted and cause a failure. In any case, we should probably handle both situations gracefully.

muemart:
I executed the scripts from the mingw64 shell, and `where python2` tells me this:
C:\msys64\mingw64\bin\python2.exe
C:\msys64\usr\bin\python2.exe
So I'm definitely using msys Python. I found some stackoverflow questions with the same problem, and the only good solution seems to be to use cygpath.

get_deps.py
```diff
@@ -33,7 +33,10 @@ def main():
 		if dll_path.lower().startswith('/c/windows') or dll_path.lower().startswith('c:/windows'):
 			print('Skipping system DLL %s' % dll_path)
 			continue
+
 		print('Copying %s...' % dll_path)
+		# Python wants Windows-style paths, not Unix-style
+		dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
```
JayFoxRox: Shouldn't this be done before:

```
if dll_path.lower().startswith('/c/windows') or dll_path.lower().startswith('c:/windows'): 
```

so that it can be simplified?

---

What happens if `cygpath` binary does not exist?

JayFoxRox:
Thanks for looking into this! This PR fixes #62 for me.

I think we should consider the case where `cygpath` does not exist in the future, but it can also be fixed later if the issue arises (the current issue is probably more common).

Aside from what I commented on: LGTM

get_deps.py
```diff
@@ -30,14 +30,23 @@ def main():
...
 			continue
-		if dll_path.lower().startswith('/c/windows') or dll_path.lower().startswith('c:/windows'):
+		try:
+			# ldd on msys2 gives Unix-style paths, but Python wants them Windows-style
+			# Try cygpath to convert them
+			dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
+		except:
+			# cygpath doesn't exist or failed. Carry on and hope for the best
+			pass
+		if dll_path.lower().startswith('c:\\windows'):
```
JayFoxRox: heh, damnit.. now you might have 'c:\\windows'  (if cygpath succeeded) or 'c:/windows' (if cygpath failed). Maybe python os path can fix it?

Anyhow this feels like bikeshedding.

get_deps.py
```diff
@@ -30,14 +30,23 @@ def main():
...
 
 		print('Copying %s...' % dll_path)
-		# Python wants Windows-style paths, not Unix-style
-		dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
-		shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
+		try:
+			shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
+		except shutil.Error as err:
+			# copyfile produces an exception if the files are the same
+			print('Copying failed: %s' % err)
```
JayFoxRox: Failing the copy is probably a good thing so we know the dist is broken.

muemart: This exception appears if the file to be copied is exactly the same as the one at the destination. If you build twice, you'll get exceptions the second time because the dlls are already there. If they're not the same, the destination gets overridden no questions asked. If the destination isn't writable, according to the docs you'll get an IOError which won't be caught here.
It's a bit weird, but apparently that's how it is.

get_deps.py
muemart: Yeah, those issues are probably best dealt with when they appear. Because you have to build xqemu with msys2, I'm not convinced we're ever in the situation where cygpath isn't available. Not to mention you have to run the script in msys2 anyway, otherwise ldd isn't available.

get_deps.py
JayFoxRox: Just checked the docs. You are right. Thanks for your response

nit: the comment could make this more clear

get_deps.py
JayFoxRox: nit: [`os.path.normcase(path)` could be used](https://docs.python.org/2/library/os.path.html#os.path.normcase) instead of the .lower() on every use and it would solve this issue.

muemart: Hm, normcase replaces backslashes by forward slashes on msys, which goes against the docs. The offical Python release for Windows behaves as documented though. I'll hold off on this for now, because it doesn't seem to improve the situation.

JayFoxRox:
Did test and review again: still works with new changes. Also checked AppVeyor CI and it still works.

@mborgerson please review and **squash**-merge if you accept this change.

get_deps.py
```diff
@@ -30,11 +30,24 @@ def main():
 		if dll_name.startswith('???'):
 			print('Unknown DLL?')
 			continue
-		if dll_path.lower().startswith('/c/windows') or dll_path.lower().startswith('c:/windows'):
+		try:
+			# ldd on msys2 gives Unix-style paths, but Python wants them Windows-style
+			# Try cygpath to convert them
+			dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
+		except:
+			# cygpath doesn't exist or failed. Carry on and hope for the best
```
mborgerson: I'm not a fan of swallowing exceptions without so much as an error message. Please either (a) `exit(1)` due to `cygpath` being unavailable (it should be in msys2 environment!) or (b) at least print an error message or something.

get_deps.py
```diff
@@ -30,11 +30,24 @@ def main():
...
+		if dll_path.lower().startswith('c:\\windows'):
 			print('Skipping system DLL %s' % dll_path)
 			continue
+
 		print('Copying %s...' % dll_path)
-		shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
+		try:
+			shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
+		except shutil.Error as err:
+			# This exception only gets raised if both files are exactly the same, but
```
mborgerson: Like above, not a fan of swallowing exceptions, especially when we *know* why this exception will be raised. According to the [documentation](https://docs.python.org/2/library/shutil.html#shutil.copyfile):

> If src and dst are the same files, Error is raised. The destination location must be writable; otherwise, an IOError exception will be raised. If dst already exists, it will be replaced.

I don't read this as "files are exactly the same" (meaning contents), rather the file paths are the same. Is `ldd` returning paths to the libraries copied into the `dist` directory? If yes, then this would explain why the error message is being generated. In this case I would say either (a) we copy a fresh version of the lib from the lib directory or (b) skip copying if the file path already exists.

muemart:
@mborgerson you're right, my eyes glazed over when reading the error message with long filepaths . If the dlls are already next to the exe, ldd will return paths to those, and we end up copying the dll over itself.
I reworked the script and now hopefully everyone is happy. First checking if cygpath is available before using it, and preventing copying the dll over itself. Should have been less lazy from the start 😅

get_deps.py
```diff
@@ -30,24 +37,23 @@ def main():
...
 
-		print('Copying %s...' % dll_path)
-		try:
-			shutil.copyfile(dll_path, os.path.join(args.dest, dll_name))
-		except shutil.Error as err:
-			# This exception only gets raised if both files are exactly the same, but
-			# that doesn't matter to us
-			print('Copying failed: %s' % err)
+		dest_path = os.path.join(args.dest, dll_name)
+		if os.path.normcase(os.path.realpath(dll_path)) == os.path.normcase(os.path.realpath(dest_path)):
```
JayFoxRox: Why not https://docs.python.org/2/library/os.path.html#os.path.samefile ?

muemart: Unix only, sadly.

JayFoxRox: I'm stupid. yes.

get_deps.py
```diff
@@ -30,11 +37,23 @@ def main():
 		if dll_name.startswith('???'):
 			print('Unknown DLL?')
 			continue
-		if dll_path.lower().startswith('/c/windows') or dll_path.lower().startswith('c:/windows'):
+		if has_cygpath:
```
mborgerson: I don't understand why we continue on if cygpath isn't available... if we don't have cygpath shouldn't we just fail? Because otherwise (according to the comment) we would be incorrectly feeding UNIX style paths into Python.

...but more generally I haven't looked into this issue on my own machine recently, but this originally used to work for me and on the CI system. I'm still guessing that *somehow* native Python is being invoked (despite `which` showing msys2 version) to call this script. Notice how in the appveyor.bat script we explicitly update `PATH` to include msys2 binaries first.

JayFoxRox: This is bikeshedding by now. Let's stick to any of the proposed solutions which work (this, or the original version which just resulted in a backtrace).

I brought up the case of cygpath not existing as older versions of MSYS2 did not ship it. There are probably non-MSYS2/CYGWIN environments which could be used to compile QEMU.
I felt getting a best-effort approach or seeing a meaningful error message would be better than seeing a backtrace in those cases.

I'll test which python is being ran tomorrow *again*. Also keep in mind that MSYS has different versions of python for different purposes. In my opinion, either version of python should work, so using cygpath sounds good to me (with cygwin being a popular / recommended environment).

*(Note that I did not test this last revision, and I'd like to test it before merge, because I'm affected by this issue and want to be sure that this fixes it)*

muemart: I don't know much about how the folder layout on Appveyor looks like, but here's my guess about why it works.
- In the mingw64 shell (NOT the msys2 shell), `os.path.abspath('/mingw64/test')` results in `C:/mingw64/test`, which is wrong. Strangely enough, in the msys2 shell the result is correct (`/mingw64/test`).
- ldd gives back paths like `/mingw64/bin/libffi-6.dll`. Based on the above, Python will interpret this path as `C:/mingw64/bin/libffi-6.dll`. For me, this is wrong in on two accounts. First, msys2 is installed in `C:/msys64`, and secondly, I'm building on disk D, so Python actually interprets the path as `D:/mingw64/bin/libffi-6.dll`.
- Based on `appveyor.bat`, Appveyor only installs python2 but not mingw-w64-x86_64-python2 (which I have installed too), so it uses the Python version that converts paths correctly and works fine.
Regarding the absence of cygpath, I don't really feel strongy one way or the other. I guess it depends on how many scenarios and build environments we try to support.

mborgerson: I was confused about why CI builds were working before and after this patch, so I've done a little testing on my Windows 10 VM with the MINGW64 MSYS2 shell with the MSYS2 build of Python 2. Interestingly, it appears to handle paths in both the UNIX style format **and** the Windows native path format. But I'm guessing the version of Python that you have installed is a little pickier and only accepts the Windows style paths.

Some version info and testing:
```
$ python2
Python 2.7.13 (default, Feb 14 2017, 14:46:01)
[GCC 6.3.0] on msys
Type "help", "copyright", "credits" or "license" for more information.
>>> f = open(r'C:\Windows\notepad.exe', 'rb')
>>> len(f.read())
245760
>>> f2 = open('/c/Windows/notepad.exe', 'rb')
>>> len(f2.read())
245760
>>> import os.path
>>> os.path.abspath('/c/Windows/notepad.exe')
'/c/Windows/notepad.exe'
>>> os.path.abspath(r'C:\Windows\notepad.exe')
'C:\\Windows\\notepad.exe'
>>>
``` 

So this explains why it seems that CI is working before/after this PR. For this reason I'm okay with just using `cygpath` to convert paths to Windows format.

As for the missing `cygpath` issue, let's just print an error message and `exit(1)` if we fail to run `cygpath`. If the situation arises where we need to support an environment without `cygpath` in the future, we can solve it then.

Thanks!

muemart: For the record, this is the output in the mingw64 shell after installing `mingw-w64-x86_64-python2`:
```
$ python2
Python 2.7.15 (default, May  4 2018, 07:39:01)  [GCC 7.3.0 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os.path
>>> os.path.abspath('/c/windows.notepad.exe')
'C:/c/windows.notepad.exe'
>>> os.path.abspath('c:\\windows\\notepad.exe')
'c:/windows/notepad.exe'
>>>
```

get_deps.py
```diff
@@ -22,6 +22,13 @@ def main():
 	elif not os.path.isdir(args.dest):
 		print('File exists with destination name')
 		sys.exit(1)
+		
+	try:
+		subprocess.check_output(['cygpath', '--help'])
+	except OSError as err:
+		print("Couldn't execute cygpath (Reason: '%s')." % err)
+		print("Make sure you're using a recent version of MSYS2 and that it works correctly.")
+		sys.exit(1)
```
JayFoxRox: nit: I think it would have been nicer to throw this error on actual use below.

get_deps.py
```diff
@@ -30,11 +37,22 @@ def main():
 		if dll_name.startswith('???'):
 			print('Unknown DLL?')
 			continue
-		if dll_path.lower().startswith('/c/windows') or dll_path.lower().startswith('c:/windows'):
+		# ldd on msys gives Unix-style paths, but mingw Python wants them Windows-style
+		# Use cygpath to convert the paths, because both mingw and msys Python can handle them
+		dll_path = subprocess.check_output(['cygpath', '-w', dll_path]).strip()
```
JayFoxRox: nit: `-m` (I always prefer forward slashes)

*Would also need change in comparison below and might have other implications; again this is nitpicking = optional request.*

JayFoxRox:
Tested the latest version and it still works fine for me.

Unless there's further (possibly breaking) changes, I'll leave this PR to @mborgerson

Alexpux:
Mingw python can't get with posix paths because Windows programs dont understand them. Only MSYS python can work with UNIX paths and properly translate them.

mborgerson:
Thanks @muemart !

elieux:
As Alexey says, mingw-w64-python doesn't understand POSIX paths.  I suggest using [`ntldd`](https://github.com/LRN/ntldd) instead of Cygwin `ldd`.  This idea exchanges the Windows-specific dependency on cygpath with another Windows-specific dependency on ntldd, but IMO that's more reasonable.

PR #83, opened 2018-07-13
Properly check for epoxy

### LINKED ISSUE #67: "ERROR: sizeof(size_t) doesn't match GLIB_SIZEOF_SIZE_T." during compilation
This is a wrong error description introduced by #61 .

**It just means that libepoxy was not found during compilation:**

You can solve this issue by installing it, by following the steps at https://xqemu.com/getting-started/#building-xqemu-from-source which contain the libepoxy dependency.

---

Someone should investigate why this happen and make the error message more descriptive, so we can close this issue.

muemart:
In the configure script, check if epoxy is installed. Fixes #67. Pretty bare-bones, but does the job.

### DIFF
--- a/configure
+++ b/configure
@@ -2962,7 +2962,7 @@ if test "$linux" = "yes" ; then
 fi
 
 ##########################################
-# Epoxy
+# epoxy probe
 
 cat > $TMPC <<EOF
 #include <epoxy/gl.h>
@@ -2970,7 +2970,8 @@ int main(void) { return 0; }
 EOF
 epoxy_libs="-lepoxy"
 if ! compile_prog "" "$epoxy_libs" ; then
-    error_exit "epoxy check failed. Please install epoxy/libepoxy with your package manager and try again."
+    error_exit "epoxy check failed" \
+        "Make sure to have the epoxy libs and headers installed."
 fi
 LIBS="$LIBS $epoxy_libs"
 
--- a/configure
+++ b/configure
@@ -848,7 +848,7 @@ if test "$mingw32" = "yes" ; then
   QEMU_CFLAGS="-D__USE_MINGW_ANSI_STDIO=1 $QEMU_CFLAGS"
   # MinGW needs -mthreads for TLS and macro _MT.
   QEMU_CFLAGS="-mthreads $QEMU_CFLAGS"
-  LIBS="-lwinmm -lws2_32 -liphlpapi -lopengl32 -lepoxy -lgdi32 $LIBS"
+  LIBS="-lwinmm -lws2_32 -liphlpapi -lopengl32 -lgdi32 $LIBS"
   write_c_skeleton;
   if compile_prog "" "-liberty" ; then
     LIBS="-liberty $LIBS"
@@ -2948,7 +2948,7 @@ fi
 
 # Xbox hack to include ogl framework
 if test "$darwin" = "yes" ; then
-    sdl_libs="$sdl_libs -framework OpenGL -lepoxy"
+    sdl_libs="$sdl_libs -framework OpenGL"
 #     cat > $TMPC << EOF
 # #include <OpenGL/gl.h>
 # #include <OpenGL/CGLTypes.h>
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -2958,9 +2958,22 @@ if test "$darwin" = "yes" ; then
 fi
 
 if test "$linux" = "yes" ; then
-    sdl_libs="$sdl_libs -lepoxy -lGLU -lGL"
+    sdl_libs="$sdl_libs -lGLU -lGL"
 fi
 
+##########################################
```
mborgerson: What about cflags?

mborgerson:
Thanks for helping to clean this up!

A bit of verbose background info here: the `./configure` script will do a check for libepoxy when "OpenGL support" is enabled, a feature which is confusingly named in the context of XQEMU because we actually *do* use OpenGL. The "OpenGL support" feature in original QEMU context is related to the UI display code which we currently do not use OpenGL for, but we do use OpenGL for offscreen rendering of 3D graphics (which the framebuffer is then *eventually* displayed on the "non-GL" UI window). Soon we will be using OpenGL for the UI code as well, so stay tuned for that, I have been working on this!

So because we don't actually enable/use this *yet*, in the mean time I have been just conveniently hacking this on to the end of the SDL2 library dependencies which we (XQEMU) always expect to build with. Note that upstream QEMU supports multiple UI implementations like SDL, GTK, and Cocoa--so building for one of these UI types would break our hacky epoxy addition.

However, independent of the UI frontend, we always require the epoxy library in order to resolve our OpenGL functions in the core nv2a graphics code. So for that reason, I do think this addition is appropriate for the time being--but may need some additional tweaks before being merged.

muemart:
Cool, thanks for the info. I was wondering why the opengl option was disabled, because epoxy is included there too.

JayFoxRox:
Thanks for also looking into this!

I was going to point out what @mborgerson also said. But I'm fine with the change as-is.

I also hope you can fixup this to be sightly cleaner: Basically the comment style and wording of the error message is non-standard. Please follow the comment and messag style established by the other tests.

I did not test this yet, but will try it in the next days (this is low-priority for me, because it only fixes the error message).

muemart:
Okay, I adjusted the wording to match the rest of the file more.

configure
```diff
@@ -2962,15 +2962,16 @@ if test "$linux" = "yes" ; then
...
 
 cat > $TMPC <<EOF
 #include <epoxy/gl.h>
 int main(void) { return 0; }
 EOF
 epoxy_libs="-lepoxy"
 if ! compile_prog "" "$epoxy_libs" ; then
-    error_exit "epoxy check failed. Please install epoxy/libepoxy with your package manager and try again."
+    error_exit "epoxy check failed" \
+        "Make sure to have the epoxy libs and headers installed."
```
mborgerson: This is better but I think can be more succinct. Here's an example from the pixman check which we should follow:

```
  error_exit "pixman >= 0.21.8 not present." \
      "Please install the pixman devel package."
```

So basically replace `pixman` with epoxy, and you can probably drop the `>= 0.21.8` part.

JayFoxRox:
Reviewed again and tested on Linux by temporarily renaming lib and / or include folder.

I agree what mborgerson said about the message, but the current one is also fine for me (nit: still nice if it was the "devel" style message).

I don't care so much for pkg-config myself.

So as long as there are no more logic changes in this, this has my blessing: LGTM.
It's up to @mborgerson now.

mborgerson:
I'm happy to merge once the error message and pkg-config changes I requested are made @muemart

muemart:
Made the appropriate changes to use pkg-config to find epoxy like pixman.

mborgerson:
Looks good to me, thanks. @JayFoxRox wanted to test it once more after switching to pkg-config, so I'll leave approval and "Squash and merge" up to him then.

JayFoxRox:
Tested on Linux and Windows.

I noticed that it still gives the poor message when the include path is missing on linux, but I figured most people will either have lib+include+pkg-config, or have none of it.

Merged.

---

Thanks for the contribution @muemart 

Also [consider joining us on Discord or IRC](https://github.com/XboxDev/XboxDev#want-to-help-contact-us) (see blue header for all links).

Issue #91, opened 2018-07-24
Some size-independent MMIO access does not work as intended


JayFoxRox:
QEMUs MMIO design is quite horrible. Let me provide an example MMIO interface with address, type and use:

0x01000000: `uint8_t control_a`
0x01000001: `uint8_t control_b`
0x01000002: `uint16_t control_c`

The issue is that writing 32 bit to 0x01000000 will only trigger one MMIO callback. As the size is wrong for `control_a` some code might choose to ignore this. In other cases, the size is not checked, and `control_a` gets a value which is far too big for it to handle.

Also, given the example from before, there is no additional callback for access to `control_b` and `control_c` as the code often looks like:

```c
switch(address) {
  case CONTROL_A:
    if (size == 1) { ... }
    break; 
  case CONTROL_B:
    ...
}
```

*(Callback is only executed for `address == CONTROL_A`)*

---

The first time I've encountered this was in AC97 when trying to use OpenXDK.
The OpenXDK audio code does access more than 1 register at a time atomically by writing 32 bits at once (almost exactly like my example above).
The AC97 issue in particular should be fixed upstream, as it also affects QEMU.

However, I'm sure we have other cases in our own Xbox specific code too.

These problems are insanely hard to track and we should always at least `assert` for the expected size.
I did not look for other existing solutions to this problem.
It's possible that QEMU already has some way of dealing with this somehow.
My workaround for AC97 (not implemented in XQEMU at time of writing), was to manually loop over all affected registers.

*The reason we don't see this problem more often is because most operating systems have only one function to access such standard hardware registers. So implementing only one way to access a register is fine most of the time.*

PatrickvL:
86box handles this by doing all MMIO at byte granularity

MoochMcGee:
@PatrickvL Only for the nVidia cards and only for reads. 86Box actually deals with this by having up to 6 MMIO handlers. One is 8-bit reads, one is 8-bit writes, one is 16-bit reads, and so on.

Issue #38, opened 2018-06-25
Grand Theft Auto San Andreas Assertion Failed


gandalfthewhite19890404:
I have got exception after pressing Start:
MinGW Runtime Assertion
Assertion failed!
Program: D:\xb\xqemu2.0\xqemu.exe
C:/projects/xqemu-c5j6o/hw/xbox/dsp/dsp_cpu.c, Line 1018
Expression: address < DSP_XRAM_SIZE

Tested on latest 5d4058d, tested with bat-file, not xqemu manager (with xqemu-manager it stucks after copyright screen).
Here is bat-file content, maybe it can help to fix xqemu-manager:
`xqemu.exe -cpu pentium3 -machine xbox,short_animation,bootrom=mcpx.bin -bios bios.bin -m 64 -drive index=0,media=disk,file=hdd.qcow2,locked -drive index=1,media=cdrom,file=gtasa.iso -usb -device usb-xbox-gamepad`

JayFoxRox:
Please always provide information which commit / revision of XQEMU is used.
Otherwise ` File: C:/projects/xqemu-c5j6o/hw/xbox/dsp/dsp_cpu.c, Line 1018` (which is the most important part) does not help us.
Thanks!

gandalfthewhite19890404:
Latest 5d4058d, tested with bat-file, not xqemu manager (with xqemu-manager it stucks after copyright screen).

JayFoxRox:
Looks like this was solved, as reported in https://github.com/xqemu/xqemu/pull/92#issuecomment-408762985

Closed.

PR #92, opened 2018-07-24
Implement support for APU DSP Memory access (fix EP, add GP)

JayFoxRox:
This is required to read-back results of MCPX APU DSP calculations more easily, so without this change, debugging the DSP from the outside is a lot harder.

This is code from 2017, but it has gotten a lot of additional unit testing and research on actual hardware the past weeks.

Once this patch is applied, we can use xboxpy and my existing dsp-homebrew script to send custom code to the DSP. This allows us to verify instruction behaviour, as the script also works with a real Xbox.

I was unsure about the DSP memory sizes, but then decided to confirm them:

<details><summary>I have written a program to measure the DSP data memory sizes (Click to unfold)</summary><br>

Here is the code for measuring Y-Memory size:

```
start
  move y:$<address>, a0
  move a0, x:$000000
  jmp start
```

Where `<address>` was raised using a step-size.
I then wrote a random value to `y:$<address>` using the CPU and checked if it appeared in `x:$000000`. I did this for 2 random values per `<address>`. When any of these checks failed, the previous `<address>` was reverted and a smaller step-size was used.

For measuring the size of X-Memory instead of Y-Memory I have simply swapped `x:` and `y:` (and respective lookups on the CPU).
For measuring the size of the MIXBUF, I set a higher start offset (0x1400) in X-Memory and re-ran the same program as before for measuring X-Size. Strangely, I got 0x7E0 words here. The last 0x20 words were not accessible. See known problem below.

</details>   <details><summary>I also wrote a program to measure the DSP program memory size (Click to unfold)</summary>

Here is the code for measuring P-Memory size:

```
start
  move y:$000000, a0
  <nops>
  move a0, x:$000000
  jmp start
```

Where `<nops>` is an increasing number of `nop` instructions. By raising the number of `nop` instructions, the write to `x:$000000` will eventually be outside of program memory. This will result in either the value in `y:$000000` not updating or the Xbox crashing.
Keep in mind that the other instructions take up some memory (6 words) so the number of `nop` instructions (1 word each) is not the same as the program memory size.
</details><br>

[The results of these tests have been added to XboxDevWiki](http://xboxdevwiki.net/DSP#Memory_Size).

According to existing code, the last 0x400 words of GP X-Memory (so, offset 0xC00) should be MIXBUF. My tests could not confirm this behaviour. I have observed MIXBUF at `x:$001400`, 0x380 words (See problem regarding size below).
I assume that DirectSound uses the last 0x400 words of X-Memory RAM for the DMA from the actual MIXBUF; but X-Memory RAM has no relation to MIXBUF in hardware.

#### Only 32 bit access works

Also see #91.

This might cause trouble in games, so this PR should be tested with a handful of games.
I did not test a single game yet.

#### MIXBUF size not known

I was unable to access the last 0x20 words of MIXBUF with my unit test.

I assume this might have to do with VP or EP having control and overwriting my data. I still assume a size of 0x400 bytes, not 0x380; so I implemented 0x400 in XQEMU, but documented 0x380 for now. To be fair, I was surprised that I had write access to MIXBUF at all, as I had assumed the VP would instantly overwrite data.

#### EP and GP use the same memory map in XQEMU

Memory size defines are used by both DSP instances, so we use the same memory map for GP and EP.

This is not so much a problem with this PR, but XQEMU in general. This even means that the EP currently has MIXBUF access which is not possible on real hardware.

I'll create an issue after merge. I did not fix it here because it leads to design decisions and what we have works (our EP is just more powerful than on real hardware).

---

This patch *could* fix bugs in the following games:

* Top Gear RPM Tuning
* Grand Theft Auto: San Andreas (#38)

These games previously ran into the following issue in dsp_cpu: `assert(address < DSP_XRAM_SIZE);`.

However, I did not test those games because it's not the motivation for this patch: I just want to be able to debug the DSP.

---

If you want to [test this PR with my DSP homebrew](https://github.com/JayFoxRox/xbox-tools/tree/master/python-scripts): nxdk-rdt still does not support `CALL` in master yet (meaning it can't remote control the DSP using existing code).
You will need to look for nxdk-rdt and xboxpy PRs to work around this.
Testing using XBDM should work fine.


hw/xbox/mcpx_apu.c
```diff
@@ -471,9 +502,31 @@ static void gp_write(void *opaque, hwaddr addr,
 {
     MCPXAPUState *d = opaque;
 
+    assert(size == 4);
+
     MCPX_DPRINTF("mcpx apu GP: [0x%llx] = 0x%llx\n", addr, val);
 
     switch (addr) {
+    case NV_PAPU_GPXMEM ... NV_PAPU_GPXMEM + 0xFFF*4: {
```
dracc: I'd prefer if you defined the different memregion sizes/offsets as constants.

JayFoxRox: I'm following the style that has been established in the existing code.

I suggest we revise this when addressing what I described in "EP and GP use the same memory map in XQEMU": if we split EP and GP config, we will also define constants for their respective size probably.

gandalfthewhite19890404:
Here is screenshot with this PR:
https://cdn.discordapp.com/attachments/430847916857622539/471565766299222016/unknown.png

hw/xbox/dsp/dsp.c
```diff
@@ -198,27 +198,47 @@ uint32_t dsp_disasm_address(DSPState* dsp, FILE *out, uint32_t lowerAdr, uint32_
     return dsp_pc;
 }
 
-uint32_t dsp_read_memory(DSPState* dsp, char space_id, uint32_t address)
+uint32_t dsp_read_memory(DSPState* dsp, char space, uint32_t address)
```
mborgerson: Ok with the change but in general I think these proxy functions are kind of wonky.

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -972,8 +972,6 @@ uint32_t dsp56k_read_memory(dsp_core_t* dsp, int space, uint32_t address)
             return dsp->read_peripheral(dsp, address);
         } else if (address >= DSP_MIXBUFFER_BASE && address < DSP_MIXBUFFER_BASE+DSP_MIXBUFFER_SIZE) {
             return dsp->mixbuffer[address-DSP_MIXBUFFER_BASE];
-        } else if (address >= DSP_MIXBUFFER_READ_BASE && address < DSP_MIXBUFFER_READ_BASE+DSP_MIXBUFFER_SIZE) {
```
mborgerson: Okay looking at the change to the base address, this is now covered by `DSP_MIXBUFFER_BASE`, correct?

hw/xbox/mcpx_apu.c
```diff
@@ -461,7 +465,34 @@ static uint64_t gp_read(void *opaque,
 {
     MCPXAPUState *d = opaque;
 
-    uint64_t r = d->gp.regs[addr];
+    assert(size == 4);
```
mborgerson: Let's be safe here and also `assert` that the access is aligned... e.g.:
```
assert(addr & 3 == 0)
```

hw/xbox/mcpx_apu.c
```diff
@@ -461,7 +465,34 @@ static uint64_t gp_read(void *opaque,
...
+
+    uint64_t r = 0;
+    switch (addr) {
+    case NV_PAPU_GPXMEM ... NV_PAPU_GPXMEM + 0xFFF*4: {
+        uint32_t xaddr = (addr - NV_PAPU_GPXMEM) / 4;
```
mborgerson: Prefer shift operation over division, but compiler should automatically do that for us right?

hw/xbox/mcpx_apu.c
```diff
@@ -461,7 +465,34 @@ static uint64_t gp_read(void *opaque,
...
+        r = dsp_read_memory(d->gp.dsp, 'Y', yaddr);
+        break;
+    }
+    case NV_PAPU_GPPMEM ... NV_PAPU_GPPMEM + 0xFFF*4: {
+        uint32_t paddr = (addr - NV_PAPU_GPPMEM) / 4;
+        r = dsp_read_memory(d->gp.dsp, 'P', paddr);
+        break;
+    }
+    default:
+        r = d->gp.regs[addr];
```
mborgerson: Prefer to `assert` here that addr is not out of bounds

hw/xbox/mcpx_apu.c
```diff
@@ -471,9 +502,31 @@ static void gp_write(void *opaque, hwaddr addr,
 {
     MCPXAPUState *d = opaque;
 
+    assert(size == 4);
```
mborgerson: Same comment as before w.r.t. alignment.

hw/xbox/mcpx_apu.c
```diff
@@ -527,19 +581,24 @@ static void ep_write(void *opaque, hwaddr addr,
...
-    case NV_PAPU_EPYMEM ... NV_PAPU_EPYMEM + 0x100 * 4: {
-        assert(false);
+    case NV_PAPU_EPYMEM ... NV_PAPU_EPYMEM + 0xFF*4: {
+        uint32_t yaddr = (addr - NV_PAPU_EPYMEM) / 4;
+        dsp_write_memory(d->ep.dsp, 'Y', yaddr, val);
         break;
     }
-    case NV_PAPU_EPPMEM ... NV_PAPU_EPPMEM + 0x1000 * 4: {
-        assert(false);
+    case NV_PAPU_EPPMEM ... NV_PAPU_EPPMEM + 0xFFF*4: {
```
mborgerson: These offsets strike me as odd... These implicit switch cases are inclusive right? Should it not be `NV_PAPU_EPPMEM ... NV_PAPU_EPPMEM + 0x1000*4-1`

hw/xbox/dsp/dsp.c
JayFoxRox: Agreed. I don't like them either. For simplicity I kept them in this PR.

I hope that we at least get rid of the 'X, 'Y', 'P' or the constants in the future; the mapping is really stupid. However, I suggest doing that in a cleanup later. For now, the focus is on having at least a rough skeleton for research, debugging and actual audio support.

hw/xbox/dsp/dsp_cpu.c
JayFoxRox: This code was broken to begin with: Seemingly based on how DirectSound named these regions in software; not how the hardware worked.

With DirectSound it appears that there's a mirror because it's being DMA'd all the time: from the 0x400 bytes of MIXBUF (x:001400; Previously DSP_MIXBUFFER_READ_BASE) to the last 0x400 bytes of X RAM (x:000C00; Previously DSP_MIXBUFFER_BASE).

My tests suggest that in hardware, there's a 0x400 MIXBUF (x:001400; Now DSP_MIXBUFFER_BASE) and the memory at x:0x000C00 is just ordinary X RAM - it has no special link to MIXBUF (but in DirectSound they are linked by setting up a DMA transfer for each audio frame).

Also see PR description.

hw/xbox/mcpx_apu.c
JayFoxRox: Will do, but this will delay the PR as we have to test at least a handful of games.

hw/xbox/mcpx_apu.c
JayFoxRox: Yes. The generated code will not do a complicated division.

hw/xbox/mcpx_apu.c
JayFoxRox: Well.. it's kind-of the same. As we assert 32 bit access it makes no difference (actually the `- 1` solution could write more data than expected then).

I just modified existing code, but I also knew we do this same sort of stuff in the GPU:

```c
case NV097_SET_VERTEX4F ... NV097_SET_VERTEX4F + 12: {
```

*(where 4 values = 16 bytes, can be passed to NV097_SET_VERTEX4F. So the formula is `(4-1)*4`)*

I can do the `<size> * 4 - 1` if you want that.

Please tell me if you want this changed. I personally think we should introduce macros at some point in the future and do an overall cleanup.

hw/xbox/mcpx_apu.c
JayFoxRox: This was copied from the existing `gp_write` / `ep_read`. Neither of those do any checks.

Let's do it in a cleanup later. I don't think this will happen anyway.

hw/xbox/mcpx_apu.c
mborgerson: In this case it may not cause an issue because the factor is 4, but for larger values this certainly would cause a problem. Consider if instead we had used a factor of 8, then we would lose the ability to address the last dword.

For correctness and readability, I'd prefer it (for now) if we did:

```c
case <base> ... <base> + <size> * 4 - 1:
```

But yes, I do agree that at some point we should probably cleanup this switch stuff and use a macro or some table lookup thing.

hw/xbox/mcpx_apu.c
mborgerson: Please and thank you!

JayFoxRox:
I've reviewed the code and asked people to test it for accidental triggers of `assert`.

It was tested with:

- Midtown Madness 3 
- Fallout Brotherhood of steel
- Timesplitters Future Perfect
- Call of Duty 3
- Silent Hill 4
- Broken Sword 3 
- Evox
- nxdk-rdt

*Thanks to @Ale100000 who tested some of these games.*

---

If any of the `assert` is triggered and this causes regressions, we can probably fix it easily anyway.
We have also tested first-party and third-party games.

I believe we are ready to merge now @mborgerson. Thanks!

gandalfthewhite19890404:
I confirm - GTA San Andreas works as should (slow but I think - its not issue, many games are very slow).

mborgerson:
Thanks!

PR #97, opened 2018-08-02
ui: Stop appending "-dirty" to clean AppVeyor builds

### LINKED ISSUE #93: Clean builds showing "-dirty" in title bar
The window title bar which shows the revision also shows "-dirty" if the tree has been modified. However, for clean appveyor builds it's also showing up as "-dirty".

dracc:
Closes #93 .

This is a fix for the Makefile bug which led to AppVeyor builds showing the revision as dirty.

Reasoning for switching from `git diff-index` to `git diff` stems from how these two operate.
On a fresh cloned repo `git diff-index` has no local index to look at, it just sees fresh files and reports them as new. `git diff` on the other hand actually looks at the files and the upstream tree, finding that the files are unmodified copies downloaded from the repo. It reports no changes and updates the local index.

This updating behaviour leads to `git diff-index` being able to correctly identify unchanged files - this led to my first proposed solution of running `git diff` before running the "dirty-check".

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -443,7 +443,7 @@ qemu-version.h: FORCE
 		else \
 			if test -d .git; then \
 				pkgvers=$$(git rev-parse --short HEAD 2>/dev/null | tr -d '\n');\
-				if ! git diff-index --quiet HEAD &>/dev/null; then \
+				if ! git diff --quiet HEAD >/dev/null 2>&1; then \
 					pkgvers="$${pkgvers}-dirty"; \
 				fi; \
 			fi; \


JayFoxRox:
## DO NOT review or merge yet.

For the record: This PR is currently broken. 
It's currently being fixed.

dracc:
I've tested piping output of `git diff-index` into a text file before running the actual versioning part, resulted in 6125 lines.
I've also tested with clean & dirty repos locally, works as expected.
I've tested injecting a comment into `Makefile` before versioning, resulted in "-dirty" being appended.
I'd say it's somewhat complete.

Also edited piping from "`&>/dev/null`" to posix-compliant "`>/dev/null 2>&1`", as requested upstream.

Qemu mailing list regarding this change [here](https://lists.gnu.org/archive/html/qemu-devel/2018-08/msg00450.html).

JayFoxRox:
Thanks for looking into this.

I'll wait until upstream QEMU feedback arrives. If that doesn't happen within a week I'll just review independently.

MayeulC:
As far as I know, `git diff` is a porcelain command; which might be subjected to syntax changes. A lower-level approach might be preferable, using some plumbing commands. Running `git update-index` (with `--refresh`?) might update the index and fix this issue.

It could be ran unconditionally, or only if it was deemed dirty:
`git diff-index --quiet HEAD &>/dev/null || (git update-index; git diff-index --quiet HEAD &>/dev/null)`

mborgerson:
Any updates on this PR?

dracc:
Adding `-p` does not solve the problem, "-dirty" is still appended on AppVeyor builds.

Running the command `git update-index` before the check did not seem to do anything useful - "-dirty" still appended.

mborgerson:
According to @dracc on Discord:

> Removing "-index" gives expected results both locally and on appveyor, yes.

Approved. Thanks!

PR #102, opened 2018-08-18
nv2a: Fix EXP and LOG

dracc:
THPS2X is broken, character is stretched until it disappears.
It seems this behaviour is due to EXP and LOG being broken.
This piece of code provided by @JayFoxRox fixes that.


hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -674,13 +674,24 @@ static const char* vsh_header =
     "#define EXP(dest, mask, src) dest.mask = _EXP(_in(src).x).mask\n"
     "vec4 _EXP(float src)\n"
     "{\n"
-    "  return vec4(exp2(src));\n"
+    "  vec4 result;\n"
+    "  result.x = exp2(floor(src));\n"
+    "  result.y = src - floor(src);\n"
+    "  result.z = exp2(src);\n"
```
mborgerson: Is this correct? Upon a cursory glance, `result.z` doesn't seem to match what is in the [spec [2.14.1.10.15]](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_vertex_program.txt). It looks like it should be something like:

```c
result.z = result.x * APPX(result.y);

"APPX is an implementation dependent approximation of exponential base 2 such that | exp(q.y*log(2.0))-APPX(q.y) | < 1/(2^11)"
```

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -674,13 +674,24 @@ static const char* vsh_header =
...
+    "  result.w = 1.0;\n"
+    "  return result;\n"
     "}\n"
     "\n"
     "#define LOG(dest, mask, src) dest.mask = _LOG(_in(src).x).mask\n"
     "vec4 _LOG(float src)\n"
     "{\n"
-    "  return vec4(log2(src));\n"
+    "  float tmp = abs(src);\n"
```
mborgerson: Should we specifically handle the case when `src == 0.0f` as the spec says?

mborgerson:
Please see inline review comments. Additionally in the Git commit message, it would be useful to provide a short reference to the [the spec](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_vertex_program.txt) saying why the change is necessary.

mborgerson:
Thanks @dracc, I've finished my review and it looks good to me. I'll let @JayFoxRox do a final review and merge here.

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -674,13 +674,25 @@ static const char* vsh_header =
...
+    "  if (tmp == 0.0) { return vec4(-INFINITY, 1.0f, -INFINITY, 1.0f); }\n"
```
mborgerson: nit: prefer newline indentation here

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -674,13 +674,25 @@ static const char* vsh_header =
     "#define EXP(dest, mask, src) dest.mask = _EXP(_in(src).x).mask\n"
     "vec4 _EXP(float src)\n"
     "{\n"
-    "  return vec4(exp2(src));\n"
+    "  vec4 result;\n"
+    "  result.x = exp2(floor(src));\n"
+    "  result.y = src - floor(src);\n"
```
JayFoxRox: I've mentioned this on IRC before; from 2.14.1.11 (Vertex Program Floating Point Requirements):

> The floor operations used by the ARL and EXP instructions must
> operate identically.  Specifically, the EXP instruction's floor(t.x)
> intermediate result must exactly match the integer stored in the
> address register by the ARL instruction.

Also see https://github.com/xqemu/xqemu/pull/79 for our current hack, which breaks this assumption.
Although I can't think of an example algorithm, having these in sync could be very critical, so we should respect this (there's probably good a reason why the spec is so specific).

I suggest adding our own `biased_floor` function or global constant for the bias (maybe pick a better name which respects our GLSL style), and using it in ARL and EXP.

Please also think about cases where our bias might run into problems (I worry about the result.y line because it might break assumption about returning [0.0;1.0[ ), so we can review the use of a bias or change it's value to avoid those problems.
So if it would break assumptions, we might have to add the bias to `src` for *all* operations in EXP.

I'm also fine with just documenting this as an issue if we can't find a good solution to integrate the bias.

hw/xbox/nv2a/nv2a_vsh.c
JayFoxRox: For the record: We don't handle the 0, INF, NAN cases anywhere right now because they are not as easy as it seems.

GLSL and NV VP floats differ in how those are treated, so by handling it, it's quite easy to accidentally introduce more errors. Look into NAN representation, FTZ and other things in the NV VP spec.

This is also one of the reasons I said I want to move VP into a CPU JIT in the future.

I'm fine with handling the 0.0 case here, but keep in mind that the proposed code is probably not a complete solution yet.

hw/xbox/nv2a/nv2a_vsh.c
```diff
@@ -674,13 +674,25 @@ static const char* vsh_header =
...
     "\n"
     "#define LOG(dest, mask, src) dest.mask = _LOG(_in(src).x).mask\n"
     "vec4 _LOG(float src)\n"
     "{\n"
-    "  return vec4(log2(src));\n"
+    "  float tmp = abs(src);\n"
+    "  if (tmp == 0.0) { return vec4(-INFINITY, 1.0f, -INFINITY, 1.0f); }\n"
+    "  vec4 result;\n"
+    "  result.x = floor(log2(tmp));\n"
+    "  result.y = tmp / exp2(floor(log2(tmp)));\n"
```
JayFoxRox: When adding the bias in EXP, we might have to undo it here.
Would be nice if someone thought about the consequences and reported back what could go wrong here.

JayFoxRox:
There's lots of room for improvement still (even aside from missing features like INF, NAN, 0.0 handling), but this is way better than master.

Merging, and moving some of my review comments to a new issue.

Issue #28, opened 2018-06-23
EEPROM is hardcoded in source code


JayFoxRox:
We currently hardcode a EEPROM image in the source code which is loaded into the virtual EEPROM at startup.
Ideally we'd store the EEPROM contents in a file which is read and written at runtime.

dracc:
This looks like an easy task, but where is said EEPROM located?

mborgerson:
To implementer: Add a prop on machine args for `eeprom` like we do with `bootrom`. If not specified use default EEPROM that is in the tree now.

JayFoxRox:
My personal preference would be to get rid of the EEPROM from code (and add it to the HLE later if necessary), and then include it on the same repo as the HDD dummy image (repo should probably be renamed).

I'd accept the other solution for usability though.
But my thought is users have to copy files around anyway, might as well have them copy both + simplify our codebase.

mborgerson:
I'd argue that the code wouldn't really be any simpler (less two tiny array definitions), and it would require yet *another* file to run XQEMU with no real added benefit. It doesn't really hurt to just leave it in for now and provide the option to pass one in.

dracc:
I'd like to leave it in but maybe add a few functions to enable modifying it - shouldn't be too hard right?

MayeulC:
I would rather have a json or other format file that is parsed to generate an eeprom, since the file is well documented:
http://xboxdevwiki.net/EEPROM

All we need is the eeprom_key that is used to encrypt the first portion, which can be extracted from somewhere (not entierely clear to me, probably the kernel, which needs the mcpx to be decrypted): http://xboxdevwiki.net/Boot_Process#Kernel_decryption

So, maybe `0x00-0x2F` can be hardcoded for now, and the other fields dynamically generated. Not many fields are that interesting for now, but maybe:
* Language ID
* Video Settings
* Video Standard

That should probably be user-configurable.

Further down the road, HDD key, MAC address and maybe online key could be dynamically generated using a system-dependent method (host mac address?) or account system.

JayFoxRox:
@MayeulC we are LLE, and we should want to keep binary images around.
Modifying the EEPROM is not XQEMUs mission statement and it's an offtopic task.
Most of your requests are HLE heavy.

If we are nice, we provide options in the GUI for it (especially as we are heading towards HLE and can't expect users to have a working dashboard). However, other than that, it should be a [seperate project](https://github.com/JayFoxRox/xbox-tools/pull/46).
[...]

PR #107, opened 2018-09-04
Read EEPROM from file

dracc:
Touches #28, but does not quite go the full distance.
Code is more or less @mborgerson's bootrom work, just refitted for this purpose.

What works:
- [x] Load EEPROM from file
- [x] Use bunnies EEPROM if file not supplied

### DIFF
--- a/hw/xbox/xbox.c
+++ b/hw/xbox/xbox.c
@@ -227,11 +227,62 @@ static void xbox_memory_init(PCMachineState *pcms,
     xbox_flash_init(rom_memory);
 }
 
+uint8_t *load_eeprom(void)
+{
+    char *filename;
+    int fd;
+    int rc;
+    int eeprom_file_size;
+    const int eeprom_size = 256;
+
+    uint8_t *eeprom_data = g_malloc(eeprom_size);
+
+    const char *eeprom_file = object_property_get_str(qdev_get_machine(),
+                                                      "eeprom", NULL);
+    if ((eeprom_file != NULL) && *eeprom_file) {
+        filename = qemu_find_file(QEMU_FILE_TYPE_BIOS, eeprom_file);
+        assert(filename);
+
+        eeprom_file_size = get_image_size(filename);
+        if (eeprom_size != eeprom_file_size) {
+            fprintf(stderr,
+                    "qemu: EEPROM file size != %d bytes. (Is %d bytes)\n",
+                    eeprom_size, eeprom_file_size);
+            g_free(filename);
+            exit(1);
+            return NULL;
+        }
+
+        fd = open(filename, O_RDONLY | O_BINARY);
+        if (fd < 0) {
+            fprintf(stderr, "qemu: EEPROM file '%s' could not be opened.\n", filename);
+            g_free(filename);
+            exit(1);
+            return NULL;
+        }
+
+        rc = read(fd, eeprom_data, eeprom_size);
+        if (rc != eeprom_size) {
+            fprintf(stderr, "qemu: Could not read the full EEPROM file.\n");
+            close(fd);
+            g_free(filename);
+            exit(1);
+            return NULL;
+        }
+
+        close(fd);
+        g_free(filename);
+    } else {
+        memcpy(eeprom_data, default_eeprom, eeprom_size);
+    }
+    return eeprom_data;
+}
 
 /* PC hardware initialisation */
 static void xbox_init(MachineState *machine)
 {
-    xbox_init_common(machine, default_eeprom, NULL, NULL);
+    uint8_t *eeprom_data = load_eeprom();
+    xbox_init_common(machine, eeprom_data, NULL, NULL);
 }
 
 void xbox_init_common(MachineState *machine,
@@ -410,6 +461,22 @@ static void machine_set_bootrom(Object *obj, const char *value,
     ms->bootrom = g_strdup(value);
 }
 
+static char *machine_get_eeprom(Object *obj, Error **errp)
+{
+    XboxMachineState *ms = XBOX_MACHINE(obj);
+
+    return g_strdup(ms->eeprom);
+}
+
+static void machine_set_eeprom(Object *obj, const char *value,
+                                        Error **errp)
+{
+    XboxMachineState *ms = XBOX_MACHINE(obj);
+
+    g_free(ms->eeprom);
+    ms->eeprom = g_strdup(value);
+}
+
 static void machine_set_short_animation(Object *obj, bool value,
                                         Error **errp)
 {
@@ -430,6 +497,11 @@ static inline void xbox_machine_initfn(Object *obj)
                             machine_set_bootrom, NULL);
     object_property_set_description(obj, "bootrom",
                                     "Xbox bootrom file", NULL);
+
+    object_property_add_str(obj, "eeprom", machine_get_eeprom,
+                            machine_set_eeprom, NULL);
+    object_property_set_description(obj, "eeprom",
+                                    "Xbox EEPROM file", NULL);
 
     object_property_add_bool(obj, "short-animation",
                              machine_get_short_animation,
--- a/hw/xbox/xbox.h
+++ b/hw/xbox/xbox.h
@@ -28,6 +28,8 @@
 
 #define MAX_IDE_BUS 2
 
+uint8_t *load_eeprom(void);
+
 void xbox_init_common(MachineState *machine,
                       const uint8_t *eeprom,
                       PCIBus **pci_bus_out,
@@ -47,6 +49,7 @@ typedef struct XboxMachineState {
 
     /*< public >*/
     char *bootrom;
+    char *eeprom;
     bool short_animation;
 } XboxMachineState;
 


hw/xbox/xbox.c
```diff
@@ -253,7 +253,35 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
...
+
+    const char *eeprom_file = object_property_get_str(qdev_get_machine(),
+                                                      "eeprom", NULL);
+    if ((eeprom_file != NULL) && *eeprom_file) {
+        filename = qemu_find_file(QEMU_FILE_TYPE_BIOS, eeprom_file);
+        assert(filename);
+
+        eeprom_size = get_image_size(filename);
+        assert(eeprom_size <= 256);
+        if (eeprom_size < 256) {
```
mborgerson: EEPROM should always be 256 bytes.

hw/xbox/xbox.c
```diff
@@ -253,7 +253,35 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
 /* PC hardware initialisation */
 static void xbox_init(MachineState *machine)
 {
-    xbox_init_common(machine, default_eeprom, NULL, NULL);
+    char *filename;
```
mborgerson: This works for Xbox machine but would have to be duplicated for Chihiro support. We should think of a common solution so both can support it via the same code.

hw/xbox/xbox.c
```diff
@@ -253,7 +253,35 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
...
+    if ((eeprom_file != NULL) && *eeprom_file) {
+        filename = qemu_find_file(QEMU_FILE_TYPE_BIOS, eeprom_file);
+        assert(filename);
+
+        eeprom_size = get_image_size(filename);
+        assert(eeprom_size == 256);
+
+        eeprom_data = g_malloc(eeprom_size);
+        fd = open(filename, O_BINARY);
+        assert(fd >= 0);
```
mborgerson: I know the bootrom code does this exact same thing, but I think (in both cases) we should fail more gracefully here and print an error message saying that the file could not be opened. It's very possible that we don't have permission or something.

hw/xbox/xbox.c
```diff
@@ -253,7 +253,35 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
...
+    uint8_t *eeprom_data;
+
+    const char *eeprom_file = object_property_get_str(qdev_get_machine(),
+                                                      "eeprom", NULL);
+    if ((eeprom_file != NULL) && *eeprom_file) {
+        filename = qemu_find_file(QEMU_FILE_TYPE_BIOS, eeprom_file);
+        assert(filename);
+
+        eeprom_size = get_image_size(filename);
+        assert(eeprom_size == 256);
```
mborgerson: Same as below, we should fail more gracefully. It's highly likely the user will pass some invalid file here and we will just crash.

hw/xbox/xbox.c
```diff
@@ -253,7 +253,35 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
 /* PC hardware initialisation */
 static void xbox_init(MachineState *machine)
 {
-    xbox_init_common(machine, default_eeprom, NULL, NULL);
+    char *filename;
+    int fd;
+    int eeprom_size = 256;
```
mborgerson: Make this a const and compare against it below.

hw/xbox/xbox.c
```diff
@@ -253,7 +253,53 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
...
+            g_free(filename);
+            exit(1);
+            return;
+        }
+
+        close(fd);
+        g_free(filename);
+    } else {
+        memcpy(eeprom_data, default_eeprom, eeprom_size);
+    }
```
JayFoxRox: I think the complexity of this code section warrants putting it in a separate function (~`load_eeprom` or something)

hw/xbox/xbox.c
```diff
@@ -253,7 +253,53 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
...
+    char *filename;
+    int fd;
+    int rc;
+    int eeprom_file_size;
+    const int eeprom_size = 256;
+    uint8_t *eeprom_data = g_malloc(eeprom_size);
+
+    const char *eeprom_file = object_property_get_str(qdev_get_machine(),
+                                                      "eeprom", NULL);
+    if ((eeprom_file != NULL) && *eeprom_file) {
```
JayFoxRox: nit: is `*eeprom_file` really necessary and the QEMU way of doing things?

hw/xbox/xbox.c
```diff
@@ -253,7 +253,53 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
...
+        filename = qemu_find_file(QEMU_FILE_TYPE_BIOS, eeprom_file);
+        assert(filename);
+
+        eeprom_file_size = get_image_size(filename);
+        if (eeprom_size != eeprom_file_size) {
+            fprintf(stderr,
+                    "qemu: EEPROM file size != 256 bytes. (Is %i bytes)\n",
```
JayFoxRox: Use `"...%d...", ..., eeprom_size)`

hw/xbox/xbox.c
mborgerson: Agree (what I was hinting at with above comment too)

hw/xbox/xbox.c
mborgerson: Yes, it is necessary. What are you suggesting?

hw/xbox/xbox.c
```diff
@@ -253,7 +253,53 @@ void nv2a_init(PCIBus *bus, int devfn, MemoryRegion *ram);
...
+                    eeprom_file_size);
+            g_free(filename);
+            exit(1);
+            return;
+        }
+
+        fd = open(filename, O_BINARY);
```
JayFoxRox: nit: Open as read-only?

JayFoxRox:
- Commits should be renamed to prefix style ("eeprom: ..." or something)
- EEPROM image should be provided somewhere (this should happen before this is merged, so we get closer to getting rid of the built-in image)

I think write can happen in a follow-up PR.
I always considered this PR WIP because it shows 2/3 tasks complete, but it looks like you are not actively working on this anymore.

hw/xbox/xbox.c
JayFoxRox: `strlen(eeprom_file) > 0` or something?

I think it's just not very self-documenting.

mborgerson: Ah yes, agree

hw/xbox/xbox.c
```diff
@@ -150,7 +150,7 @@ static void xbox_flash_init(MemoryRegion *rom_memory)
     const char *bootrom_file = object_property_get_str(qdev_get_machine(),
                                                        "bootrom", NULL);
 
-    if ((bootrom_file != NULL) && *bootrom_file) {
+    if ((bootrom_file != NULL) && *botrom_file) {
```
mborgerson: ?

dracc: I made the edit from my cellphone, this change was unintended. Will fix asap

hw/xbox/xbox.c
```diff
@@ -227,11 +227,62 @@ static void xbox_memory_init(PCMachineState *pcms,
...
+    int fd;
+    int rc;
+    int eeprom_file_size;
+    const int eeprom_size = 256;
+
+    uint8_t *eeprom_data = g_malloc(eeprom_size);
+
+    const char *eeprom_file = object_property_get_str(qdev_get_machine(),
+                                                      "eeprom", NULL);
+    if ((eeprom_file != NULL) && (strlen(eeprom_file) > 0)){
```
JayFoxRox: Please undo this change - the original code was more similar to existing code (the `bootrom_file`).
If `*eeprom_file` is the standard way of doing it, we should keep doing so (or change *all* instances at once - which doesn't fit into this PR).

hw/xbox/xbox.c
```diff
@@ -227,11 +227,62 @@ static void xbox_memory_init(PCMachineState *pcms,
     xbox_flash_init(rom_memory);
 }
 
+uint8_t* load_eeprom(void)
```
mborgerson: Nit: `*` goes to the immediate left of the identifier (e.g. `uint8_t *load_eeprom`). Repeated a couple times below.

hw/xbox/xbox.c
```diff
@@ -227,11 +227,62 @@ static void xbox_memory_init(PCMachineState *pcms,
...
+        if (eeprom_size != eeprom_file_size) {
+            fprintf(stderr,
+                    "qemu: EEPROM file size != %d bytes. (Is %d bytes)\n",
+                    eeprom_size, eeprom_file_size);
+            g_free(filename);
+            exit(1);
+            return NULL;
+        }
+
+        fd = open(filename, O_RDONLY);
```
JayFoxRox: nit: might still want the binary flag to not break on shitty platforms (Windows in particular). not sure which `open` variant is exposed here and if QEMU deals with this.

dracc:
Review comments taken care of.
Regarding EEPROM binary distribution; could bunnies EEPROM be distributed in the same vein as the HDD image?

hw/xbox/xbox.c
```diff
@@ -227,11 +227,62 @@ static void xbox_memory_init(PCMachineState *pcms,
...
+                    eeprom_size, eeprom_file_size);
+            g_free(filename);
+            exit(1);
+            return NULL;
+        }
+
+        fd = open(filename, O_RDONLY | O_BINARY);
+        if (fd < 0) {
+            fprintf(stderr, "qemu: EEPROM file '%s' could not be opened.\n", filename);
+            g_free(filename);
```
mborgerson: Not `free`ing `eeprom_data`, but it doesn't really matter here because we are exiting.

mborgerson:
Thanks @dracc

hw/xbox/xbox.c
dracc: Should still be considered a leak, no?

JayFoxRox: Yes, and some cleanup should fix it eventually because it just polutes valgrind logs.

Issue #106, opened 2018-09-04
Appveyor Windows builds appear to be faulty


mborgerson:
People have reported that the last Appveyor build of the master branch that was usable was [667fe85](https://ci.appveyor.com/project/mborgerson/xqemu-c5j6o/build/1.0.90). Subsequent builds from [c01c27d](https://ci.appveyor.com/project/mborgerson/xqemu-c5j6o/build/1.0.96) and up (including pull-requests) have strange issues seemingly related to stack protection.

Interesting bits:
- Program will segfault with dereference of `fs:0`. `fs` not being correctly loaded with canary location?
- See @xbox7887's disassembly notes on Discord (FIXME: Inline here)
- MSYS2 appears to have been upgraded. [See here](https://www.appveyor.com/updates/)
- People have reported success downloading sources and building themselves. We need version numbers though, particularly GCC.

Grovespaz:
I've built successfully locally with GCC 7.3.0 on MSYS2. Doesn't matter if binaries are stripped or not. It doesn't make a difference when I use the dll's from AppVeyor with a locally built xqemu.exe, so the problem does not appear to be in a dependency but the executable itself.

xbox7887's notes start at https://discordapp.com/channels/428359196719972353/428359372842991616/486342982606258188

Ernegien:
I suggest adding `--disable-stack-protector \` after https://github.com/xqemu/xqemu/blob/667fe8536d17d1435d18089477f28d4ffc2036bf/build.sh#L62 as a short-term workaround

GXTX:
I'm running into this issue while attempting to compile myself. MSYS2 was installed from `msys2-x86_64-20180531` then fully updated via their initial update guide.

```
$ pacman -Qe | grep gcc
gcc-libs 7.3.0-3
mingw-w64-x86_64-gcc 8.2.0-1
$ gcc --version
gcc.exe (Rev1, Built by MSYS2 project) 8.2.0
```

Using @Ernegien's additional `configure` variable does produce a working build.

mborgerson:
Resolved by #108. Thank you all for your help with investigation of this issue, and a special thanks to @Ernegien.

PR #108, opened 2018-09-05
Update configure with --disable-stack-protector

GXTX:
As noted in #106 this will produce working builds with GCC versions over 7.3.0-2. Upstream has already done this work-a-round.


mborgerson:
Thanks! I'm disappointed that we have to disable this hardening measure for now but because multiple systems are affected by this, I think this is the solution which has lowest impact overall. The alternative choice is to force a specific, older version of GCC which I think would be more combersome. On the bright side, we should see a performance benefit to not having these checks inside every function.

PR #111, opened 2018-09-07
 build.sh: allow out-of-tree builds

Tilka:


.gitignore
```diff
@@ -1,3 +1,4 @@
+/build/
```
MayeulC: Isn't that somewhat arbitrary?

Tilka: Not really, but I removed it.

illwieckz:
it would probably be more convenient to just change directory before calling both `configure` and `make`, this way seems safe:

```sh
cd "$(dirname "$(realpath "${0}")")"
```

It's not hard to pass a directory to `make`, but it would be harder to handle that with the windows quirks, it's definitely easier to change directory.

illwieckz:
> @illwieckz, the idea here is to generate the makefile and other build artifacts in another folder. Using `cd` would kind of defeat the purpose, I think.

Oh I see ! I thought at first glance it was just to be able to call it from everywhere, you're right !

JayFoxRox:
Tested and confirmed to work. Thanks!

I'm very sorry how long it took us to get this merged.
All of us do in-tree builds so nobody felt responsible and it was soon forgotten.

Issue #114, opened 2018-09-12
Disable debug, enable optimizations in build.sh script


mborgerson:
Currently the `build.sh` script is set up for building a version of xqemu that can be more conveniently debugged (this script originates from my dev machine during rebase work). Since this is not the typical use case for xqemu, we should remove the debugging features and enable optimization. Perhaps we should add a `--debug`/`--release` option to the build script to make building a little easier. CI systems should build both optimized and debuggable versions to catch compilation issues in either case.

JayFoxRox:
> CI systems should build both optimized and debuggable versions to catch compilation issues in either case.

In my opinion we should not do this. That's be compiler bugs and we shouldn't have to worry about those. CI builds are for testing in nature, so they should probably be debug builds, with the exception of some release/final builds (which we should start providing soon).

I agree about supporting `--debug` / `--release` command line options though (just that we don't have to test / deploy both of them for every release).

mborgerson:
> In my opinion we should not do this. That's be compiler bugs and we shouldn't have to worry about those. CI builds are for testing in nature, so they should probably be debug builds, with the exception of some release/final builds (which we should start providing soon).

I was not really suggesting we hunt for *compiler* bugs when I suggested building both debug/release, what I had in mind was:
- Debug builds and release builds can (and often do) contain different code. Consider simply `#if DEBUG` code. With debug blocks like this, it's pretty common for release builds or debug builds fail, depending on what mode the developer was writing code in. Maybe a variable definition was left in a `DEBUG` block and used outside. Maybe variables defined for debug builds are unused in release builds. These would cause compilation failures.
- Most people will want to download and run release builds for performance reasons. They may also want to use debug builds to help us track down issues. We can easily distribute both through the CI systems.
- It doesn't really cost us anything but PR/merge build time to do both builds in the CI systems. This is a perfect application of the matrix features of CI config scripts.

illwieckz:
@mborgerson what would be the optimization options to pass to configure?

mborgerson:
> what would be the optimization options to pass to configure? - @illwieckz

Basically for release builds:
* Remove `--enable-debug` config item
* Remove `-O0 -g` from `CFLAGS`

GXTX:
Wouldn't you want something like `-O3` or `-Ofast` for release - and for debug something like `-Og`?

MayeulC:
Keep in mind that in the past (that probably improved, but is still somewhat valid) compiling with `-03`:
[...]

PR #115, opened 2018-09-13
build.sh: pass options, ref #114

illwieckz:
These changes makes user able to pass `-j` make option and `--enable/disable` configure options to `./build.sh`

For example:

```sh
./build.sh -j3 --enable-debug --disable-kvm
```

An interactive help is built in.

Since bash is assumed I used nestable subcalls: `$(command)` instead of simple backquotes

I fixed the bash shebang.

I switched indentation to tabs to allow indented heredocs

I secured variables except when expansion was required.

Note that we can use `true` as a `post_build` nop to be able to secure the variable.

The script is now able to pass `$CFLAGS` env variable inherited from user shell if exists.

Now only env variables inherited from user shell are written in upper case to avoid confusion.

Perhaps I can investigate the bash arrays to handle lists a better way if it's possible to pop (I don't know) otherwise the current code already works.

Do someone knows if `nproc` is available on both MacOS and Windows in all supported cases?

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,49 +1,118 @@
-#!/bin/bash
+#! /usr/bin/env bash
 
 set -e # exit if a command fails
 set -o pipefail # Will return the exit status of make if it fails
 
-CFLAGS_COMMON="-O0 -g -DXBOX=1" # Compilation flags for all platforms
-POST_BUILD=""
+user_configure_options=''
+user_make_options=''
+debug_preset_enabled='false'
+post_build_task=''
 
-package_windows() { # Script to prepare the windows exe
-    mkdir -p dist
-    cp i386-softmmu/qemu-system-i386.exe dist/xqemu.exe
-    cp i386-softmmu/qemu-system-i386w.exe dist/xqemuw.exe
-    python2 ./get_deps.py dist/xqemu.exe dist
-    strip dist/xqemu.exe
-    strip dist/xqemuw.exe
+print_help() {
+	tab="$(printf '\t')"
+	cat <<-EOF
+	Usage: $(basename ${0}) [-t|--help] [-j<NUMBER>] [--enable-debug] [--<OPTION>]…
+
+	${tab}-h, --help
+	${tab}${tab}print this help
+
+	${tab}-j<NUMBER>, --jobs=<NUMBER>
+	${tab}${tab}set make job number
+
+	${tab}--enable-debug
+	${tab}${tab}enable debug build
+
+	${tab}--<OPTION>
+	${tab}${tab}set configure option, see below
+
+	$(./configure --help | grep '^Usage: ' -A9999 | tail -n +2)
+	EOF
+	exit
 }
 
+print_error() {
+	printf 'ERROR: %s\n' "${1}" >&2
+	exit 1
+}
+
+package_windows() { # Script to prepare the windows exe
+	mkdir -p dist
+	cp i386-softmmu/qemu-system-i386.exe dist/xqemu.exe
+	cp i386-softmmu/qemu-system-i386w.exe dist/xqemuw.exe
+	python2 ./get_deps.py dist/xqemu.exe dist
+	strip dist/xqemu.exe
+	strip dist/xqemuw.exe
+}
+
+if echo "${@}" | tr ' ' '\n' | grep -q '^-h$\|--help$'
+then
+	print_help
+fi
+
+while [ ! -z "${1}" ]
+do
+	case "${1}" in
+	'-j'*|'--jobs='*)
+		user_make_options="${user_make} ${1}"
+		;;
+	'--enable-debug')
+		user_configure_options="${user_configure_options} ${1}"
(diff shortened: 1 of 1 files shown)


illwieckz:
I edited the code to test the availability of `nproc`, waiting for buildbots ending their run to check if it fixes it…

illwieckz:
broadcasting the discord comment there because I'm going to bed so I can get any answer there about that suggestion later:

> I can make it for more simpler if needed, like appending `-j*` to a `$make_opts` and `--[en|dis]able-*` to a `$configure_opts` and forget any check and rely on people knowing what they do

to reduce drastically the build script size to not defeat the purpose of having a simple layer above `configure` & all.

```

**Comment by MayeulC** [comment]
To be frank, I do not like it much, as if users want to configure their build, that's already quite advanced, and they can:
* call configure and make themselves
* edit the script themselves

In my opinion, the build script is here only to supply a very simple way to get started, and provide a list of sane defaults.
Making it much more complex does not serve these purposes (still IMO).

There are however a few things I like, such as moving the configure options above, using `env`, a separate variable for the number of jobs.

Regarding `nproc`, you can use something like this if you want better compatibility: https://github.com/ValveSoftware/Proton/blob/proton_3.7/build_proton.sh#L19

I previously aimed for simplicity.
And finally, I am not sure if there's a real policy around this here, but I would prefer the refactoring to be split across multiple small commits, so that it's easier to track what's going on ;)

**Comment by illwieckz** [comment]
I made the code simpler, it simply extend `make` command line with `-j*|--jobs=*` options and extend `configure` command line with other `--*` options.

I added the `sysctl` trick for windows but I did explicit and proper `if; else; fi` tests because even if I find Valve's hack clever, I’m really not confident with their quick&dirty tricks at all, having lost a complete `/home` because of one bug in their script like [this one](https://github.com/valvesoftware/steam-for-linux/issues/3671) (hopefully I run steam stuff in a chroot so that deleted `/home` only contained steam stuff, nothing else).

**Comment by illwieckz** [comment]
The script now adjusts cflags when `--enable-debug' is set or not

**Comment by illwieckz** [comment]
I imagine that the `uname` trick does not work when cross-compiling (is it possible to cross-compile?) in this case it would be good to be able to enforce the target system.

**Comment by MayeulC** [comment]
Well, I still think that the main issue was not addressed: **not simple enough** (a couple other comments are still valid as well).

------
I do not see the solution that was merged in Proton's repository as particularly hackish or clever, `command || fail` is pretty standard practice. ;)
`nproc` works for the CI, but if that's a problem on macOS, and too hard to make it simple, I'd suggest dropping it, or only adjusting it in the `Linux)` case. macs seldom have more than 4 CPUs, AFAIK, anyways.

Nope, uname only works for native compilation. But if you start cross-compiling, you will likely be able to call `configure` by yourself, anyway.

-----
If you want to make it more versatile at a minimal cost, I would suggest something along the lines of:
```
if [[ $0 == --release ]]; then
    CFLAGS="-O0 -g"
    shift 2
else
    CFLAGS="-O3"
    shift
fi
```
and then, pass on the remaining options to configure via `"$@"`

I am not sure about `-O3`, however. Better agree on this before, and make it a separate commit.

----------
If you want to write a fully-fledged compilation script, that's OK as well, but I don't think this belongs here. Maybe we could make a separate folder for helper scripts/tools (as the source code is cluttered enough as-is, IMO).

JayFoxRox:
I think the solutions proposed by @MayeulC sound good: keep it very simple and logical.

More importantly, I agree: it's still not simple enough.
There's good changes here, like the shebang or the overall cleanup, but a simplifcation script should not have to provide such a convoluted `--help` in my opinion (code is documentation in this case).

So I don't see it being merged in this state; it would also diverge a lot from the original proposal, if the suggested changes are made. Therefore: Closed.
Hope this is not an issue to you @illwieckz; and thanks for taking the time to create a proposal in the first place!

MayeulC:
Well, I agree with most of what you said. The code I wrote in the comment might also have a few issues, bash isn't my mother tongue, and I wrote it as an example/general direction ;)

And I do not hold proton's (or valve's) code on a pedestal, either. I just happened to see that code earlier in the day, and thought it provided an interesting example of a failure mode.

However, line count IS important to fully grasp a program. As a random example, if you have to parse  and understand 150 lines of OS detection code before hitting real logic, that's a big barrier to entry (especially as the buils.sh is itself a barrier to entry). However, I feel that arguing about this is besides the point. And one-liners are fine, as long as you don't need to think too much to understand them.

I just got another idea to extend functionality, while not impacting readability nor simplicity, please tell me what you think of it (that's NOT necessarily standard practice, but would work just fine). How about using environment variables to pass extra options? Like `DEBUG_BUILD=1 JOBS=4 ./build.sh --extra-configure-args`. The logic to handle them should be very minimal, with the bath syntax `${VARIABLE:-default}`.

I would also change your --help output into a comment (maybe together with a message that hints at editing build.sh?)

PR #116, opened 2018-09-24
Enable HAXM support

mborgerson:
This PR enables use of HAXM, finally bringing some form of CPU virtualization to Windows and macOS users. Heavily CPU dependent portions of titles should see significant improvements when utilizing the HAXM accelerator on these platforms. There are a couple of caveats however:

1. HAXM itself required a couple of patches to fix emulation of instructions which touch MMIO registers. The latest publicly available build ([v7.3.0](https://software.intel.com/en-us/articles/intel-hardware-accelerated-execution-manager-intel-haxm) at time of writing) __will not work__! The patches have been upstreamed to the HAXM repository and will be incorporated in a future release. Until then--if you want to use HAXM, you must build the driver for yourself from source.

1. HAXM does not yet have dirty page tracking. This has an impact on nv2a as we cannot tell if the CPU has modified a resource in memory without adding some sort of hashing. This means that there are even more graphical issues than usual. Some titles are only marginally impacted by this, while others appear to be completely broken. Fortunately this feature [appears to be in the works](https://github.com/intel/haxm/issues/6), or at least part of some plan--interested people should do the work and submit a PR to the HAXM project! For XQEMU-specific use case, I don't think it would be too difficult to add this to the HAXM driver, but I look forward to seeing it developed for general use in the HAXM project (and I definitely don't want to run a fork!).

See commits for details.


target/i386/hax-mem.c
```diff
@@ -310,8 +317,17 @@ static void hax_ram_block_added(RAMBlockNotifier *n, void *host, size_t size)
     }
 }
 
+#ifdef XBOX
+static void hax_ram_block_removed(RAMBlockNotifier *n, void *host, size_t size)
+{
```
mborgerson: No-op is probably wrong, but I'm not sure what to put here yet.

JayFoxRox: - Why doesn't QEMU upstream run into this issue?
- Which handler is used in upstream QEMU?
- Did you report this upstream?
- Why not just add the NULL-check instead?

I think it's a weird Xbox specific hack which could also affect other platforms. QOM typically ignores NULL callbacks too, if I remember correctly.

mborgerson: > Why doesn't QEMU upstream run into this issue?

I guess they hadn't had any RAM blocks removed in testing. We run into this when we swap out VGA RAM during NV2A basic init.

> Which handler is used in upstream QEMU?

What do you mean?

> Did you report this upstream?

No, but it should be fixed upstream. I'll probably report it in the coming days.

> Why not just add the NULL-check instead?

I've run into this problem with other QEMU code too--mostly due to interface changes. I also think it is best practice to check for `NULL` handlers before invoking them, but in this instance I think HAXM *should* support this method, so adding the handler is appropriate.

PR #120, opened 2018-10-08
Replace ffs() calls with ctz32()

### LINKED ISSUE #88: Switch from `ffs` to `ctz32`
Checkpatch recommends the use of `ctz32` instead of `ffs`.
We should replace all instances in our existing code.

Additionally (and the reason why this issue exists), we ship our own `ffs` implementation for `__WINNT__` near the top of nv2a.c .
As we remove use of `ffs`, we can also remove this function.

thrimbor:
This replaces all ffs() calls in NV2A and MCPX with calls to ctz32() and adjusts the surrounding code. It also removes the ffs() implementation in NV2A provided for Windows.
The changes build and run fine under Arch Linux and Windows 7 (as far as I can tell).

Resolves #88.

### DIFF
--- a/hw/xbox/mcpx_apu.c
+++ b/hw/xbox/mcpx_apu.c
@@ -121,12 +121,12 @@ static const struct {
 
 #define MCPX_HW_MAX_VOICES 256
 
-#define GET_MASK(v, mask) (((v) & (mask)) >> (ffs(mask) - 1))
+#define GET_MASK(v, mask) (((v) & (mask)) >> ctz32(mask))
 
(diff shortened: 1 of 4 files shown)


PatrickvL:
In most calls, ffs() could be optimized away, because it was "called" with a constant. Does that apply to ctz32() too?

PatrickvL:
I'll take your word for it, as I've got no inside knowledge about how various compilers handle such cases - at least in c++ there's the keyword "constexpr" which tries to evaluate as much code as possible at compile time (avoiding run time costs)

mborgerson:
Thanks!

PR #121, opened 2018-10-08
build.sh: pass options, fix #114 [smaller]

### LINKED ISSUE #114: Disable debug, enable optimizations in build.sh script

illwieckz:
Well, I was waiting for #111 to be merged to work on #115 again, but they were closed in a row…

So, this is a newer `build.sh` supporting job count and configure options, like in #115 but stripped down.

- [fix] rely on `/usr/bin/env` to call `bash`
- [fix] always limit the number of spawned processes
- [fix] do not overwrite CFLAGS standard environment variable
- [enhancement] allows to pass `-j[something]` as argument
- [enhancement] enable `--release` to disable debug builds
- [enhancement] pass all other options to `configure`
- [enhancement] detect `--enable-debug` option to set cflags for debug builds
- [enhancement] make the configure path more reliable when building out of tree
- [enhancement] print error to `/dev/stderr`
- [bikeshedding] unify indention (it was mixed space/tab, now space like xqemu)

> I still think we don't need to do nproc at all

In fact I had the need to set the job count since months and I was gonna make a PR in any case. When I've seen #114 I thought it would be time to make a better `build.sh` that fixes both #114 needs and mines.

Note that `build.sh` makes no divination about cpu count anymore, it's up to the user to use the command from his system.

This is an example of possible usage of that new `build.sh`:

```sh
CFLAGS='-march=znver1' ./build.sh -j$(nproc) --release --disable-kvm
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,10 +1,7 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e # exit if a command fails
 set -o pipefail # Will return the exit status of make if it fails
-
-CFLAGS_COMMON="-O0 -g -DXBOX=1" # Compilation flags for all platforms
-POST_BUILD=""
 
 package_windows() { # Script to prepare the windows exe
     mkdir -p dist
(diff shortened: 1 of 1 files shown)


illwieckz:
I squashed them, I really don't need `nproc` to be in `build.sh` at all, I just want to be able to do:

> ./build.sh -j$(nproc)

illwieckz:
Note that the additions are just two `if` tests now:

```
if [ "${1:0:2}" = '-j' ]
then
	job_count="${1:2}"
	shift
fi

if echo " ${@} " | grep -q '[[:space:]]--enable-debug[[:space:]]'
then
	build_cflags='-O0 -g'
fi
```

All the other changes are minor changes to make things better, not added features.

build.sh
```diff
@@ -1,49 +1,62 @@
...
+		sys_opts='--python=python2 --disable-cocoa --disable-opengl'
+		postbuild='package_windows' # set the above function to be called after build
+		;;
+	*)
+		echo "could not detect OS $(uname -s), aborting" >&2
+		exit -1
+		;;
 esac
 
+configure="$(dirname "$(readlink -f "${0}")")/configure"
```
JayFoxRox: Probably want to move this to move the generated `configure` directory path and explain why this is so complicated in a comment

build.sh
```diff
@@ -1,49 +1,62 @@
...
+build_cflags='-O3'
+job_count='4'
+
+if [ "${1:0:2}" = '-j' ]
+then
+	job_count="${1:2}"
+	shift
+fi
+
+if echo " ${@} " | grep -q '[[:space:]]--enable-debug[[:space:]]'
```
JayFoxRox: Either we go back to the previous style #115 of looping, or this should be in a function which hints what this does:

```
function find_in_arguments() {
  echo " ${@} " | grep -q '[[:space:]]$1[[:space:]]'
}
...
if $(find_in_arguments("--enable-debug")); then
```

*(Don't trust my bash code - it's untested)*

build.sh
```diff
@@ -1,49 +1,62 @@
...
-    cp i386-softmmu/qemu-system-i386w.exe dist/xqemuw.exe
-    python2 ./get_deps.py dist/xqemu.exe dist
-    strip dist/xqemu.exe
-    strip dist/xqemuw.exe
+	mkdir -p dist
+	cp i386-softmmu/qemu-system-i386.exe dist/xqemu.exe
+	cp i386-softmmu/qemu-system-i386w.exe dist/xqemuw.exe
+	python2 ./get_deps.py dist/xqemu.exe dist
+	strip dist/xqemu.exe
+	strip dist/xqemuw.exe
```
JayFoxRox: QEMU uses 4 spaces instead of tab. I don't think we should do this any differently in a support script, unless the language requires or strongly recommends it.

build.sh
```diff
@@ -1,49 +1,62 @@
-#!/bin/bash
+#! /usr/bin/env bash
```
JayFoxRox: nit: `#!/usr..` without space?

The interpreter starts right after the `#!`, I'm not sure if this makes a practical difference

illwieckz: oh yes the space is a bad habit from some very old age that belong to the earlier history of unix (so it precedes me), it can be dropped even if supported, it makes no sense to add it, it's just my fingers that are adding them in an automated manner :p

build.sh
illwieckz: it's ok to use space, there was just a mix of both (see the os switch and configure indentation)

build.sh
```diff
@@ -1,49 +1,62 @@
...
 
+postbuild=''
+build_cflags='-O3'
+job_count='4'
+
+if [ "${1:0:2}" = '-j' ]
+then
+	job_count="${1:2}"
+	shift
+fi
```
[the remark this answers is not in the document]
illwieckz: > `configure-xqemu.sh --release && make -j 4`

I like it. A lot.

There is no need for `.sh` though, I said nothing for `build.sh` because there was already a `build` dir and it was an easy way to avoid file name collision, but [commandname extensions considered harmful](https://www.talisman.org/~erlkonig/documents/commandname-extensions-considered-harmful/).

build.sh
```diff
@@ -1,49 +1,62 @@
...
+
+if echo " ${@} " | grep -q '[[:space:]]--enable-debug[[:space:]]'
+then
+	build_cflags='-O0 -g'
+fi
```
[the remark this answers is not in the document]
illwieckz: I can do that

mborgerson:
Thanks for this! I'm glad we finally have a `--release` option. Now we need to update Appveyor to build in release mode.

PR #122, opened 2018-10-16
Fix EEPROM writeback; implement writing EEPROM to file

0DaveX:
Fixes a bug while writing to the eeprom (writing 0xABCD) would write (0xAB00) 
https://github.com/0DaveX/xqemu/blob/9344483a0ce7079c874f7e0db9ea38e1bc576ba9/hw/xbox/amd_smbus.c#L115
https://github.com/0DaveX/xqemu/blob/84e63a64a42197f01785c6df93a4f48e6448f0b2/hw/xbox/amd_smbus.c#L115

also followup to @dracc his EEPROM PR 
> Ideally we'd store the EEPROM contents in a file which is read and written at runtime.

so it fixes #107

### DIFF
--- a/hw/i2c/smbus_eeprom.c
+++ b/hw/i2c/smbus_eeprom.c
@@ -26,6 +26,7 @@
 #include "hw/hw.h"
 #include "hw/i2c/i2c.h"
 #include "hw/i2c/smbus.h"
+#include "hw/xbox/xbox.h"
 
 //#define DEBUG
 
@@ -84,6 +85,8 @@ static void eeprom_write_data(SMBusDevice *dev, uint8_t cmd, uint8_t *buf, int l
     len -= n;
     if (len)
         memcpy(eeprom->data, buf + n, len);
+
+    save_eeprom(eeprom->data);
 }
 
 static uint8_t eeprom_read_data(SMBusDevice *dev, uint8_t cmd, int n)
--- a/hw/xbox/xbox.c
+++ b/hw/xbox/xbox.c
@@ -52,7 +52,9 @@
 
 #define MAX_IDE_BUS 2
 
-// XBOX_TODO: Should be passed in through configuration
+int eeprom_fd = -1;
+const int eeprom_size = 256;
+
 /* bunnie's eeprom */
 const uint8_t default_eeprom[] = {
     0xe3, 0x1c, 0x5c, 0x23, 0x6a, 0x58, 0x68, 0x37,
@@ -227,10 +229,8 @@ static void xbox_memory_init(PCMachineState *pcms,
 uint8_t *load_eeprom(void)
 {
     char *filename;
-    int fd;
     int rc;
     int eeprom_file_size;
-    const int eeprom_size = 256;
 
     uint8_t *eeprom_data = g_malloc(eeprom_size);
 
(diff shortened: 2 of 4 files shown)


hw/xbox/xbox.c
```diff
@@ -250,31 +250,40 @@ uint8_t *load_eeprom(void)
...
         g_free(filename);
     } else {
         memcpy(eeprom_data, default_eeprom, eeprom_size);
     }
     return eeprom_data;
 }
 
+void save_eeprom(uint8_t *eepromBuffer)
+{
+   int status = -1;
```
mborgerson: Written but never read?

hw/xbox/xbox.c
```diff
@@ -250,31 +250,40 @@ uint8_t *load_eeprom(void)
...
+
+    if(eeprom_fileDescriptor > 0)  {
+        lseek(eeprom_fileDescriptor,0,SEEK_SET);
+        status = write(eeprom_fileDescriptor, eepromBuffer, 256);
```
mborgerson: What if this fails?

hw/xbox/xbox.c
```diff
@@ -52,7 +52,8 @@
 
 #define MAX_IDE_BUS 2
 
-// XBOX_TODO: Should be passed in through configuration
+int eeprom_fileDescriptor;
```
mborgerson: Naming convention is not very consistent. By the way *should* we keep the file open always? If yes do we need to write the whole buffer every time?

JayFoxRox:
Please move the EEPROM writeback fix into a separate commit (or even PR). If we undo anything from this, we'd still want to keep this bugfix.
I'm surprised we haven't ran into issues with this word-write bug already.

hw/i2c/smbus_eeprom.c
```diff
@@ -84,6 +85,8 @@ static void eeprom_write_data(SMBusDevice *dev, uint8_t cmd, uint8_t *buf, int l
     len -= n;
     if (len)
         memcpy(eeprom->data, buf + n, len);
+
+   save_eeprom(eeprom->data);
```
mborgerson: Indentation should be 4 spaces instead of 3

hw/xbox/xbox.c
```diff
@@ -250,31 +250,40 @@ uint8_t *load_eeprom(void)
...
     }
     return eeprom_data;
 }
 
+void save_eeprom(uint8_t *eepromBuffer)
+{
+   int status = -1;
+
+    if(eeprom_fileDescriptor > 0)  {
+        lseek(eeprom_fileDescriptor,0,SEEK_SET);
```
mborgerson: nit: spaces after `,` please

hw/xbox/xbox.c
0DaveX: leftover from debugging - so I'd remove it, right?

hw/xbox/xbox.c
0DaveX: We most likely already printed an error (while reading the eeprom file)
So if we can't write to the file we don't try to :)

hw/xbox/xbox.c
0DaveX: Oky sound good.

hw/xbox/xbox.c
JayFoxRox: Here you use `256`, but above you have `eeprom_size = 256`.

0DaveX: thanks - took care

hw/xbox/xbox.c
```diff
@@ -52,7 +52,9 @@
 
 #define MAX_IDE_BUS 2
 
-// XBOX_TODO: Should be passed in through configuration
+int eeprom_fd = -1;
```
JayFoxRox: This will have to be in `XboxMachineState`, in case someone creates more than one `xbox` machine.

hw/xbox/xbox.c
```diff
@@ -250,31 +250,38 @@ uint8_t *load_eeprom(void)
...
 
-        close(fd);
         g_free(filename);
     } else {
         memcpy(eeprom_data, default_eeprom, eeprom_size);
     }
     return eeprom_data;
 }
 
+void save_eeprom(uint8_t *eepromBuffer)
```
JayFoxRox: Parameter name still bad. Not sure about a better one, but if you want to keep it, this should still be `eeprom_buffer`

hw/xbox/xbox.h
```diff
@@ -22,10 +22,12 @@
 #define HW_XBOX_H
 
 #include "hw/boards.h"
+#include "include/hw/i386/pc.h"
 
 #define MAX_IDE_BUS 2
 
 uint8_t *load_eeprom(void);
+void save_eeprom(uint8_t *eepromBuffer);
```
JayFoxRox: Same as other instance

hw/i2c/smbus_eeprom.c
```diff
@@ -84,6 +85,8 @@ static void eeprom_write_data(SMBusDevice *dev, uint8_t cmd, uint8_t *buf, int l
     len -= n;
     if (len)
         memcpy(eeprom->data, buf + n, len);
+
+    save_eeprom(eeprom->data);
```
JayFoxRox: How would QEMU know *which* eeprom to use?
You'll need some QOM magic here to figure out who this eeprom belongs to.

I can't really help with it either, because of my hatred and resulting inexperience with QOM.

JayFoxRox:
I think we should close this because of the design issues and lack of activity.

Currently, the EEPROM object has to call a global function which is part of the Xbox (so the eeprom shouldn't even have to know about it), to save its contents.
A proper implementation would tell the EEPROM where it should store its contents (recommended), *or* the Xbox should fetch contents of the EEPROM and handle storing itself.

`eeprom: Fix a word-write bug ` should be resubmitted immediately on its own.
The other commit needs more work.

Thoughts?

PR #125, opened 2018-10-31
Pixel shader cleanup

JayFoxRox:
*I've originally intended to do more changes as part of a larger PR, but am splitting it into managable chunks now. This code has not been tested but it should work. Please test and review.*

This PR makes our c0 / c1 names more logical, and removes some dead-code. We had some variables which were only written, but never read.
I've also re-ordered texture/const-declaration and alpha-ref-logic code, because our order didn't make sense.

There should be no changes in observed behaviour.

### DIFF
--- a/hw/xbox/nv2a/nv2a_psh.c
+++ b/hw/xbox/nv2a/nv2a_psh.c
@@ -167,7 +167,6 @@ enum PS_FINALCOMBINERSETTING
 
 struct InputInfo {
     int reg, mod, chan;
-    bool invert;
 };
 
 struct InputVarInfo {
@@ -176,9 +175,6 @@ struct InputVarInfo {
 
 struct FCInputInfo {
     struct InputInfo a, b, c, d, e, f, g;
-    int c0, c1;
-    //uint32_t c0_value, c1_value;
-    bool c0_used, c1_used;
     bool v1r0_sum, clamp_sum, inv_v1, inv_r0, enabled;
 };
 
@@ -191,8 +187,6 @@ struct PSStageInfo {
     struct InputVarInfo rgb_input, alpha_input;
     struct OutputInfo rgb_output, alpha_output;
     int c0, c1;
-    //uint32_t c0_value, c1_value;
-    bool c0_used, c1_used;
 };
 
 struct PixelShader {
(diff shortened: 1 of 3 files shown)


hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -245,36 +239,23 @@ static QString* get_var(struct PixelShader *ps, int reg, bool is_dest)
         }
         break;
     case PS_REGISTER_C0:
-        /* TODO: should the final stage really always be unique? */
         if (ps->flags & PS_COMBINERCOUNT_UNIQUE_C0 || ps->cur_stage == 8) {
-            QString *reg = qstring_from_fmt("c_%d_%d", ps->cur_stage, 0);
+            QString *reg = qstring_from_fmt("c0_%d", ps->cur_stage);
```
mborgerson: Order got swapped here, is that right? What would have been `c_X_0` is now `c0_X`

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -506,12 +487,11 @@ static void add_final_stage_code(struct PixelShader *ps, struct FCInputInfo fina
     QString *d = get_input_var(ps, final.d, false);
     QString *g = get_input_var(ps, final.g, false);
 
-    add_var_ref(ps, "r0");
-    qstring_append_fmt(ps->code, "r0.rgb = %s + mix(vec3(%s), vec3(%s), vec3(%s));\n",
+    qstring_append_fmt(ps->code, "fragColor.rgb = %s + mix(vec3(%s), vec3(%s), vec3(%s));\n",
```
mborgerson: Why change this here? Don't we set `fragColor = r0` later on?

hw/xbox/nv2a/nv2a_psh.c
JayFoxRox: Yes, that's the intention with this commit.

The register names (used by D3D) are C0 and C1. GL calls them GL_CONSANT_COLOR0_NV and GL_CONSANT_COLOR1_NV.

I think it's weird to call them "C[i]0" and "C[i]1" (which is basically what we do in master).

hw/xbox/nv2a/nv2a_psh.c
JayFoxRox: Using `r0` etc. is a leftover from Cxbx code where they used D3D ASM shaders. So they had to use these sort of temporaries.

The real output variable is called `fcoc` ([See "3.8.12.3  Final Combiner Stage Operation"](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_register_combiners.txt)).

`fragColor` is much closer and shows the intention.

*I might rename `fragColor` to `fcoc` in the future or add a temporary again (for other reasons), but it's not really beneficial here. Naming it `r0` is just confusing.*

mborgerson:
I think we still have a lot of room for improvement in our shader code, and this is a step in the right direction. Thanks!

PR #126, opened 2018-10-31
Force GL surface transfer on surface address or pitch changes

JayFoxRox:
Our surface logic is really flawed.

In "Tony Hawk's American Wasteland" the game has data in buffer A [640x480], then it switches to buffer B [320x240], then buffer C [640x480], and then back to buffer A [640x480].

Our current code downloads (GL to RAM) the surface when going from A -> B (so A is up-to-date), it's not uploaded (RAM to GL) again when going from C -> A. So the GL buffer for A remains what was in the GL buffer for C. If we switch to another buffer now, A might be downloaded to RAM again, and will have completly wrong contents (GL buffer contents of C!).

That is because our current code only appears to check if the CPU has written a surface OR if the surface shape has changed before re-uploading. It doesn't even care if the address changed. The same issue exists for surface pitch.

This PR changes the debug code so this can be observed in apitrace, and then it addresses the problem, by manually marking the surface as dirty when the address or pitch are touched.
This fixes the described issue in "Tony Hawk's American Wasteland" (and probably other games using bloom or other similar screen-space effects).

I'm not sure if my solution is good, because I haven't worked with our rather complicated surface logic in a while.

This change can be very bad for performance, as a lot more transfers are possibly forced. It should eventually be implemented closer to what we do for `surface_shape` already (only force an update, if the actual value changed). However, someone else can fix it later.

*There might be other cases where we run into similar issues, there'll be a separate issue about this*

### DIFF
--- a/hw/xbox/nv2a/nv2a_pgraph.c
+++ b/hw/xbox/nv2a/nv2a_pgraph.c
@@ -738,16 +738,21 @@ static void pgraph_method(NV2AState *d,
             GET_MASK(parameter, NV097_SET_SURFACE_PITCH_COLOR);
         pg->surface_zeta.pitch =
             GET_MASK(parameter, NV097_SET_SURFACE_PITCH_ZETA);
+
+        pg->surface_color.buffer_dirty = true;
+        pg->surface_zeta.buffer_dirty = true;
         break;
     case NV097_SET_SURFACE_COLOR_OFFSET:
         pgraph_update_surface(d, false, true, true);
 
         pg->surface_color.offset = parameter;
+        pg->surface_color.buffer_dirty = true;
         break;
     case NV097_SET_SURFACE_ZETA_OFFSET:
         pgraph_update_surface(d, false, true, true);
 
         pg->surface_zeta.offset = parameter;
+        pg->surface_zeta.buffer_dirty = true;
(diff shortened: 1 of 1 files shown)


mborgerson:
It's late in the evening for me right now, but this PR makes sense: whenever the surface address is changed, we should consider the surface dirty.

I ran a simple test with Halo:CE MP and noticed an interesting new artifact--a rectangle is now drawn on the ground below the warthog. But..I think this is actually better than master (or approaching correct, anyway), as this *appears* to be the warthog shadow which has thus far been missing!

Before:
Imgur

After:
Imgur

Expected (crappy screengrab taken from [this video](https://www.youtube.com/watch?v=XVSdaNRKcQQ)):
Imgur

[YouTube demonstration video here](https://www.youtube.com/watch?v=9fx94aEr3wc).

mborgerson:
Per my comment above, and based on my testing in Halo:CE, I think this change makes sense and is a step in the right direction. I did not observe a significant performance penalty, but as mentioned in the PR description, it's likely that we may find one--I believe that's an acceptable cost for now, which would be mitigated once we implement lazy surfaces anyhow. For these reasons I think this is ready to merge, though I would like to see additional games tested.

Issue #149, opened 2018-12-26
DSP SR is truncated to 16 bits (lack of EMR)


JayFoxRox:
I tried to compare XQEMU to my hardware Xbox using https://github.com/JayFoxRox/xbox-tools/pull/70.
However, most SR output is broken, because our DSP code was made for DSP56k where SR was 16 bit.

A lot of the existing code does something like:

```c
    dsp->registers[DSP_REG_SR] &= BITMASK(16)-((1<<DSP_SR_V)|(1<<DSP_SR_C));
    dsp->registers[DSP_REG_SR] |= newsr;
```

DSP56300 added the EMR, [see section "5.4.1.2 Status Register (SR)"](https://www.nxp.com/docs/en/reference-manual/DSP56300FM.pdf), so SR should be 24 bits.

We have to find these, and modify it for `BITMASK(24)`.
We could also use `registers_mask[DSP_REG_SR]` and move this SR modification into a function, which will make it easier to avoid such issues altogether.

Issue #148, opened 2018-12-26
DSP accumulator extension registers are not sign-extended when used as source


JayFoxRox:
I ran into an issue when doing `move A2, X:$0`. I had expected to see 0xFFFF80 (as on hardware), but I got 0x000080 (in XQEMU).

---

[See page 6-6 in the DSP5600 manual](https://www.nxp.com/docs/en/reference-manual/DSP56000UM.pdf):

> When an accumulator extension register acts as a source operand, it occupies the low-
order portion (bits 0–7) of the word and the high-order portion (bits 8–23) is sign extended
(see Figure 6-4).

XQEMU seems to ignore this rule.

We can probably look for a pattern like:

```c
        if ((numreg1==DSP_REG_A) || (numreg1==DSP_REG_B))
            emu_pm_read_accu24(dsp, numreg1, &save_reg1);
        else
            save_reg1 = dsp->registers[numreg1];
```

(also exists with other variable names)

Then we have to replace it with something like this (should be in a function):

```c
        if ((numreg==DSP_REG_A) || (numreg==DSP_REG_B))
            emu_pm_read_accu24(dsp, numreg, &value);
        else {
            value = dsp->registers[numreg];

            /* When an accumulator extension register acts as a source operand,
             * it occupies the low-order portion (bits 0–7) of the word and the
             * high-order portion (bits 8–23) is sign extended */
            if ((numreg==DSP_REG_A2) || (numreg==DSP_REG_B2))
                if (value & 0x80)
                    value |= 0xffff00;
        } 
```

I don't think we can easily find all of these cases, but we can iteratively improve our code.

PR #152, opened 2018-12-26
 dsp: Support 'inc d'  and 'dec d'

JayFoxRox:
Saw 'inc d' in some games GP FX or EP AC3 code once, so we will probably need it for audio emulation; also using it in DSP homebrew now. Adding 'dec d' was more of an afterthought for completeness.

I tested my implementation using https://github.com/JayFoxRox/xbox-tools/pull/70 (requires https://github.com/XboxDev/a56/pull/9).
I ran a couple of test cases to see how flags behave and if A / B register selection works. Everything seems to work fine, except readback issues which are likely caused by #148 and #149.

I did not implement the respective disassemblers yet (and don't intend to do so).

I didn't check for correctness of SR.S and SR.L bits, simply because they don't seem to be updated explicitly by most of our DSP code. These will have to be fixed in the future, if they are broken.


mborgerson:
I've compared these additions against the (presumably correct) [DSP56300 manual](https://www.nxp.com/docs/en/reference-manual/DSP56300FM.pdf), and against the existing emulation instructions. I would like to see the disassembler routines added at some point since they are pretty trivial, but we can leave it for another PR. Thanks!

PR #159, opened 2018-12-26
apu: Access system memory directly

### LINKED ISSUE #156: APU memory access is slow
The APU uses `stb_phys(&address_space_memory, paddr, ptr[i]);` to write bytes to scratch space, and `ldub_phys(&address_space_memory, paddr);` to read from scratch space.

Instead, we should probably be using something like NV2A, where xbox.c passes the `ram_memory`, so we can access the memory directly.

The same will be relevant for the FIFOs and voices.

We also use `ldl_le_phys` to get scratch space PRDs as part of the memory access loop - that shouldn't be necessary (unless a memory access overwrites the PRD, which we should probably check for).

JayFoxRox:
This improves the APU memory access performance. This will be more important when we implement better DMA (possibly more data to transfer), VP (audio buffers must be loaded) and EP (FIFO must be read and written back).

Closes #156.

*I did not check for self-modifying PRDs (I don't think this is likely to happen, until we write our own crazy DSP homebrew).*

---

This has gotten very little testing. Please review my new logic for `scratch_rw`.

### DIFF
--- a/hw/xbox/mcpx_apu.c
+++ b/hw/xbox/mcpx_apu.c
@@ -2,6 +2,7 @@
  * QEMU MCPX Audio Processing Unit implementation
  *
  * Copyright (c) 2012 espes
+ * Copyright (c) 2018 Jannik Vogel
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -23,6 +24,8 @@
 #include "hw/pci/pci.h"
 #include "cpu.h"
 #include "hw/xbox/dsp/dsp.h"
+
+#include "hw/xbox/mcpx_apu.h"
 
 #define NV_PAPU_ISTS                                     0x00001000
 #   define NV_PAPU_ISTS_GINTSTS                               (1 << 0)
@@ -139,6 +142,9 @@ static const struct {
 
 typedef struct MCPXAPUState {
     PCIDevice dev;
+
+    MemoryRegion *ram;
+    uint8_t *ram_ptr;
 
     MemoryRegion mmio;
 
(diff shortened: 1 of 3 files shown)


hw/xbox/mcpx_apu.c
```diff
@@ -403,25 +408,42 @@ static const MemoryRegionOps vp_ops = {
...
+                       hwaddr sge_base, unsigned int max_sge,
                        uint8_t *ptr, uint32_t addr, size_t len, bool dir)
 {
-    int i;
-    for (i = 0; i < len; i++) {
-        unsigned int entry = (addr + i) / TARGET_PAGE_SIZE;
-        assert(entry < max_sge);
+    unsigned int page_entry = addr / TARGET_PAGE_SIZE;
+    unsigned int offset_in_page = addr % TARGET_PAGE_SIZE;
+    unsigned int bytes_to_copy = TARGET_PAGE_SIZE - offset_in_page;
```
mborgerson: We could replace these division ops with `page_entry = addr >> TARGET_PAGE_BITS` and `offset_in_page = addr & (TARGET_PAGE_BITS-1)`, though the compiler will probably optimize to this anyhow.

JayFoxRox: Yeah, I assumed the compiler will optimize this, and this version is probably more readable.

*I assume you mean `offset_in_page = addr & (TARGET_PAGE_SIZE-1)`, which also depends on alignment of the page-size.*

hw/xbox/mcpx_apu.c
```diff
@@ -24,6 +24,8 @@
 #include "cpu.h"
 #include "hw/xbox/dsp/dsp.h"
```
mborgerson: Nit: newline

JayFoxRox: Same style as https://github.com/xqemu/xqemu/blob/master/hw/xbox/nv2a/nv2a.c#L41

mborgerson: I prefer the includes to not be double spaced like this so I'd remove the newline in nv2a.c as well. But I'm not too concerned about it, just a nit.

hw/xbox/mcpx_apu.c
```diff
@@ -403,25 +408,42 @@ static const MemoryRegionOps vp_ops = {
...
+
         uint32_t prd_address = ldl_le_phys(&address_space_memory,
-                                           sge_base + entry * 8 + 0);
+                                           sge_base + page_entry * 8 + 0);
         /* uint32_t prd_control = ldl_le_phys(&address_space_memory,
-                                            sge_base + entry * 8 + 4); */
+                                            sge_base + page_entry * 8 + 4); */
 
-        hwaddr paddr = prd_address + (addr + i) % TARGET_PAGE_SIZE;
+        hwaddr paddr = prd_address + offset_in_page;
```
mborgerson: Hopefully paddr is in [0, 128M)... Might be good to have a sanity check here before we access RAM?

JayFoxRox: Yeah, I'll add this. Good catch.

hw/xbox/mcpx_apu.c
```diff
@@ -403,25 +409,44 @@ static const MemoryRegionOps vp_ops = {
     .write = vp_write,
 };
 
-static void scratch_rw(hwaddr sge_base, unsigned int max_sge,
+static void scratch_rw(MCPXAPUState *d,
```
JayFoxRox: We should probably rename this to `scatter_gather_rw`.

It hasn't been a "scratch" accessor for some time, and will *very* soon be used for FIFO and other VP access.

PR #166, opened 2018-12-27
dsp: Add dummy EMR to SR

### LINKED ISSUE #149: DSP SR is truncated to 16 bits (lack of EMR)

JayFoxRox:
Closes #149 

The motivation is purely so I can compare register dumps from a real Xbox to XQEMU. I don't intend to implement EMR at all.

---

I've attempted to "search & replace" all instances where the DSP modifies SR (which also limits to 16 bit).

**I've replaced the old code which clears SR flags, with a new function, called `emu_clear_sr_flags`.**
It's assumed that this function will always be inlined most of the time. I did not verify this yet.
I also explicitly reduce SR to its size, because the original code also always reduced size to 16 bit.

**This allowed to bump the `registers_mask` size to 24 bits for SR.**
This means we have space for the EMR now

**I then had to change the way SR is printed in disasm through `disasm_reg_compare`.**
The old printf was meant for 4 hex digits, but we need 6 now. I was able to just move the `DSP_REG_SR` case.

**I've replaced a weird mask in `dsp_write_reg` with asserts, so we can catch invalid SR access.**
I don't fully understand the old mask:

* Scaling: No idea why this was masked; either it's not supported, or this is something in the DSP I don't understand.
* Reserved: This looks fine.
* Double-Precision: I assume that the code doesn't support it, but it should probably error, instead of silently ignoring it.

The new asserts handle each of those cases, but also check for EMR = 0xC0 (which is the default at boot, and the only mode we support).

**I also added SR initialization to the CPU reset.**
This is [as documented in the DSP FM section 5.4.1.2 "Status Register (SR)"](https://www.nxp.com/docs/en/reference-manual/DSP56300FM.pdf). I'm not sure why the existing code didn't do it.

**I decided to ignore checkpatch: the existing code is not clean either.**

---

This is a low-priority PR. We should triple-check so my "search & replace" doesn't break anything.
I did not test this PR, aside from running a very tiny DSP program.


JayFoxRox:
I should probably split this into different commits, so the search & replace is largely isolated.
EMR could be added in a separate commit.

JayFoxRox:
This PR is missing some change in `dsp_get_register_address` probably.

JayFoxRox:
Would be cool if someone could review the DSP code to see if I missed anything.

Again: this PR is not critical at all, it just makes debugging slightly easier.

It can just continue to marinade here until someone picks it up or we confirm it. I'll probably run into SR / EMR issues again in the future, and will then take care of this myself if nobody beats me to it.

hw/xbox/dsp/dsp.c
```diff
@@ -314,184 +314,3 @@ void dsp_print_registers(DSPState* dsp)
     printf("SP: %02x    SSH: %04x  SSL: %04x\n", 
         dsp->core.registers[DSP_REG_SP], dsp->core.registers[DSP_REG_SSH], dsp->core.registers[DSP_REG_SSL]);
 }
-
```
mborgerson: Should we keep the other debugger functions, which are also unused (e.g. `dsp_print_registers`, `dsp_info`, etc.)?

JayFoxRox: No; see my reasoning from PR description:

> These functions weren't suitable to be called from gdb (pointer arguments), and also weren't used by our code.

`dsp_print_registers` and `dsp_info` can be called from gdb and provide valuable information.

By contrast, calling `dsp_get_register_address` from gdb is a lot more complicated and pointless. We can just `print dsp->core.registers[DSP_REG_A0]` instead, or just use `call dsp_print_registers(&dsp)` (which provides the same information).

`dsp_get_register_address` would only be needed if we want to retrieve bits of a single register by name from within our code (like in a GUI, a custom repl, etc.).

mborgerson:
According to the manual, the scaling bit (sr[7]) is...
> Set when a result moves from accumulator A or B to the XDB or YDB buses (during an accumulator-to-memory or accumulator-to-register move) and remains set until explicitly cleared by an instruction or by a hardware rest; that is, the Scaling (S) bit is a sticky bit. This bit is computed, according to the logical equations shown here when an instruction or a parallel move reads the contents of accumulator A or Bto the XDB or YDB bus.
> The S bit detects data growth, which is required in Block Floating-Point FFT operation. The S bit is set if the absolute value in the accumulator, before scaling, is greater than or equal to 0.25 and smaller than 0.75. Typically, the bit is tested after each pass of a radix 2 decimation-in-time FFT and, if it is set, the appropriate scaling mode should be activated in the next pass. 

It seems to me that we do support scaling (see `emu_pm_read_accu24`), but (based on my limited inspection) for some reason don't appear to ever this bit. Interestingly, it's not even defined in dsp_cpu.h

-- 

Apart from this, I've reviewed the code and the patches look simple and straightforward. I have not yet tested this, but based on inspection of the patches I expect it to work exactly as before (save for the printing stuff).

We should discuss the future of the debug handler functions, of which you've removed two but we still retain several unused functions (not a blocking issue). Though unused, these may be useful when debugging (as you mentioned, for example when calling from gdb).

hw/xbox/dsp/dsp.c
mborgerson: I'm confused, by this response: did you mean to say *yes* we should keep them, because they can be called from gdb and are useful?

JayFoxRox: - The ones I'm removing are not useful / are hard to call from gdb (so they should be removed).
- The ones I did not remove are useful to be called from gdb (so we should keep them).

JayFoxRox:
> It seems to me that we do support scaling (see `emu_pm_read_accu24`), but (based on my limited inspection) for some reason don't appear to ever [sic; - set?] this bit. Interestingly, it's not even defined in dsp_cpu.h

Yes, the [code in `emu_pm_read_accu24`](https://github.com/xqemu/xqemu/blob/f94bcbacfa224664060fd6337681a3f6629cdca2/hw/xbox/dsp/dsp_emu.inl#L5143) (and other code) respects the scaling mode (`sr[10-11]`) but never sets the scaling bit (`sr[7]`).

*I also believe the linked manual has an error in `sr[5]`:  it also mentions an "S Bit Equation" but continues to explain E without mentioning S (This [DSP56000 manual](https://www.nxp.com/docs/en/reference-manual/DSP56000UM.pdf) calls this "Integer Portion" instead - no mention of S).*

I assume it's an unused feature in Hatari so it was never implemented. I'm not sure if Xbox depends on it. We should probably create an issue so we don't forget about it.

I assume I made that comment in the PR description to mention all SR bits that were originally forced to 0. However, I don't really understand (or remember) what my comment was supposed to express. Maybe I mistook it for scaling mode at the time?

> We should discuss the future of the debug handler functions, of which you've removed two but we still retain several unused functions (not a blocking issue). Though unused, these may be useful when debugging (as you mentioned, for example when calling from gdb).

I had responded here: https://github.com/xqemu/xqemu/pull/166#discussion_r322480084

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -1079,7 +1080,14 @@ static void dsp_write_reg(dsp_core_t* dsp, uint32_t numreg, uint32_t value)
             dsp->registers[DSP_REG_OMR] = value & 0xc7;
             break;
         case DSP_REG_SR:
-            dsp->registers[DSP_REG_SR] = value & 0xaf7f;
+
+            /* Check for unhandled bits */
+            assert((value & (1 << 7)) == 0); /* Scaling */
+            assert((value & (1 << 12)) == 0); /* Reserved */
+            assert((value & (1 << 14)) == 0); /* Double-Precision */
+            assert((value & 0xFF0000) == 0xC00000); /* EMR */
```
JayFoxRox: nit: This should probably use constants like `DSP_SR_S` and `DSP_SR_DM` and `DSP_SR_EMR` (or individual bits from `DSP_SR_EMR`)

That way, once we implement them, we won't be able to forget to update this.

PR #167, opened 2018-12-28
Support FIFO output in DMA

JayFoxRox:
This adds support for FIFO output in APU DSP DMA transfers. This is used during the boot-animation to send the audio output to AC97.
During normal operation (post boot-animation), this DMA to AC97 is handled by the EP (which we currently don't emulate). So in practice, we don't need FIFOs at the moment.

However, this lays the groundwork for circular scatter gather. This is used for buffer type 0xE (scratch space with circular addressing). Buffer 0xE is used to send output from the GP to the EP. Buffer 0xE is broken in the master revision (and not handled by this PR), but we can easily fix this in a follow-up PR.

I based my code on [the findings I've documented in XboxDevWiki](http://xboxdevwiki.net/DSP#Command_blocks).
The code in this PR has been tested against real hardware, and was confirmed to work in various configurations.

I did not look too closely at input FIFOs yet, so it's possible that the code is broken.
I also didn't add the DMA handler for input FIFOs yet; it will happen in a follow-up PR (probably much later, when we need support).

### DIFF
--- a/hw/xbox/dsp/dsp_dma.c
+++ b/hw/xbox/dsp/dsp_dma.c
@@ -221,6 +221,7 @@ static void dsp_dma_run(DSPDMAState *s)
                  scratch_size);
 
 
+        size_t transfer_size = count * item_size;
         uint8_t* scratch_buf = calloc(count, item_size);
 
         if (direction) {
@@ -241,13 +242,35 @@ static void dsp_dma_run(DSPDMAState *s)
                 }
             }
 
-            // write to scratch memory
-            s->scratch_rw(s->scratch_rw_opaque,
-                scratch_buf, scratch_addr, count*item_size, 1);
+            /* FIXME: Move to function; then reuse for both directions */
+            switch (buf_id) {
+            case 0x0:
+            case 0x1:
+            case 0x2:
+            case 0x3: {
+                unsigned int fifo_index = buf_id;
+                s->fifo_rw(s->rw_opaque,
+                    scratch_buf, fifo_index, transfer_size, 1);
+                break;
+            }
+            case 0xE:
+            case 0xF:
+                s->scratch_rw(s->rw_opaque,
+                    scratch_buf, scratch_addr, transfer_size, 1);
+                break;
+            default:
+                fprintf(stderr, "Unknown DSP DMA buffer: 0x%x\n", buf_id);
+                assert(false);
+                break;
+            }
         } else {
+
+            /* FIXME: Support FIFOs */
+            assert(buf_id == 0xE || buf_id == 0xF);
+
             // read from scratch memory
-            s->scratch_rw(s->scratch_rw_opaque,
-                scratch_buf, scratch_addr, count*item_size, 0);
+            s->scratch_rw(s->rw_opaque,
+                scratch_buf, scratch_addr, transfer_size, 0);
 
             int i;
             for (i=0; i<count; i++) {
--- a/hw/xbox/mcpx_apu.c
+++ b/hw/xbox/mcpx_apu.c
@@ -419,7 +419,7 @@ static void scatter_gather_rw(MCPXAPUState *d,
     unsigned int bytes_to_copy = TARGET_PAGE_SIZE - offset_in_page;
 
     while (len > 0) {
-        assert(page_entry < max_sge);
+        assert(page_entry <= max_sge);
 
         uint32_t prd_address = ldl_le_phys(&address_space_memory,
                                            sge_base + page_entry * 8 + 0);
--- a/hw/xbox/mcpx_apu.c
+++ b/hw/xbox/mcpx_apu.c
@@ -55,7 +55,9 @@
 #define NV_PAPU_XGSCNT                                   0x0000200C
 #define NV_PAPU_VPVADDR                                  0x0000202C
 #define NV_PAPU_GPSADDR                                  0x00002040
+#define NV_PAPU_GPFADDR                                  0x00002044
 #define NV_PAPU_EPSADDR                                  0x00002048
+#define NV_PAPU_EPFADDR                                  0x0000204C
 #define NV_PAPU_TVL2D                                    0x00002054
 #define NV_PAPU_CVL2D                                    0x00002058
 #define NV_PAPU_NVL2D                                    0x0000205C
@@ -66,7 +68,42 @@
 #define NV_PAPU_CVLMP                                    0x00002070
 #define NV_PAPU_NVLMP                                    0x00002074
 #define NV_PAPU_GPSMAXSGE                                0x000020D4
+#define NV_PAPU_GPFMAXSGE                                0x000020D8
 #define NV_PAPU_EPSMAXSGE                                0x000020DC
+#define NV_PAPU_EPFMAXSGE                                0x000020E0
+
+/* Each FIFO has the same fields */
+#define NV_PAPU_GPOFBASE0                                0x00003024
+#   define NV_PAPU_GPOFBASE0_VALUE                          0x00FFFF00
+#define NV_PAPU_GPOFEND0                                 0x00003028
+#   define NV_PAPU_GPOFEND0_VALUE                           0x00FFFF00
+#define NV_PAPU_GPOFCUR0                                 0x0000302C
+#   define NV_PAPU_GPOFCUR0_VALUE                           0x00FFFFFC
+#define NV_PAPU_GPOFBASE1                                0x00003034
+#define NV_PAPU_GPOFEND1                                 0x00003038
+#define NV_PAPU_GPOFCUR1                                 0x0000303C
+#define NV_PAPU_GPOFBASE2                                0x00003044
+#define NV_PAPU_GPOFEND2                                 0x00003048
+#define NV_PAPU_GPOFCUR2                                 0x0000304C
+#define NV_PAPU_GPOFBASE3                                0x00003054
+#define NV_PAPU_GPOFEND3                                 0x00003058
+#define NV_PAPU_GPOFCUR3                                 0x0000305C
+
+/* Fields are same as for the 4 output FIFOs, but only 2 input FIFOs */
+#define NV_PAPU_GPIFBASE0                                0x00003064
+#define NV_PAPU_GPIFEND0                                 0x00003068
+#define NV_PAPU_GPIFCUR0                                 0x0000306C
+#define NV_PAPU_GPIFBASE1                                0x00003074
+#define NV_PAPU_GPIFEND1                                 0x00003078
+#define NV_PAPU_GPIFCUR1                                 0x0000307C
+
+/* Fields, strides and count is same as for GP FIFOs */
+#define NV_PAPU_EPOFBASE0                                0x00004024
+#define NV_PAPU_EPOFEND0                                 0x00004028
+#define NV_PAPU_EPOFCUR0                                 0x0000402C
+#define NV_PAPU_EPIFBASE0                                0x00004064
+#define NV_PAPU_EPIFEND0                                 0x00004068
+#define NV_PAPU_EPIFCUR0                                 0x0000406C
 
 #define NV_PAPU_GPXMEM                                   0x00000000
 #define NV_PAPU_GPMIXBUF                                 0x00005000
@@ -122,6 +159,11 @@ static const struct {
 #   define NV_PAVS_VOICE_TAR_PITCH_LINK_NEXT_VOICE_HANDLE   0x0000FFFF
 
 
+#define GP_OUTPUT_FIFO_COUNT  4
+#define GP_INPUT_FIFO_COUNT   2
+
+#define EP_OUTPUT_FIFO_COUNT  4
+#define EP_INPUT_FIFO_COUNT   2
 
 #define MCPX_HW_MAX_VOICES 256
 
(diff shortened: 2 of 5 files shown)


hw/xbox/dsp/dsp_dma.c
```diff
@@ -157,13 +158,33 @@ static void dsp_dma_run(DSPDMAState *s)
...
+                    scratch_buf, fifo_index, transfer_size, 1);
+                break;
+            }
+            case 0xE:
+            case 0xF:
+                s->scratch_rw(s->rw_opaque,
+                    scratch_buf, scratch_addr, transfer_size, 1);
+                break;
+            default:
+                break;
```
JayFoxRox: Print a message and `assert(false)`

hw/xbox/mcpx_apu.c
```diff
@@ -66,7 +68,23 @@
...
+#define NV_PAPU_GPFMAXSGE                                0x000020D8
 #define NV_PAPU_EPSMAXSGE                                0x000020DC
+#define NV_PAPU_EPFMAXSGE                                0x000020E0
+
+#define NV_PAPU_GPOFBASE0                                0x00003024
+#define NV_PAPU_GPOFEND0                                 0x00003028
+#define NV_PAPU_GPOFCUR0                                 0x0000302C
+#define NV_PAPU_GPIFBASE0                                0x00003064
+#define NV_PAPU_GPIFEND0                                 0x00003068
+#define NV_PAPU_GPIFCUR0                                 0x0000306C
```
JayFoxRox: I don't like the formatting of these blocks. There should be a comment saying that there's 4 output FIFOs and 2 input FIFOs, with a stride of 0x10 bytes between the register sets.

I tried to find a similar case in nv2a but was unable to find a solution. Maybe someone else can find one?

Suggestions how to word this?

hw/xbox/mcpx_apu.c
```diff
@@ -473,6 +491,87 @@ static void ep_scratch_rw(void *opaque,
...
+static void circular_scatter_gather_rw(MCPXAPUState *d,
+                                       hwaddr sge_base, unsigned int max_sge,
+                                       uint8_t *ptr,
+                                       uint32_t base, uint32_t end,
+                                       uint32_t *cur,
+                                       size_t len,
+                                       bool dir)
+{
+    while (len > 0) {
+        unsigned int bytes_to_copy = end - *cur;
```
JayFoxRox: nit: Should be size_t

I decided to use `unsigned int` because I've also used that in the `scatter_gather_rw` (which is similar). I think we should do a larger cleanup in the future, `unsigned int` is fine for now and doesn't raise the question: "why is it size_t here, but not there?".

hw/xbox/mcpx_apu.c
```diff
@@ -473,6 +491,87 @@ static void ep_scratch_rw(void *opaque,
...
+        assert(index < 4);
+        base = d->regs[NV_PAPU_GPOFBASE0 + 0x10 * index];
+        end = d->regs[NV_PAPU_GPOFEND0 + 0x10 * index];
+        cur = &d->regs[NV_PAPU_GPOFCUR0 + 0x10 * index];
+    } else {
+        assert(index < 2);
+        base = d->regs[NV_PAPU_GPIFBASE0 + 0x10 * index];
+        end = d->regs[NV_PAPU_GPIFEND0 + 0x10 * index];
+        cur = &d->regs[NV_PAPU_GPIFCUR0 + 0x10 * index];
+    }
```
JayFoxRox: I consider this ugly. We have a similar block for the EP.

Suggestions?

hw/xbox/mcpx_apu.c
```diff
@@ -473,6 +491,87 @@ static void ep_scratch_rw(void *opaque,
                       ptr, addr, len, dir);
 }
 
+static void circular_scatter_gather_rw(MCPXAPUState *d,
+                                       hwaddr sge_base, unsigned int max_sge,
+                                       uint8_t *ptr,
+                                       uint32_t base, uint32_t end,
+                                       uint32_t *cur,
```
JayFoxRox: nit: This is kind-of ugly. Maybe just make this function return a `uint32_t` with the updated address?

*(On the other hand: this is probably closer to what the hardware does - updating register while writes are happening)*

hw/xbox/mcpx_apu.c
PatrickvL: I've seen this construct:
```
#define NV_PAPU_GPOFCUR(i)                                 (0x0000302C+((i)*0x10))
#define NV_PAPU_GPIFBASE(i)                                (0x00003064+((i)*0x10))
#define NV_PAPU_GPIFEND(i)                                 (0x00003068+((i)*0x10))
#define NV_PAPU_GPIFCUR(i)                                 (0x0000306C+((i)*0x10))
```

Possibly, add and use these?
```
#define NV_PAPU_GP_STRIDE 0x10
#define NV_PAPU_GP_COUNT 4
```

hw/xbox/mcpx_apu.c
```diff
@@ -66,7 +68,23 @@
 #define NV_PAPU_CVLMP                                    0x00002070
 #define NV_PAPU_NVLMP                                    0x00002074
 #define NV_PAPU_GPSMAXSGE                                0x000020D4
+#define NV_PAPU_GPFMAXSGE                                0x000020D8
 #define NV_PAPU_EPSMAXSGE                                0x000020DC
+#define NV_PAPU_EPFMAXSGE                                0x000020E0
+
+#define NV_PAPU_GPOFBASE0                                0x00003024
+#define NV_PAPU_GPOFEND0                                 0x00003028
```
JayFoxRox: Only stores `value & 0xFFFF00`

hw/xbox/mcpx_apu.c
```diff
@@ -66,7 +68,23 @@
 #define NV_PAPU_CVLMP                                    0x00002070
 #define NV_PAPU_NVLMP                                    0x00002074
 #define NV_PAPU_GPSMAXSGE                                0x000020D4
+#define NV_PAPU_GPFMAXSGE                                0x000020D8
 #define NV_PAPU_EPSMAXSGE                                0x000020DC
+#define NV_PAPU_EPFMAXSGE                                0x000020E0
+
+#define NV_PAPU_GPOFBASE0                                0x00003024
```
JayFoxRox: Only stores `value & 0xFFFF00`

hw/xbox/mcpx_apu.c
```diff
@@ -66,7 +68,23 @@
...
+#define NV_PAPU_GPOFEND0                                 0x00003028
+#define NV_PAPU_GPOFCUR0                                 0x0000302C
```
JayFoxRox: Only stores `value & 0xFFFFFC`

hw/xbox/dsp/dsp_dma.c
JayFoxRox: Done.

hw/xbox/dsp/dsp_dma.c
```diff
@@ -241,13 +242,35 @@ static void dsp_dma_run(DSPDMAState *s)
                 }
             }
 
-            // write to scratch memory
-            s->scratch_rw(s->scratch_rw_opaque,
-                scratch_buf, scratch_addr, count*item_size, 1);
+            /* FIXME: Move to function; then reuse for both directions */
+            switch (buf_id) {
+            case 0x0:
```
mborgerson: Prefer named constants here

hw/xbox/dsp/dsp_dma.c
```diff
@@ -241,13 +242,35 @@ static void dsp_dma_run(DSPDMAState *s)
...
+            case 0x1:
+            case 0x2:
+            case 0x3: {
+                unsigned int fifo_index = buf_id;
+                s->fifo_rw(s->rw_opaque,
+                    scratch_buf, fifo_index, transfer_size, 1);
```
mborgerson: nit: Some comment w.r.t. named constants and value for direction field. Helps with readability.

hw/xbox/dsp/dsp_dma.c
```diff
@@ -221,6 +221,7 @@ static void dsp_dma_run(DSPDMAState *s)
                  scratch_size);
 
 
+        unsigned int transfer_size = count * item_size;
```
mborgerson: nit: Prefer `const size_t`

hw/xbox/mcpx_apu.c
```diff
@@ -473,6 +514,120 @@ static void ep_scratch_rw(void *opaque,
...
+
+static void gp_fifo_rw(void *opaque, uint8_t *ptr,
+                       unsigned int index, size_t len,
+                       bool dir)
+{
+    MCPXAPUState *d = opaque;
+    uint32_t base;
+    uint32_t end;
+    hwaddr cur_reg;
+    if (dir) {
```
mborgerson: nit: Again, prefer named constant here

hw/xbox/mcpx_apu.c
```diff
@@ -473,6 +514,120 @@ static void ep_scratch_rw(void *opaque,
...
+    }
+
+    cur = circular_scatter_gather_rw(d,
+        d->regs[NV_PAPU_GPFADDR], d->regs[NV_PAPU_GPFMAXSGE],
+        ptr, base, end, cur, len, dir);
+
+    SET_MASK(d->regs[cur_reg], NV_PAPU_GPOFCUR0_VALUE, cur);
+}
+
+static void ep_fifo_rw(void *opaque, uint8_t *ptr,
```
mborgerson: This function appears to be identical to `gp_fifo_rw`, except for the register offsets. We may want to refactor this into an abstract function. I wouldn't block merge as-is, but food for thought if we develop on it further.

hw/xbox/dsp/dsp_dma.c
JayFoxRox: I agree, but this is a problem across the entire codebase.
Also affects nv2a for example.

We should fix this everywhere in a larger refactor.

hw/xbox/dsp/dsp_dma.c
JayFoxRox: I agree, and will consider changing it to `size_t`. Reason against it might still be what also affected https://github.com/xqemu/xqemu/pull/167#discussion_r244278041.

I don't agree regarding `const`.

hw/xbox/mcpx_apu.c
```diff
@@ -473,6 +514,120 @@ static void ep_scratch_rw(void *opaque,
...
+static uint32_t circular_scatter_gather_rw(MCPXAPUState *d,
+                                           hwaddr sge_base,
+                                           unsigned int max_sge,
+                                           uint8_t *ptr,
+                                           uint32_t base, uint32_t end,
+                                           uint32_t cur,
+                                           size_t len,
+                                           bool dir)
+{
+    while (len > 0) {
```
mborgerson: What if `len` >  `end-base`?

hw/xbox/mcpx_apu.c
JayFoxRox: Yes, this is one of the key design-questions I have had: https://github.com/xqemu/xqemu/pull/167#discussion_r244278110

hw/xbox/mcpx_apu.c
```diff
@@ -473,6 +514,120 @@ static void ep_scratch_rw(void *opaque,
                       ptr, addr, len, dir);
 }
 
+static uint32_t circular_scatter_gather_rw(MCPXAPUState *d,
+                                           hwaddr sge_base,
+                                           unsigned int max_sge,
+                                           uint8_t *ptr,
+                                           uint32_t base, uint32_t end,
```
mborgerson: Is `end` supposed to be inclusive or exclusive?

hw/xbox/dsp/dsp_dma.c
```diff
@@ -241,13 +242,35 @@ static void dsp_dma_run(DSPDMAState *s)
...
+            }
         } else {
+
+            /* FIXME: Support FIFOs */
+            assert(buf_id == 0xE || buf_id == 0xF);
+
             // read from scratch memory
-            s->scratch_rw(s->scratch_rw_opaque,
-                scratch_buf, scratch_addr, count*item_size, 0);
+            s->scratch_rw(s->rw_opaque,
```
mborgerson: Why the rename?

hw/xbox/mcpx_apu.c
```diff
@@ -122,6 +159,10 @@ static const struct {
 #   define NV_PAVS_VOICE_TAR_PITCH_LINK_NEXT_VOICE_HANDLE   0x0000FFFF
 
 
+#define GP_OUTPUT_FIFO_COUNT 4
```
mborgerson: nit: The misalignment hurts my OCD

hw/xbox/dsp/dsp_dma.c
JayFoxRox: Agreed, but can be done later. This is only a small refactor of the existing code.

I also tried naming them in my unit tests and ended up with renaming these constants over and over as I found out new stuff (and some things are still not known).

hw/xbox/mcpx_apu.c
```diff
@@ -55,7 +55,9 @@
 #define NV_PAPU_XGSCNT                                   0x0000200C
```
mborgerson: Can we move register definitions to a mpcx_apu_regs.h? I'm ok for follow up PR doing so.

hw/xbox/dsp/dsp_dma.c
JayFoxRox: Because it's also used for the FIFO now (`s->fifo_rw`)

hw/xbox/mcpx_apu.c
JayFoxRox: Will change

hw/xbox/mcpx_apu.c
JayFoxRox: Yeah, I also want to see this, especially with a lot of VP stuff coming in.
But it should be done in a follow-up.

*(Ideally someone else will do it)*

hw/xbox/mcpx_apu.c
JayFoxRox: I don't remember if I tested this on hardware, but this case does also not occur in the DirectSound code.

I assume the behaviour to be as it is here - and we can fix it later if this assumption is wrong.

hw/xbox/mcpx_apu.c
mborgerson: [ Replying to above comment... ]

hw/xbox/mcpx_apu.c
mborgerson: I consider it ugly too, but for now I think this is the least intrusive way to do it. If we extend functionality and have more duplication between EP/GP, then we should refactor. I think it's fine for now.

hw/xbox/mcpx_apu.c
JayFoxRox: I don't remember but the boundaries should be correct, as this implementation was tested on hardware.

Anyhow, this code strongly implies it's exclusive (`end - cur` or `cur >= end`).

As these are hardware registers (`NV_PAPU_xPxFENDx`), we should document the inclusive / exclusive stuff on XboxDevWiki.
I'm not sure if documenting in the code is necessary (given the strong implications + better docs in the wiki). We also don't document similar stuff in nv2a if it's well defined in other docs (OpenGL specs for example).

hw/xbox/dsp/dsp_dma.c
mborgerson: Why not just use the constants we already have here?

```
    DMA_DIRECTION_TO_DEVICE = 0,
    DMA_DIRECTION_FROM_DEVICE = 1,
```

hw/xbox/mcpx_apu.c
mborgerson: Perhaps we should `assert` to catch instances when our assumption here is wrong.

hw/xbox/dsp/dsp_dma.c
JayFoxRox: Because I wasn't aware of those when I wrote the code.
I was only made aware of the existing QEMU DMA code by espes after this PR was finished. We don't use it in nv2a either.

Note that my code is just a variation of the existing scratch read/write code. So the same signature and code-style was kept.
So the other argument still applies: We should just refactor this in a larger refactor across our entire codebase.

However, for the DSP side, we'd also have to review the terminology in the DSP world before using such constants: the names could be confusing as the DSP might also call the host CPU side "device space" or "peripheral space", as the DSP is often equivalent to the main CPU.
For nv2a it can be equally confusing because some of the DMA is not `FROM_DEVICE` or `TO_DEVICE` but `FROM_CACHE` and `TO_CACHE` etc.

espes also recommended to use the DMA functions internally in the APU DMA code. That seems like a good idea, but needs testing, because this could be performance critical and might need some specific behavior, too.

hw/xbox/mcpx_apu.c
JayFoxRox: If need-be we can just test this on hardware and document it.

But as we have unit testing for this anyway, it's not critical to do before a merge either. Detecting problems here, including debugging this code, is very easy. I have also confirmed this does not occur in DirectSound.

Even *if* this code does not work, we'll be able to *hear* that it's wrong. So errors in this code wouldn't go unnoticed.

hw/xbox/mcpx_apu.c
JayFoxRox: The address is exclusive.

Confirmed on hardware, and also makes a lot of sense as the lower bits of these registers are not stored (which always forces a 0x100 aligned address).

hw/xbox/dsp/dsp_dma.c
JayFoxRox: I changed it to `size_t` but not a `const` for the reasons I've mentioned here.

mborgerson:
Thanks

PR #175, opened 2019-01-09
xid: Port fixes made to xid.c to xid-sdl.c

mborgerson:
The recent fixes to xid.c (32bf810) didn't make it to xid-sdl.c. This
patch ports over those changes. In the future, the parts common to both
xid.c and xid-sdl.c should be factored out so this doesn't happen again.

Original work done by Jannik Vogel (aka JayFoxRox) on Jul 5th, 2018.


JayFoxRox:
I just took the time to finally review the code of this: xid and xid-sdl should be behaving the same now (if no other differences existed before; as I only compared changes this PR did).

There's only some minor code style differences which remain that we should address eventually.

Issue #119, opened 2018-10-05
Halo: Combat Evolved weapon scope does not work.


CYRiXplaysHalo:
NOTE: This is using the dev/halo branch by mborgerson.

When zooming in with a weapon, the screen goes black.

Xbox with pistol no scope.
![xbox_nozoom](https://user-images.githubusercontent.com/43896332/46564163-baa79680-c8d3-11e8-8a0f-469fc3750674.PNG)

Xbox with pistol scope.
![xbox_zoom](https://user-images.githubusercontent.com/43896332/46564164-baa79680-c8d3-11e8-957d-263038d1e16a.PNG)

xqemu with pistol no scope.
![xqemu_nozoom](https://user-images.githubusercontent.com/43896332/46564165-baa79680-c8d3-11e8-8a77-bfd988f0c16a.PNG)

xqemu with pistol scope. Screen goes black besides some HUD elements.
![xqemu_zoom](https://user-images.githubusercontent.com/43896332/46564166-bb402d00-c8d3-11e8-9623-8e316ca002bb.PNG)

PR #176, opened 2019-01-09
nv2a: Fix NV097_SET_VERTEX_DATA2S scaling

### LINKED ISSUE #165: NV097_SET_VERTEX_DATA2S should not scale values
Currently we are scaling vertex data values provided via `NV097_SET_VERTEX_DATA2S` incorrectly. Based on non-exhaustive testing from Halo:CE, these values are apparently passed as simple 16-bit integers (packed into a 32-bit value), which we should then convert to floats without scaling. The game uses this format for providing texture coordinates for several render-to-texture effects such as the radar beacon, weapon scope, and shadows, which are all resolved after removing the scaling.

Related:
* #32 
* #119

mborgerson:
Vertex attribute values provided via NV097_SET_VERTEX_DATA2S are
apparently two 16-bit signed integers, packed into 32 bits, which are
then to be directly mapped to floating point values in the range
[-32768.0, 32767.0].

Halo:CE uses this format to provide texture (u,v) coordinates for
render-to-texture techniques, including weapon scope, dynamic shadows,
and radar beacon--all of which are now working as expected with this
patch (*on branch [mborgerson:dev/halo](https://github.com/mborgerson/xqemu/tree/dev/halo), which allows for some additional 
unimplemented features which we will `assert` for on master branch*).

Fixes #165 
PartiallyFixes #119

### DIFF
--- a/hw/xbox/nv2a/nv2a_pgraph.c
+++ b/hw/xbox/nv2a/nv2a_pgraph.c
@@ -2169,15 +2169,10 @@ static void pgraph_method(NV2AState *d,
     case NV097_SET_VERTEX_DATA2S ...
             NV097_SET_VERTEX_DATA2S + 0x3c: {
         slot = (method - NV097_SET_VERTEX_DATA2S) / 4;
-        assert(false); /* FIXME: Untested! */
         VertexAttribute *attribute = &pg->vertex_attributes[slot];
         pgraph_allocate_inline_buffer_vertices(pg, slot);
-        /* FIXME: Is mapping to [-1,+1] correct? */
-        attribute->inline_value[0] = ((int16_t)(parameter & 0xFFFF) * 2.0 + 1)
-                                         / 65535.0;
-        attribute->inline_value[1] = ((int16_t)(parameter >> 16) * 2.0 + 1)
-                                         / 65535.0;
-        /* FIXME: Should these really be set to 0.0 and 1.0 ? Conditions? */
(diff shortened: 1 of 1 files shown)


JayFoxRox:
Thanks!

[Please document the behaviour in XboxDevWiki](http://xboxdevwiki.net/NV2A/Vertex_attributes#NV097_SET_VERTEX_DATA2S_..._NV097_SET_VERTEX_DATA2S_.2B_0x3c), ideally with comparisons to [similar OpenGL 1.x/2.x features like `glVertex2s`](https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glVertex.xml).
If you aren't sure about the exact range ([-32768.0, 32767.0] vs [-32767.5, 32767.5] etc.) you should add an citation-needed tag or FIXME tag with a note.

Also, are you sure this fixes #119? I believe when I looked at the code, I concluded it was the error I described in my comment, because the sampler is all 0 - the address change might help, but I don't think it's enough to fix the problem.

I'll test later, in particular that Halo issue (but also some other games affected by this).

PatrickvL:
Shouldn't this apply to `NV097_SET_VERTEX_DATA4S_M` as well?

mborgerson:
> Shouldn't this apply to `NV097_SET_VERTEX_DATA4S_M` as well?

Yes probably, I've mentioned it before. But I decided to leave it for a follow up PR. You're welcome to fix it and verify @PatrickvL!

Issue #155, opened 2018-12-26
DSP interpreter is slow


JayFoxRox:
As already mentioned in #151, we should be able to run 106k cycles per APU frame (0.666...ms).

I've created a branch to measure the performance: https://github.com/JayFoxRox/xqemu/pull/22

Running 40000 GP DSP cycles, I get the following (Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz):

**During boot animation (EP is disabled, and GP uses custom code to handle everything):**

```
1 frames: 6287446 ns elapsed (47 ns dryrun, 6287446 ns per frame [3.976193% realtime]), -6037445 ns of budget left
1 frames: 7076872 ns elapsed (44 ns dryrun, 7076872 ns per frame [3.532648% realtime]), -6826871 ns of budget left
1 frames: 6119711 ns elapsed (70 ns dryrun, 6119711 ns per frame [4.085177% realtime]), -5869710 ns of budget left
1 frames: 6037012 ns elapsed (54 ns dryrun, 6037012 ns per frame [4.141138% realtime]), -5787011 ns of budget left
```

**During XDK dashboard (Some work handled by EP, so GP is less busy; running code depends on running sound effects):**

```
1 frames: 6677766 ns elapsed (78 ns dryrun, 6677766 ns per frame [3.743782% realtime]), -6427765 ns of budget left
1 frames: 5734665 ns elapsed (45 ns dryrun, 5734665 ns per frame [4.359470% realtime]), -5484664 ns of budget left
1 frames: 5918075 ns elapsed (48 ns dryrun, 5918075 ns per frame [4.224363% realtime]), -5668074 ns of budget left
1 frames: 6522690 ns elapsed (81 ns dryrun, 6522690 ns per frame [3.832790% realtime]), -6272689 ns of budget left
```

---

The above suggests, that we run only at about ~4% of the expected speed (equivalent to 6.4MHz instead of 160MHz).
40000 cycles appears to be the minimum we should be able to handle, and that'd equate to about 60MHz, so we are still only reaching a bit more than 10% of the speed.

Someone has to profile where we waste our time in `dsp_run` (@mborgerson maybe?). The branch I've linked above should help, because you can arbitrarily raise the number of frames, so more time is spent in the DSP code.

[...]

PR #179, opened 2019-01-15
dsp: Add lookup table for lookup_opcode function

0DaveX:
implements a low hanging fruit from #155 regarding lookup_opcode() performance.

A lookup table is quite fast and since usually only a few different opcodes are being executed on the dsp a lot of cache hits are to be expected.

### DIFF
--- a/hw/xbox/dsp/dsp_cpu.c
+++ b/hw/xbox/dsp/dsp_cpu.c
@@ -159,6 +159,11 @@ static bool match_MMMRRR(uint32_t op)
     return true;
 }
 
+static OpcodeEntry* search_opcode(uint32_t op);
+
+/*nonparallel opcodes only occupy the lower 20bits*/
+OpcodeEntry *OpcodeEntryArray[0xFFFFF+1];
+
 static const OpcodeEntry nonparallel_opcodes[] = {
     { "0000000101iiiiii1000d000", "add #xx, D", dis_add_imm, emu_add_imm },
     { "00000001010000001100d000", "add #xxxx, D", dis_add_long, emu_add_long },
@@ -378,6 +383,10 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
             nonparallel_matches[i][0] = mask;
             nonparallel_matches[i][1] = match;
         }
+
+        for(i=0; i < ARRAYSIZE(OpcodeEntryArray); i++) {
+            OpcodeEntryArray[i] = search_opcode(i);
+        }
     }
 
     /* Memory */
@@ -420,21 +429,28 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
     dsp->disasm_prev_inst_pc = 0xFFFFFFFF;
 }
 
-static OpcodeEntry lookup_opcode(uint32_t op) {
-    OpcodeEntry r = {0};
+static OpcodeEntry* search_opcode(uint32_t op)
+{
+    OpcodeEntry* r;
     int i;
     for (i=0; i<ARRAYSIZE(nonparallel_opcodes); i++) {
         if ((op & nonparallel_matches[i][0]) == nonparallel_matches[i][1]) {
             if (nonparallel_opcodes[i].match_func 
                 && !nonparallel_opcodes[i].match_func(op)) continue;
-            if (r.template != NULL) {
-                printf("qqq %x %s\n", op, r.template);
+            if (r->template != NULL) {
+                //printf("qqq %x %s\n", op, r.template);
             }
-            assert(r.template == NULL);
-            r = nonparallel_opcodes[i];
+            //assert(r.template == NULL);
+            r = &nonparallel_opcodes[i];
         }
     }
     return r;
+}
+
+static inline OpcodeEntry* lookup_opcode(uint32_t op)
+{
+    assert(op < ARRAYSIZE(OpcodeEntryArray));
+    return OpcodeEntryArray[op];
 }
 
 static uint16_t disasm_instruction(dsp_core_t* dsp, dsp_trace_disasm_t mode)
@@ -459,12 +475,12 @@ static uint16_t disasm_instruction(dsp_core_t* dsp, dsp_trace_disasm_t mode)
     dsp->disasm_parallelmove_name[0] = 0;
 
     if (dsp->disasm_cur_inst < 0x100000) {
-        const OpcodeEntry op = lookup_opcode(dsp->disasm_cur_inst);
-        if (op.template) {
-            if (op.dis_func) {
-                op.dis_func(dsp);
+        OpcodeEntry* op = lookup_opcode(dsp->disasm_cur_inst);
+        if (op->template) {
+            if (op->dis_func) {
+                op->dis_func(dsp);
             } else {
-                sprintf(dsp->disasm_str_instr, "%s", op.name);
+                sprintf(dsp->disasm_str_instr, "%s", op->name);
             }
         } else {
             dis_undefined(dsp);
@@ -659,11 +675,11 @@ void dsp56k_execute_instruction(dsp_core_t* dsp)
     }
             
     if (dsp->cur_inst < 0x100000) {
-        const OpcodeEntry op = lookup_opcode(dsp->cur_inst);
-        if (op.emu_func) {
-            op.emu_func(dsp);
+        OpcodeEntry* op = lookup_opcode(dsp->cur_inst);
+        if (op->emu_func) {
+            op->emu_func(dsp);
         } else {
-            printf("%x - %s\n", dsp->cur_inst, op.name);
+            printf("%x - %s\n", dsp->cur_inst, op->name);
             emu_undefined(dsp);
         }
     } else {


Ernegien:
A 1,048,575 element array of OpcodeEntry objects is surely not what they meant by lookup table; perhaps a basic map or smaller lookup tables broken down into instructions and arguments would be better suited here?

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -420,23 +431,29 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
...
     for (i=0; i<ARRAYSIZE(nonparallel_opcodes); i++) {
         if ((op & nonparallel_matches[i][0]) == nonparallel_matches[i][1]) {
             if (nonparallel_opcodes[i].match_func 
                 && !nonparallel_opcodes[i].match_func(op)) continue;
             if (r.template != NULL) {
-                printf("qqq %x %s\n", op, r.template);
+                //printf("qqq %x %s\n", op, r.template);
             }
-            assert(r.template == NULL);
+            //assert(r.template == NULL);
```
JayFoxRox: The way these are disabled is bad.

Also we still want to capture these cases, so you should return some meta-instruction handler which can `printf` and `assert` if necessary.

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -420,23 +431,29 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
...
             r = nonparallel_opcodes[i];
         }
     }
     return r;
 }
 
+static OpcodeEntry lookup_opcode(uint32_t op)
+{
+    return OpcodeEntryArray[op];
+}
```
JayFoxRox: With the changed design it might not be necessary to have this as part of a function, so it could be inline probably?

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -159,6 +159,10 @@ static bool match_MMMRRR(uint32_t op)
     return true;
 }
 
+static OpcodeEntry lookup_opcode_manually(uint32_t op);
+OpcodeEntry OpcodeEntryArray[1048575];
```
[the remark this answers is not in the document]
0DaveX: do you mean 0xFFFFF?
0xFFFFF /  1048575 was used on purpose since the first 4 bit of an opcode seem to be always 0.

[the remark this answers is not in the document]
0DaveX: Thanks, took care

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -159,6 +159,10 @@ static bool match_MMMRRR(uint32_t op)
...
+static bool OpcodeEntryArray_initialised;
```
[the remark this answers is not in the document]
0DaveX: dsp_cpu reset is called relatively often so it does make sense to do this only once.
agree re `matches_initialised` - but what would be a good generic name?

`dsp_data_initialised?`

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -380,6 +384,13 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
         }
     }
 
+    if(!OpcodeEntryArray_initialised) {
+        OpcodeEntryArray_initialised = true;
+        for(int i=0; i < 1048575; i++) {
```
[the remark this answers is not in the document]
0DaveX: Thanks done

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -420,23 +431,29 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
     dsp->disasm_prev_inst_pc = 0xFFFFFFFF;
 }
 
-static OpcodeEntry lookup_opcode(uint32_t op) {
+static OpcodeEntry lookup_opcode_manually(uint32_t op)
```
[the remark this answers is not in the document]
0DaveX: is `opcode_identify` good?

hw/xbox/dsp/dsp_cpu.c
0DaveX: could this be done in a ifdef debug like manner?

hw/xbox/dsp/dsp_cpu.c
0DaveX: Yeah :thumbsup:

0DaveX:
> A 1,048,575 element array of OpcodeEntry objects is surely not what they meant by lookup table; perhaps a basic map or smaller lookup tables broken down into instructions and arguments would be better suited here?

Well I came up with the lookup table :P
But yeah 1,048,575 elements and an array size of 8MB seem to shy away people.
But imo 8mb is not that much and an acceptable overhead wich in turn leads to good performance (in the lookup function)
smaller lookup tables or even some sort of a  map would require more logic and will be slower.

0DaveX:
> I agree with @Ernegien. I think we even talked about this on Discord: the simplest form of a table (on 64-bit systems, with 64-bit pointers) is 1048575 * 8 bytes = 8 MegaBytes. This implementation uses a structure which is 5 pointers long, so this is around 40 MegaBytes. We also have 2 DSPs in the Xbox, so your code consumes 80 MegaBytes of memory: more RAM than the original Xbox had in total.
> 
> Critically, this isn't very good for cache performance (and we talked about this on Discord, [with this being linked](https://gist.github.com/jboner/2841832)).
> 
> @Ernegien gave a good suggestion in splitting one table into multiple small tables. This is an okay solution while we still have a complicated interpreter loop (which does many housekeeping tasks, aside from actually running instructions = lookup is slow, but not the major bottleneck).
> However, ideally we simply wouldn't do the lookup (and not even the entire interpreter loop) for each single instruction (even in loops where the same lookup is repeated).
> 
> Therefore, a much better style of "lookup"-table, is a second DSP memory array which mirrors the program-word array. Whenever the program-word array would be changed, we'd update this shadow-copy with a new pointer to the instruction handler. One could also group some other parts of the instruction decoder main-loop in such handlers on critical points (like in branches or instructions which pause, but not in hot-code). It could even be considered as a first step towards a JIT (decoupling decoding and running, possibly establishing basic-blocks).
> 
> Alternatively, we could even use TCG to generate this shadow instruction stream (with each instruction handler being like the existing TCG-helpers). This would be a direct step towards a JIT. However, I'm not experienced with TCG, so there might be problems which prevent this.
> 
> With this said, I'm doubtful about this getting merged. I'm still going to review the code (so you can learn from mistakes, and we _could_ theoretically merge this).

First: Thanks for the reviews! So I can learn and defend the rp.

The lookup table array was updated now to only use 8mb and store pointers to OpcodeEntries.

>"We have to DSPs"< a

they still use the same static lookup table don't they?

> Critically, this isn't very good for cache performance

I may be wrong here but I'dont agree.
As stated usually only a few different opcodes are used (~200)
The whole array doesn't have to fit into cpu cache - just the portions of the array we use often.
And since these portions are only a few it easily fits into L1/2 cache.
e.g. 200 cachelines used is not much imo.

I did measure <10ns mean for lookup_opcode() calls - and don't think other solutions will be faster - but I'd be glad if the measurements would proof me wrong.

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -420,23 +431,29 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
     dsp->disasm_prev_inst_pc = 0xFFFFFFFF;
 }
 
-static OpcodeEntry lookup_opcode(uint32_t op) {
-    OpcodeEntry r = {0};
+static OpcodeEntry* opcode_identify(uint32_t op)
```
JayFoxRox: The existing naming scheme is `activity_object` so `identify_opcode`.

The name is fine. Personally, I prefer exposing *some* inner workings in the name to warn users about characteristics.
`identify_opcode` sounds like magical instant-return oracle (choice of implementation might allow this: like the array lookup wrapper you have created).
`search_opcode` sounds like it will take a while (and is literally what this does, we have no implementation choice to avoid searching).

hw/xbox/dsp/dsp_cpu.c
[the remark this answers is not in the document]
0DaveX: alright :white_check_mark:

hw/xbox/dsp/dsp_cpu.c
0DaveX: oky, changed it to `search_opcode`

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -159,6 +159,10 @@ static bool match_MMMRRR(uint32_t op)
     return true;
 }
 
+static OpcodeEntry* opcode_identify(uint32_t op);
+OpcodeEntry *OpcodeEntryArray[0xFFFFF];
```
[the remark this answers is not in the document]
0DaveX: thanks

hw/xbox/dsp/dsp_cpu.c
0DaveX: :+1:  done

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -159,6 +159,9 @@ static bool match_MMMRRR(uint32_t op)
     return true;
 }
 
+static OpcodeEntry* search_opcode(uint32_t op);
+OpcodeEntry *OpcodeEntryArray[0xFFFFF+1]; //for the lower 20bits of the opcodes
```
JayFoxRox: The interesting comment would be: *why* only the lower 20 bits?

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -420,23 +427,30 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
...
+            //assert(r.template == NULL);
+            r = &nonparallel_opcodes[i];
         }
     }
     return r;
 }
 
+static inline OpcodeEntry* lookup_opcode(uint32_t op)
+{
+    assert(op <= ARRAYSIZE(OpcodeEntryArray));
```
JayFoxRox: Off-by-one again.

0DaveX: thx, fixed

hw/xbox/dsp/dsp_cpu.c
0DaveX: alright 
hope this is better: `/*nonparallel opcodes only occupy the lower 20bits*/
	`

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -420,23 +429,30 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
...
+{
+    OpcodeEntry* r;
     int i;
     for (i=0; i<ARRAYSIZE(nonparallel_opcodes); i++) {
         if ((op & nonparallel_matches[i][0]) == nonparallel_matches[i][1]) {
             if (nonparallel_opcodes[i].match_func 
                 && !nonparallel_opcodes[i].match_func(op)) continue;
-            if (r.template != NULL) {
-                printf("qqq %x %s\n", op, r.template);
+            if (r->template != NULL) {
```
JayFoxRox: Did you try running this code?

`r` is not initialized at this point (should be `r = NULL`).

If you change from the object, which was `{0}` initialized, to pointers, then you have to do a different check here: `(r != NULL) && (r->template != NULL)` to make sure that no other opcode was found yet (which seems to be the idea behind this print), and to make sure it has a template.

You , as you have `assert(r->template == NULL)`

hw/xbox/dsp/dsp_cpu.c
JayFoxRox: Making assumptions about the idea behind these, you can keep them here (but enable them).

This check must also be for `(r != NULL) && (r.template == NULL)`, otherwise `r` might be unset.

As the printf is closely related to this, I suggest moving the `r != NULL` check into a separate condition to handle this debug stuff, maybe even both of them and an ugly `assert(false)`.

---

If all `OpcodeEntry`s have a template set (I'd assume so), you can reduce these checks to `r == NULL` / `r != NULL`.

In that case, `r.template == NULL` would have been a check if `r` was ever modified after zero-intialization (or: if the loop found more than one result). For pointers this just translates to `r == NULL`
Similar for the `printf` before: the assumption is that if `r.template != NULL` then there was a duplicate, so it should be printed, so it can be debugged.

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -159,6 +159,11 @@ static bool match_MMMRRR(uint32_t op)
     return true;
 }
 
+static OpcodeEntry* search_opcode(uint32_t op);
+
+/*nonparallel opcodes only occupy the lower 20bits*/
+OpcodeEntry *OpcodeEntryArray[0xFFFFF+1];
```
PatrickvL: Given how `OpcodeEntryArray` is used, and the other symbol names, it might be clearer to call it `nonparallel_opcode_lookup_array`.

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -420,23 +429,30 @@ void dsp56k_reset_cpu(dsp_core_t* dsp)
     dsp->disasm_prev_inst_pc = 0xFFFFFFFF;
 }
 
-static OpcodeEntry lookup_opcode(uint32_t op) {
-    OpcodeEntry r = {0};
+static OpcodeEntry* search_opcode(uint32_t op)
+{
+    OpcodeEntry* r;
```
PatrickvL: In addition to what @JayFoxRox wrote, initialize `r` (like this: `OpcodeEntry* r = NULL;`) to avoid corruption due to potential garbage data.

mborgerson:
This PR has been open for a couple of weeks now without any movement. Though intentions are good, I think the specific approach used here will cause further performance regression, opposite of the goal. I'm going to close this for now until we can come back with a better approach that can demonstrably achieve the goals of #155.

JayFoxRox:
To be fair: it wasn't open that long, and @0DaveX kept improving the implementation quickly in the first few hours. However, it also didn't see updates or feedback in the past week, with no information where it's heading, so closing it was the right choice.

I also still agree with @mborgerson that the approach wasn't the best, and if it's ever improved (design-wise) a new PR can be sent.

0DaveX:
> This PR has been open for a couple of weeks now without any movement. Though intentions are good, I think the specific approach used here will cause further performance regression, opposite of the goal. I'm going to close this for now until we can come back with a better approach that can demonstrably achieve the goals of #155.

@mborgerson I can't see this PR to cause "performance regression". But I'd be interested in possible causes. This **did** improve performance.

You wrote in xbox-dev discord (2019.01.08)
`rough profiling suggests lookup_opcode is indeed very slow`
[discord message](https://discordapp.com/channels/428359196719972353/428359372842991616/532304366317862932)

gperf with pr:
https://hastebin.com/ewizowigad.sql

without my pr:
https://hastebin.com/adamugadap.sql

the worst offender was pretty much eliminated i guess?

RE: lack of activity
it's true I didn't post/pushed anything here for 12 days.
But that was because I couldn't yet fix a presumed bug in the original opcode matching code.

@JayFoxRox wrote: 

> The way these are disabled is bad.
>Also we still want to capture these cases
RE: https://github.com/0DaveX/xqemu/blob/8ae762526d1d5611b9660336d51ceaac657f4655/hw/xbox/dsp/dsp_cpu.c#L443

That imho extended the scope of this PR 
in the current code it looks like some opcodes are not identified correctly.
e.g.
`opcode = 0xA7080 == 10100111000010000000
{ "0000101001MMMRRR1S00bbbb", "jclr #n, [X or Y]:ea, xxxx", dis_jclr_ea, emu_jclr_ea, match_MMMRRR },
{ "0000101001110RRR1WDDDDDD", "move X:(Rn + xxxx) <-> R", dis_move_x_long, emu_move_x_long },
` 
(opcode "identified" as the both above)
I couldn't fix this "soon enough" but also didn't communicate it here, wich was probably a mistake.
(Didn't expect it to be closed right away tho)

I don't see any reasons for "performance regressions" and did see improvements in the PRed code measuring with:
https://github.com/JayFoxRox/xqemu/commit/59b6d71f202cfd76c576f7ef0aa924161e5f0475

So I'm surprised by this decision without discussion.

mborgerson:
@0DaveX: Thank you for taking the time to profile your changes! Coming with hard results to a PR is always good. As you've pointed out, I did mention this function specifically and your data shows that it does take a lot of time.

It's true, you've reduced the lookup time from it's original *abysmal* full-array lookup to a constant time lookup, according to your profiling data from 60% down to 1.6% (well done!). But this was done at the cost of an 8 MiB array of pointers (as @Ernegien pointed out) for a relatively simple operation.

My hunch is that this can be done better, and without such a relatively high memory overhead. Do you think it's possible to reduce this array to smaller subset? What about the mirror lookup table approach suggested by @JayFoxRox above that lets us create a jump table--getting us closer to a JiT approach?

As always, PRs are open to discussion, even after closing. Thank you for contributing!

0DaveX:
@mborgerson Thanks for the reply!
alright the 8MiB extra memory usage are considered too much. (for this simple function)
And obviously I agree that other optimizations/ design changes will help much more and "getting us closer to a JiT" will be necessary to solve #155.

Issue #182, opened 2019-01-25
Trigger range is bad in xid-sdl


JayFoxRox:
https://github.com/xqemu/xqemu/blob/47452495c00b48a2bdcc252c3dacd62eb2929c29/hw/xbox/xid-sdl.c#L273-L280

The state is shifted by 8 bits. However, according to https://wiki.libsdl.org/SDL_GameControllerGetAxis:

> Triggers, however, range from 0 to 32767 (they never return a negative value).

32767 is 0x7FFF, so we'd be using `0x7FFF >> 8` which is 0x7F.

I assume Xbox triggers range from 0 to 0xFF. So the effective range of our triggers is only roughly 50% of the intended range.

We have gotten reports that triggers don't work in "Jet Set Radio Future" and I believe this could have to do with a 50% deadzone (although I did not confirm this).

The correct shift amount would be 7 bits, because `0x7FFF >> 7` is 0xFF, which would result in the full range.

PR #183, opened 2019-01-25
xid: Fix trigger range in xid-sdl

CakeLancelot:
Fixes xqemu/xqemu#182

Tested and the deadzone issue seems to be fixed in jJet Set Radio Future.


GXTX:
This PR fixes the triggers for me. Thank you!

JayFoxRox:
Waiting for rename of commit.

I'd also rather not have commits link to specific GitHub issues as the git should be independent of the GitHub infrastructure + GitHub quickly has bugs with showing the commit in wrong contexts.
An empty description is fine for such small / logical changes.

PR #184, opened 2019-01-25
smbus: Fix eeprom wordwrite bug

dracc:
This solution (and problem) was found by @0DaveX and submitted as part of #122 .


JayFoxRox:
nit: This shouldn't only fix eeprom bugs, but also bugs in other components.

---

What's wrong with CI?

PR #185, opened 2019-01-28
build.sh: Use Python 3.x for Windows

GXTX:
Python 2.x is turning EOL at the end of the year. All of the QEMU scripts are already 3.X compatible the only thing not compatible is our `get_deps.py` for copying Windows DLL for redistribution which is also fixed in this PR.

I've not done extensive testing but everything builds and runs fine.


mborgerson:
Thanks

PR #186, opened 2019-01-29
nv2a: Set content of disabled register combiner stages to zero

### LINKED ISSUE #168: ShaderState contains garbage from deactivated register combiner stages
`ShaderState` is the input to our shader generator, and it's hashed to look-up existing shaders from the cache.
So we must try to keep our `ShaderState` as clean as possible, so that only bits that actually contribute to the shader logic are stored.

We currently have a bug where we write the configuration for all 8 register combiner stages, even if we use fewer stages: https://github.com/xqemu/xqemu/blob/02f3b701d0869506c213dea524b01628f07bfb3f/hw/xbox/nv2a/nv2a_pgraph.c#L3091

This can lead to the following problem:

1 combiner stage is used, but the rest 7 are filled with garbage from other (previously drawn) objects.
So we get different combinations of the same shader, all doing the same thing. And if objects are drawn in a different order (changing the unused combiner stage), we generate a new shader.

The garbage that's stored for the deactivated stages, is not used in the shader generation either, so we can just keep it zero'd.

---

We should also review if other code can easily be modified in the same fashion.

thrimbor:
This fixes #168 by memset()-ing the whole arrays and only copying the content of enabled register combiner stages.

### DIFF
--- a/hw/xbox/nv2a/nv2a_pgraph.c
+++ b/hw/xbox/nv2a/nv2a_pgraph.c
@@ -3083,7 +3083,15 @@ static void pgraph_bind_shaders(PGRAPHState *pg)
         last_y = y;
     }
 
-    for (i = 0; i < 8; i++) {
+    /* FIXME: We should memset(state, 0x00, sizeof(state)) instead */
+    memset(state.psh.rgb_inputs, 0, sizeof(state.psh.rgb_inputs));
+    memset(state.psh.rgb_outputs, 0, sizeof(state.psh.rgb_outputs));
+    memset(state.psh.alpha_inputs, 0, sizeof(state.psh.alpha_inputs));
+    memset(state.psh.alpha_outputs, 0, sizeof(state.psh.alpha_outputs));
+
+    /* Copy content of enabled combiner stages */
+    int num_stages = pg->regs[NV_PGRAPH_COMBINECTL] & 0xFF;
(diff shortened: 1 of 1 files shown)


JayFoxRox:
This is fine with me, I was waiting for @mborgerson to accept it, because he initially raised concerns.

For context:

> `FIXME: We should memset(state, 0x00, sizeof(state)) instead`

We want to do that, so we don't leave uninitialized alignment data between fields, which affects the hash ([as happend for Citra](https://github.com/citra-emu/citra/pull/1704)).
For Citra this unitialized data randomly started happening after some commit. And it only happened in some compiler versions (if I remember correctly); XQEMU is not affected yet, but it might be in the future.

We also have an issue about this in #187.

Because we use struct initialization, we can't currently do the entire `memset` without restructuring code a bit more.

As @thrimbor originally only wanted to do a quickfix for #186, I'm fine with this temporary incomplete solution, as long as we also keep track of #187. I was the one who told him to `memset` the PSH fields instead of looping over the unused fields, because the `memset` is closer to our intended method to fix #187 (hence, also the comment).

mborgerson:
Thank you for taking the initiative @thrimbor! I'd prefer for us fix by `memset` of entire struct, but this is better than master so I think it's fine to merge. We can clean it up after. I'll probably get to it when I merge the performance stuff.

Issue #158, opened 2018-12-26
DSP CPU memory access is slow


JayFoxRox:
We use `dsp56k_read_memory` and `dsp56k_write_memory` to access the DSP address space.
However, those functions only transfer 1 word at a time.

During a DMA transfer, we access hundreds of words. So the addressing is repeated every time, and each call adds overhead.

We should either make it possible to obtain a pointer to the buffers, or at least make it possible to specify an element count, to read multiple items at once (possibly using the same access options that are used in DMA transfers).

PR #190, opened 2019-02-01
apu: Add VP framework

JayFoxRox:
This PR adds a small framework for implementing VP voice processing (which will happen in a future PR).
This framework includes VP &rarr; GP audio routing. Unfortunately this transfer is affected by #158 for 1024 words (32 channels * 32 samples).

For audio debugging purposes, this also adds a tone generator for injecting audio into the GP.
This tone isn't currently heard, because EP and ACI are severely broken (so audio isn't routed to speakers).
However, this tone will be useful in the future, to verify correct GP / EP behaviour, and can also be used to test audio-output without functional VP voice processing.

### DIFF
--- a/hw/xbox/mcpx_apu.c
+++ b/hw/xbox/mcpx_apu.c
@@ -2,7 +2,7 @@
  * QEMU MCPX Audio Processing Unit implementation
  *
  * Copyright (c) 2012 espes
- * Copyright (c) 2018 Jannik Vogel
+ * Copyright (c) 2018-2019 Jannik Vogel
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
@@ -24,6 +24,9 @@
 #include "hw/pci/pci.h"
 #include "cpu.h"
 #include "hw/xbox/dsp/dsp.h"
+
+#define NUM_SAMPLES_PER_FRAME 32
+#define NUM_MIXBINS 32
 
 #include "hw/xbox/mcpx_apu.h"
 
@@ -804,12 +807,28 @@ static const MemoryRegionOps ep_ops = {
     .write = ep_write,
 };
 
+static void process_voice(MCPXAPUState *d,
+                          int32_t mixbins[NUM_MIXBINS][NUM_SAMPLES_PER_FRAME],
+                          uint32_t voice)
+{
+    /* FIXME: Implement */
+}
+
+/* This routine must run at 1500 Hz */
 /* TODO: this should be on a thread so it waits on the voice lock */
 static void se_frame(void *opaque)
 {
     MCPXAPUState *d = opaque;
+    int mixbin;
+    int sample;
+
     timer_mod(d->se.frame_timer, qemu_clock_get_ms(QEMU_CLOCK_VIRTUAL) + 10);
     MCPX_DPRINTF("mcpx frame ping\n");
+
+    /* Buffer for all mixbins for this frame */
+    int32_t mixbins[NUM_MIXBINS][NUM_SAMPLES_PER_FRAME] = { 0 };
+
+    /* Process all voices, mixing each into the affected MIXBINs */
     int list;
     for (list = 0; list < 3; list++) {
         hwaddr top, current, next;
@@ -828,12 +847,24 @@ static void se_frame(void *opaque)
                     NV_PAVS_VOICE_PAR_STATE_ACTIVE_VOICE)) {
                 MCPX_DPRINTF("voice %d not active...!\n", d->regs[current]);
                 fe_method(d, SE2FE_IDLE_VOICE, d->regs[current]);
+            } else {
+                process_voice(d, mixbins, d->regs[current]);
             }
             MCPX_DPRINTF("next voice %d\n", d->regs[next]);
             d->regs[current] = d->regs[next];
         }
     }
 
+    /* Write VP results to the GP DSP MIXBUF */
+    for (mixbin = 0; mixbin < NUM_MIXBINS; mixbin++) {
+        for (sample = 0; sample < NUM_SAMPLES_PER_FRAME; sample++) {
+            dsp_write_memory(d->gp.dsp,
+                             'X', GP_DSP_MIXBUF_BASE + mixbin * 0x20 + sample,
+                             mixbins[mixbin][sample] & 0xFFFFFF);
+        }
+    }
+
+    /* Kickoff DSP processing */
     if ((d->gp.regs[NV_PAPU_GPRST] & NV_PAPU_GPRST_GPRST)
         && (d->gp.regs[NV_PAPU_GPRST] & NV_PAPU_GPRST_GPDSPRST)) {
         dsp_start_frame(d->gp.dsp);
--- a/hw/xbox/mcpx_apu.c
+++ b/hw/xbox/mcpx_apu.c
@@ -24,6 +24,7 @@
 #include "hw/pci/pci.h"
 #include "cpu.h"
 #include "hw/xbox/dsp/dsp.h"
+#include <math.h>
 
 #define NUM_SAMPLES_PER_FRAME 32
 #define NUM_MIXBINS 32
@@ -186,6 +187,9 @@ static const struct {
 #else
 # define MCPX_DPRINTF(format, ...)       do { } while (0)
 #endif
+
+/* More debug functionality */
+#define GENERATE_MIXBIN_BEEP      0
 
 typedef struct MCPXAPUState {
     PCIDevice dev;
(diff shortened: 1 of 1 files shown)


hw/xbox/mcpx_apu.c
```diff
@@ -826,12 +845,38 @@ static void se_frame(void *opaque)
...
+    /* Inject some audio to the mixbin for debugging.
+     * Signal is 1500 Hz sine wave, phase shifted by mixbin number.*/
+    for (unsigned int i = 0; i < NUM_MIXBINS; i++) {
+        for (unsigned int j = 0; j < NUM_SAMPLES_PER_FRAME; j++) {
+            /* Avoid multiple of 1.0 / NUM_SAMPLES_PER_FRAME for phase shift,
+             * or waves cancel out */
+            float offset = j / (float)NUM_SAMPLES_PER_FRAME -
+                           i / (float)(NUM_SAMPLES_PER_FRAME + 1);
+            float wave = sinf(offset * M_PI * 2.0f);
+            mixbin[i][j] += wave * 0x3FFFFF;
```
JayFoxRox: This used to be 0x7FFFFF during testing (which was way too loud and sounded like a fire alarm).
I have not tested if this new volume is high enough - I will do that soon.

hw/xbox/mcpx_apu.c
```diff
@@ -826,12 +845,38 @@ static void se_frame(void *opaque)
...
                 fe_method(d, SE2FE_IDLE_VOICE, d->regs[current]);
+            } else {
+                process_voice(d, mixbin, d->regs[current]);
             }
             MCPX_DPRINTF("next voice %d\n", d->regs[next]);
             d->regs[current] = d->regs[next];
         }
     }
 
+#if INJECT_VP_BEEP
```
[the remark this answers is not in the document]
JayFoxRox: [Actually.. this shouldn't be necessary](https://www.cs.sfu.ca/~tamaras/sinusoids318/Adding_two_sinusoids.html).

It's also easy enough to edit it *if* it becomes necessary.

hw/xbox/mcpx_apu.c
```diff
@@ -826,12 +845,38 @@ static void se_frame(void *opaque)
...
+    /* Inject some audio to the mixbin for debugging.
+     * Signal is 1500 Hz sine wave, phase shifted by mixbin number.*/
+    for (unsigned int i = 0; i < NUM_MIXBINS; i++) {
+        for (unsigned int j = 0; j < NUM_SAMPLES_PER_FRAME; j++) {
```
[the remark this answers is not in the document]
JayFoxRox: I chose `int mixbin` and `int sample`; I'm not really happy about this change, because I feel like non-C99 loops are error prone, and I'll probably have to add some linebreaks to future PRs.

Also we might want to do `uint32_t* mixbin = &mixbins[i][0];` which is now.. bad.. because it mixes indices and pointers to subsets of an array.

I also had to rename the original `mixbin` variable to `mixbins` (which is a good change though).

hw/xbox/mcpx_apu.c
```diff
@@ -826,12 +849,38 @@ static void se_frame(void *opaque)
...
+            float offset = mixbin / (float)NUM_SAMPLES_PER_FRAME -
+                           sample / (float)(NUM_SAMPLES_PER_FRAME + 1);
+            float wave = sinf(offset * M_PI * 2.0f);
+            mixbins[mixbin][sample] += wave * 0x3FFFFF;
+        }
+    }
+#endif
+
+    /* Write VP results to the GP DSP MIXBUF */
+    for (mixbin = 0; mixbin < NUM_MIXBINS; mixbin++) {
```
mborgerson: It would be nice if we could move the 4k of mix bin data off the stack and write directly into DSP memory here.

hw/xbox/mcpx_apu.c
```diff
@@ -802,12 +809,28 @@ static const MemoryRegionOps ep_ops = {
...
 {
     MCPXAPUState *d = opaque;
+    int mixbin;
+    int sample;
+
     timer_mod(d->se.frame_timer, qemu_clock_get_ms(QEMU_CLOCK_VIRTUAL) + 10);
     MCPX_DPRINTF("mcpx frame ping\n");
+
+    /* Buffer for all mixbins for this frame */
+    int32_t mixbins[NUM_MIXBINS][NUM_SAMPLES_PER_FRAME] = { 0 };
```
mborgerson: It would be nice if we could move the 4k of mix bin data off the stack and write directly into DSP memory here.

hw/xbox/mcpx_apu.c
```diff
@@ -826,12 +849,38 @@ static void se_frame(void *opaque)
...
+            float wave = sinf(offset * M_PI * 2.0f);
+            mixbins[mixbin][sample] += wave * 0x3FFFFF;
+        }
+    }
+#endif
+
+    /* Write VP results to the GP DSP MIXBUF */
+    for (mixbin = 0; mixbin < NUM_MIXBINS; mixbin++) {
+        for (sample = 0; sample < NUM_SAMPLES_PER_FRAME; sample++) {
+            dsp_write_memory(d->gp.dsp, 'X', 0x001400 + mixbin * 0x20 + sample,
```
mborgerson: Prefer `DSP_MIXBUFFER_BASE` instead of `0x001400` here

hw/xbox/mcpx_apu.c
JayFoxRox: See original PR description and issue #158.

We also do not only have to do 1024 writes.. it would be more: Each voice (up to 256) will access these mixbins (both read and write; up to 8 times). This access happens in `process_voice` in the future.

By having this temporary array, we avoid costly DSP writes, and only do it once when the final value is determined.

The unused bins seem to be cleared to zero, so we do at least 1024 writes to the DSP memory anyway.
The update is *probably* also atomic at frame start, so we need a temporary buffer.

So if any of these *strong* assumptions (based on memory sniffing on hardware), holds true, then we can't avoid this access; especially if we want multi-threading in the future.

hw/xbox/mcpx_apu.c
JayFoxRox: I agree and will add it; I'll also change it for the existing occurrences.

[I originally didn't do it because we weren't sure about the naming / correctness](https://github.com/xqemu/xqemu/pull/92/files#r205284029) (we still aren't sure, but at least a bit more confident).

Ideally #158 (or something related) will solve all of these issues by providing a virtual memory space for MIXBUF. 
This would be similar to how it's done for MMIO, which has different sections for the standard DSP56k X, Y, P, but also the GP exclusive MIXBUF. The `char` based naming is horrible.

hw/xbox/mcpx_apu.c
```diff
@@ -826,12 +849,38 @@ static void se_frame(void *opaque)
...
             d->regs[current] = d->regs[next];
         }
     }
 
+#if GENERATE_MIXBIN_BEEP
+    /* Inject some audio to the mixbin for debugging.
+     * Signal is 1500 Hz sine wave, phase shifted by mixbin number. */
+    for (mixbin = 0; mixbin < NUM_MIXBINS; mixbin++) {
+        for (sample = 0; sample < NUM_SAMPLES_PER_FRAME; sample++) {
+            /* Avoid multiple of 1.0 / NUM_SAMPLES_PER_FRAME for phase shift,
```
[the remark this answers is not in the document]
JayFoxRox: As this has no information across frames, it can only do multiples of 1500 Hz: 3000 Hz, 4500 Hz, ...
Doing another pitch shift would break the waveform.

Also, this feature is mostly meant to test / debug the following:

* Buffer use (detect underflows / overflows)
* QEMU audio output
* Timing
* EP emulation
* Clipping
* Routing
    * VP &rarr; GP
    * GP &rarr; Scratch memory
    * Scratch memory &rarr; EP
    * EP &rarr; ACI
    * Debug output

We don't need much control over the frequency to do these things - it should just be a recognizable shape, audible, and compress well.

Once we have confirmed that VP and routing works, we can also inject samples into VP via xboxpy with much greater control. That would then be used to test things like filters and envelopes.

hw/xbox/mcpx_apu.c
JayFoxRox: I had to create a new macro in `mcpx_apu.c` as the `dsp/dsp_cpu.h` is for internal DSP use - the `dsp/dsp.h` interface is also too generic at the moment; so it made sense to move the GP specific code to `mcpx_apu.c`.

Eventually we'll have to rework the `dsp.h` interface to support different memory sizes and memory banks - also see my previous post and the linked issue.

I chose a slightly different macro name than the existing one, because we typically refer to MIXBUFFER as MIXBUF. The name is also getting long which is bad for our 80 column limit.
The one in the DSP code will probably be renamed or even removed in the future (depends on if we consider MIXBUF a different bank, or just part of X memory), because it seems, that the DSP shouldn't even have to know about the existence of the MIXBUF.

hw/xbox/mcpx_apu.c
```diff
@@ -826,12 +849,38 @@ static void se_frame(void *opaque)
...
+             * or waves cancel out */
+            float offset = mixbin / (float)NUM_SAMPLES_PER_FRAME -
```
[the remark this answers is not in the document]
JayFoxRox: Should be fixed.

PR #194, opened 2019-02-05
nv2a: Fix A8Y8 texture formats

### LINKED ISSUE #36: Halo:CE terrain looks blueish / foggy in Battle Creek map
[This is a video showing the issue](https://www.youtube.com/watch?v=gm_DY6orkH4&t=2m8s).

*There is a similar issue on other maps. The best example is probably ["The Silent Cartographer" where the entire map is covered in fog](https://i.imgur.com/FVF3Qt6.png).*
*It is unknown if this is a related issue, so I've dedicated this issue to Battle Creek.*

---

**Background for debugging this issue:**

Fog is a dedicated feature of the fixed function pipeline (ffp), and in shaders it can be implemented in various ways, but the final combiner stage is ideal for it.

Considering that fog is typically the last stage, this issue looks like it's fog related (but could also be shadows or something else).
The later stages (such as fog) have the power to overwrite anything else - whereas textures typically happen very early on and are less uniform; lighting is also usually not *that* uniform either and it can't really overwrite the color in that way.

How you'd debug this depends on wether those draw calls are ffp or programmable pipeline:

- if this is programmable, I'd first make sure that if I just use "output = textures" that I get something useful (which should be an early step in the shader). then I'd remove commands at the end (where you'd expect fog) and see which one overwrites it again, until you've pinned the instruction which is at fault.
- if it's fixed function, then I'd first look into fog configuration and turn it off and just see what happens.

---

*(The following was noticed after writing the above fog description)*

I also noticed that the [base contains a foggy section](http://halo.wikia.com/wiki/File:Bcreek_bluebasesub.jpg). I wonder if this might be an effect which accidentally spreads over the entire map in XQEMU.
Note that the XQEMU wall does not appear to show any difference between the lower (fog) and upper  (no fog) portion.

Here's a screenshot from the linked video:

Battle Creek Issue Screenshot

### LINKED ISSUE #119: Halo: Combat Evolved weapon scope does not work.

JayFoxRox:
Not confirmed on hardware (and I don't intend to do so), but based on very strong assumptions and observed behavior.

Please test with a couple of games to confirm I didn't break anything.

This closes #119 and closes #36 for Battle Creek and Cartographer. It also solves Halo issues with the HUD.

Halo HUD and volumetric fog

*(This branch was rebased onto https://github.com/mborgerson/xqemu/tree/dev/halo for the screenshot)*

### DIFF
--- a/hw/xbox/nv2a/nv2a_pgraph.c
+++ b/hw/xbox/nv2a/nv2a_pgraph.c
@@ -192,7 +192,7 @@ static const ColorFormatInfo kelvin_color_format_map[66] = {
          {GL_ONE, GL_ONE, GL_ONE, GL_RED}},
     [NV097_SET_TEXTURE_FORMAT_COLOR_SZ_A8Y8] =
         {2, false, GL_RG8, GL_RG, GL_UNSIGNED_BYTE,
-         {GL_GREEN, GL_GREEN, GL_GREEN, GL_RED}},
+         {GL_RED, GL_RED, GL_RED, GL_GREEN}},
     [NV097_SET_TEXTURE_FORMAT_COLOR_LU_IMAGE_AY8] =
         {1, true, GL_R8, GL_RED, GL_UNSIGNED_BYTE,
          {GL_RED, GL_RED, GL_RED, GL_RED}},
@@ -207,7 +207,7 @@ static const ColorFormatInfo kelvin_color_format_map[66] = {
          {GL_ONE, GL_ONE, GL_ONE, GL_RED}},
     [NV097_SET_TEXTURE_FORMAT_COLOR_LU_IMAGE_A8Y8] =
         {2, true, GL_RG8, GL_RG, GL_UNSIGNED_BYTE,
-         {GL_GREEN, GL_GREEN, GL_GREEN, GL_RED}},
+         {GL_RED, GL_RED, GL_RED, GL_GREEN}},
 
     [NV097_SET_TEXTURE_FORMAT_COLOR_SZ_R6G5B5] =
         {2, false, GL_RGB8_SNORM, GL_RGB, GL_BYTE}, /* FIXME: This might be signed */


hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -207,7 +207,7 @@ static const ColorFormatInfo kelvin_color_format_map[66] = {
          {GL_ONE, GL_ONE, GL_ONE, GL_RED}},
     [NV097_SET_TEXTURE_FORMAT_COLOR_LU_IMAGE_A8Y8] =
         {2, true, GL_RG8, GL_RG, GL_UNSIGNED_BYTE,
-         {GL_GREEN, GL_GREEN, GL_GREEN, GL_RED}},
+         {GL_RED, GL_RED, GL_RED, GL_GREEN}},
```
mborgerson: This component swizzling stuff can be a little confusing when reading, especially if you don't know the byte order of the texture data. To other reviewers, when using the texture format associated with `NV097_SET_TEXTURE_FORMAT_COLOR_SZ_A8Y8`, each pixel in the texture has `Y8` as the first byte (address-wise, that is) of the pixel, then `A8` as the second byte of the pixel.

We use the GL format `GL_RG` and GL internal format `GL_RG8` format here with type `GL_UNSIGNED_BYTE`. This maps the first byte of a texel to the red component and the second byte to the green component. At texture bind time we then apply texture component swizzling such that when the fragment shader reads a sampled texel's `.r`, `.g`, or `.b` component, it will get the texel's "red" component (which in the texture data provided by the title is assumed to be the `Y8` channel), and accordingly when the shader reads a sampled texel's `.a` component, it will get the value from the "green" component of the texel (the `A8` channel in provided texture data).

mborgerson:
This looks correct to me, so I'm going to merge it now. Thanks for looking into this!

PR #196, opened 2019-02-06
nv2a: Use linear addressing for LU_IMAGE_A4R4G4B4

mborgerson:
I've not tested against hardware (don't intend to do so), but strong evidence and naming scheme (`_SZ` or not) for other textures formats suggests that the arrangement of pixels in this texture format is not swizzled and hence the `linear` field should be set to `true`. 

Fixing this issue resolves the problem with Halo 2 text not rendering correctly:
image

Note: This was done on top of mborgerson:perf-wip branch, which disables some `assert`s that prevent Halo 2 from launching on master.


JayFoxRox:
Thanks! Old code was obviously broken.

Issue #59, opened 2018-06-28
Factor out common XID stuff


mborgerson:
hw/xbox/xid.c and xid-sdl.c have a lot in common. Let's factor out the common bits to make things more maintainable.

PR #197, opened 2019-02-09
xid-sdl: Enable left/right force feedback

### LINKED ISSUE #71: Test, fix and enable `FORCE_FEEDBACK` in xid-sdl.c
Defining the `FORCE_FEEDBACK` preprocessor token should enable haptic feedback with XID-SDL.

---

I wrote xid-sdl.c when XQEMU was still using QEMU 1.x.
As QEMU 1.x had no SDL 2.x the haptic API was not ready either.

The force feedback code (to my knowledge) is largely untested, is slightly unclean and might not work.

Someone should address any compilation warnings / errors and enable force-feedback by default.

---

xid-sdl.c code has been rewritten to use the gamecontroller API since my original implementation, so you probably need to call [SDL_GameControllerGetJoystick](https://wiki.libsdl.org/SDL_GameControllerGetJoystick) to turn the SDL gamepad into an SDL joystick.

There might also be a seperate unrelated issue where force feedback doesn't work in some games (due to a problem in our XID emulation). So you should check if the game you are testing with even sets rumble strength to non-zero values.

DiscoStarslayer:
Minor bugfixes and changes to support left/right force feedback on compatible controllers.

Closes #71

### DIFF
--- a/hw/xbox/xid-sdl.c
+++ b/hw/xbox/xid-sdl.c
@@ -28,7 +28,6 @@
 
 #include <SDL2/SDL.h>
 
-//#define FORCE_FEEDBACK
 #define UPDATE
 
 //#define DEBUG_XID
@@ -59,6 +58,11 @@ enum {
     STR_PRODUCT,
     STR_SERIALNUMBER,
 };
+
+typedef enum HapticEmulationMode {
+    EMU_NONE,
+    EMU_HAPTIC_LEFT_RIGHT
+} HapticEmulationMode;
 
 static const USBDescStrings desc_strings = {
     [STR_MANUFACTURER]     = "QEMU",
@@ -105,11 +109,11 @@ typedef struct USBXIDState {
     XIDGamepadOutputReport out_state_capabilities;
 
     uint8_t device_index;
+
     SDL_GameController *sdl_gamepad;
-#ifdef FORCE_FEEDBACK
     SDL_Haptic *sdl_haptic;
     int sdl_haptic_effect_id;
-#endif
+    HapticEmulationMode haptic_emulation_mode;
 } USBXIDState;
 
 static const USBDescIface desc_iface_xbox_gamepad = {
@@ -194,36 +198,49 @@ static const XIDDesc desc_xid_xbox_gamepad = {
 
 #define BUTTON_MASK(button) (1 << ((button) - GAMEPAD_DPAD_UP))
 
-static void update_output(USBXIDState *s) {
-#ifdef FORCE_FEEDBACK
-    if (s->sdl_haptic == NULL) { return; }
-
-    SDL_HapticLeftRight effect = {
-        .type = SDL_HAPTIC_LEFTRIGHT,
-        .length = SDL_HAPTIC_INFINITY,
-        /* FIXME: Might be left/right inverted */
-        .large_magnitude = s->out_state.right_actuator_strength,
-        .small_magnitude = s->out_state.left_actuator_strength
-    };
-
-    if (s->sdl_haptic_effect_id == -1) {
-        int effect_id = SDL_HapticNewEffect(s->sdl_haptic,
-                                            (SDL_HapticEffect*)&effect);
-        if (effect_id == -1) {
-            fprintf(stderr, "Failed to upload haptic effect!\n");
-            SDL_HapticClose(s->sdl_haptic);
-            s->sdl_haptic = NULL;
-            return;
-        }
-        SDL_HapticRunEffect(s->sdl_haptic, effect_id, 1);
-
-        s->sdl_haptic_effect_id = effect_id;
-
-    } else {
-        SDL_HapticUpdateEffect(s->sdl_haptic, s->sdl_haptic_effect_id,
-                               (SDL_HapticEffect*)&effect);
+static int haptic_effect_prepare(SDL_HapticEffect *effect,
+                                 HapticEmulationMode mode,
+                                 uint16_t large_magnitude,
+                                 uint16_t small_magnitude)
+{
+    switch (mode) {
+    case EMU_HAPTIC_LEFT_RIGHT:
+        effect->leftright = (SDL_HapticLeftRight){
+            .type = SDL_HAPTIC_LEFTRIGHT,
+            .length = SDL_HAPTIC_INFINITY,
+            /* Xbox rumble returns in a range of 0x0000 -> 0xFFFF,
+             * SDL requires the range to be within 0x0000 -> 0x7FFF
+             */
+            .large_magnitude = large_magnitude >> 1,
+            .small_magnitude = small_magnitude >> 1,
+        };
+        break;
+    default:
+        fprintf(stderr, "Haptic Emulation Mode Not Supported!\n");
+        assert(false);
     }
-#endif
+
+    return 0;
+}
+
+static void update_output(USBXIDState *s)
+{
+    if (s->sdl_haptic == NULL) {
+        return;
+    }
+
+    SDL_HapticEffect effect;
+    int result = haptic_effect_prepare(&effect,
+                                       s->haptic_emulation_mode,
+                                       s->out_state.right_actuator_strength,
+                                       s->out_state.left_actuator_strength);
+
+    if (result < 0) {
+        fprintf(stderr, "Updating haptic effect failed!\n");
+        return;
+    }
+
+    SDL_HapticUpdateEffect(s->sdl_haptic, s->sdl_haptic_effect_id, &effect);
 }
 
 static void update_input(USBXIDState *s)
@@ -440,11 +457,9 @@ static void usb_xid_handle_destroy(USBDevice *dev)
 {
     USBXIDState *s = DO_UPCAST(USBXIDState, dev, dev);
     DPRINTF("xid handle_destroy\n");
-#ifdef FORCE_FEEDBACK
     if (s->sdl_haptic) {
         SDL_HapticClose(s->sdl_haptic);
     }
-#endif
     SDL_JoystickClose(s->sdl_gamepad);
 }
 #endif
@@ -462,6 +477,66 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
     uc->handle_data    = usb_xid_handle_data;
     // uc->handle_destroy = usb_xid_handle_destroy;
     uc->handle_attach  = usb_desc_attach;
+}
+
+static HapticEmulationMode select_haptic_emulation_mode(SDL_Haptic *haptic)
+{
+    unsigned int query_result = SDL_HapticQuery(haptic);
+    if (query_result & SDL_HAPTIC_LEFTRIGHT) {
+        return EMU_HAPTIC_LEFT_RIGHT;
+    }
+
+    fprintf(stderr, "No supported haptic mode found (SDL_HapticQuery result: 0x%x). Haptic disabled\n", query_result);
+
+    return EMU_NONE;
+}
+
+static int haptic_init(USBXIDState *s)
+{
+    if (SDL_HapticOpened(s->device_index)) {
+        fprintf(stderr, "Joystick haptic effect already opened, haptic effect disabled\n");
+        return -1;
+    }
+
+    SDL_Joystick *joystick = SDL_GameControllerGetJoystick(s->sdl_gamepad);
+    s->sdl_haptic = SDL_HapticOpenFromJoystick(joystick);
+    if (s->sdl_haptic == NULL) {
+        fprintf(stderr, "Joystick doesn't support haptic effects\n");
+        return -1;
+    }
+
+    s->haptic_emulation_mode = select_haptic_emulation_mode(s->sdl_haptic);
+
+    if (s->haptic_emulation_mode == EMU_NONE) {
+        SDL_HapticClose(s->sdl_haptic);
+        s->sdl_haptic = NULL;
+        fprintf(stderr, "Joystick doesn't support required haptic effects\n");
+        return -1;
+    }
+
+    SDL_HapticEffect effect;
+    haptic_effect_prepare(&effect, s->haptic_emulation_mode, 0, 0);
+
+    int effect_id = SDL_HapticNewEffect(s->sdl_haptic, &effect);
+    if (effect_id < 0) {
+        const char *error = SDL_GetError();
+        fprintf(stderr, "Failed to create haptic effect! SDL Error: %s\n", error);
+        SDL_HapticClose(s->sdl_haptic);
+        s->sdl_haptic = NULL;
+        return -1;
+    }
+
+    int run_result = SDL_HapticRunEffect(s->sdl_haptic, effect_id, 1);
+    if (run_result < 0) {
+        const char *error = SDL_GetError();
+        fprintf(stderr, "Failed to run haptic effect! SDL Error: %s\n", error);
+        SDL_HapticClose(s->sdl_haptic);
+        s->sdl_haptic = NULL;
+        return -1;
+    }
+
+    s->sdl_haptic_effect_id = effect_id;
+    return 0;
 }
 
 static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
@@ -500,7 +575,7 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
         exit(1);
     }
 
-    const char* name = SDL_GameControllerName(s->sdl_gamepad);
+    const char *name = SDL_GameControllerName(s->sdl_gamepad);
     printf("Found game controller %d (%s)\n", s->device_index, name);
 
 #ifndef UPDATE
@@ -511,18 +586,11 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
     SDL_JoystickEventState(SDL_ENABLE);
 #endif
 
-#ifdef FORCE_FEEDBACK
-    s->sdl_haptic = SDL_HapticOpenFromJoystick(s->sdl_gamepad);
-    if (s->sdl_haptic == NULL) {
-        fprintf(stderr, "Joystick doesn't support haptic\n");
+    if (!SDL_InitSubSystem(SDL_INIT_HAPTIC)) {
+        haptic_init(s);
     } else {
-        if ((SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
-            fprintf(stderr, "Joystick doesn't support necessary haptic effect\n");
-            SDL_HapticClose(s->sdl_haptic);
-            s->sdl_haptic = NULL;
-        }
+        fprintf(stderr, "SDL failed to initialize haptic feedback subsystem\n");
     }
-#endif
 }
 
 static Property xid_sdl_properties[] = {


hw/xbox/xid-sdl.c
```diff
@@ -194,21 +194,22 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
-        .length = SDL_HAPTIC_INFINITY,
-        /* FIXME: Might be left/right inverted */
-        .large_magnitude = s->out_state.right_actuator_strength,
-        .small_magnitude = s->out_state.left_actuator_strength
-    };
+    if (s->sdl_haptic == NULL) {
+        return;
+    }
+
+    SDL_memset( &effect, 0, sizeof(SDL_HapticEffect) ); // 0 is safe default
```
DiscoStarslayer: For whatever reason, passing in `SDL_HapticLeftRight` to `SDL_HapticNewEffect` failed, setting `effect_id` to `-1`. Using `SDL_HapticEffect` with `leftright` params seems to work fine instead.

We probably don't need to allocate it this way though. I was comparing the control flow to [this implementation.](https://github.com/JayFoxRox/sdl2-jstest-haptic) I was being too impatient testing the rumble within xqemu with the game I was testing with. It took a good 20-30 seconds after the game booted for `SDL_HapticRunEffect` to rumble my controller even with hard coded `large_magnitude`. The solution here isn't necessary but `sdl2-jstest-haptic` worked so I wanted to get as close to that as possible.

hw/xbox/xid-sdl.c
```diff
@@ -194,21 +194,22 @@ static const XIDDesc desc_xid_xbox_gamepad = {
 
 #define BUTTON_MASK(button) (1 << ((button) - GAMEPAD_DPAD_UP))
 
+SDL_HapticEffect effect;
```
DiscoStarslayer: This is a hacky hacky bad hack. Shared memory across all plugged in devices. Should probably be within `USBXIDState` so each controller maintains it's own force feedback event.

hw/xbox/xid-sdl.c
```diff
@@ -488,7 +488,7 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
     s->out_state_capabilities.report_id = 0;
 
     /* FIXME: Make sure SDL was init before */
-    if (SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER)) {
+    if (SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER | SDL_INIT_HAPTIC)) {
```
DiscoStarslayer: This took me an embarrassingly long time to catch. While I'm here maybe I should touch up the `FIXME` above it.

hw/xbox/xid-sdl.c
```diff
@@ -512,14 +512,16 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
 #endif
 
 #ifdef FORCE_FEEDBACK
-    s->sdl_haptic = SDL_HapticOpenFromJoystick(s->sdl_gamepad);
+    s->sdl_haptic = SDL_HapticOpen(s->device_index);
```
DiscoStarslayer: `SDL_HapticOpenFromJoystick` segfaults. Not sure why, SDL2 docs suggest you can use a gamepad like a joystick, gamepad just has convenient button mappings.

This works for now, but reminder for myself to try and get `s->sdl_gamepad` to work as I feel that is a slightly cleaner impl, maybe needs a cast?

hw/xbox/xid-sdl.c
```diff
@@ -512,14 +512,16 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
...
     if (s->sdl_haptic == NULL) {
         fprintf(stderr, "Joystick doesn't support haptic\n");
     } else {
         if ((SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
             fprintf(stderr, "Joystick doesn't support necessary haptic effect\n");
             SDL_HapticClose(s->sdl_haptic);
             s->sdl_haptic = NULL;
+        } else {
+            s->sdl_haptic_effect_id = -1;
```
DiscoStarslayer: `sdl_haptic_effect_id` is never assigned a value, so you played poker with your existing ram to see if we even enable haptic feedback. Assigning value here if device supports `SDL_HAPTIC_LEFTRIGHT`

hw/xbox/xid-sdl.c
JayFoxRox: The FIXME is still valid for non SDL UIs - we'll have to review if QEMU does the SDL init / if the SDL1 / SDL2 behaviour changed.

I believe SDL2 `SDL_InitSubSystem` is more powerful in that it doesn't need another init before.

hw/xbox/xid-sdl.c
JayFoxRox: A GameController internally uses a joystick, which you have to retrieve with [SDL_GameControllerGetJoystick](https://wiki.libsdl.org/SDL_GameControllerGetJoystick).

---

Ideally we'd do it the other way around: use a more powerful Joystick, and then only use GameController API to find a good default mapping - this was my original design but it was unfortunately changed when my code was ported to QEMU 2.x.
This redesign is obviously out of scope for this PR.

hw/xbox/xid-sdl.c
JayFoxRox: I don't think this is a good spot to set it though (rather hidden). I think it should happen independent of the haptic creation; somewhere where it's easily seen.

We should also consider creating the effect right here, so we have the feature query and effect creation together. The old method might cause a delay or lag when the effect is created during runtime, which is not ideal.

hw/xbox/xid-sdl.c
```diff
@@ -194,21 +194,22 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
+    effect.type = SDL_HAPTIC_LEFTRIGHT;
+    effect.leftright.length = SDL_HAPTIC_INFINITY;
+    effect.leftright.large_magnitude = s->out_state.right_actuator_strength; // x/32767 strength
+    effect.leftright.small_magnitude = s->out_state.left_actuator_strength; // x/32767 strengt
 
     if (s->sdl_haptic_effect_id == -1) {
-        int effect_id = SDL_HapticNewEffect(s->sdl_haptic,
-                                            (SDL_HapticEffect*)&effect);
+        int effect_id = SDL_HapticNewEffect(s->sdl_haptic, &effect);
         if (effect_id == -1) {
```
JayFoxRox: This scope is confusing, I don't know where it ends - indentation bad?

hw/xbox/xid-sdl.c
JayFoxRox: Please check if we can make it work with the smaller struct, or how this is intended in SDL (by checking their examples or contacting the API author [should be @bobbens])

hw/xbox/xid-sdl.c
JayFoxRox: Also, if we keep it, it should be guarded by `#ifdef FORCE_FEEDBACK `. But we might want to get rid of the macro anyway.

hw/xbox/xid-sdl.c
```diff
@@ -194,21 +194,22 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
-        .small_magnitude = s->out_state.left_actuator_strength
-    };
+    if (s->sdl_haptic == NULL) {
+        return;
+    }
+
+    SDL_memset( &effect, 0, sizeof(SDL_HapticEffect) ); // 0 is safe default
+    effect.type = SDL_HAPTIC_LEFTRIGHT;
+    effect.leftright.length = SDL_HAPTIC_INFINITY;
+    effect.leftright.large_magnitude = s->out_state.right_actuator_strength; // x/32767 strength
```
JayFoxRox: The comment is pretty bad. I'm also not sure about the range of `s->out_state.right_actuator_strength` etc.

We should definitely have a comment describing the Xbox and SDL range, but the current comment is confusing.

hw/xbox/xid-sdl.c
DiscoStarslayer: Agreed on both counts. Since the effect is permanent it might just make more sense to store the effect on `s`'s struct instead of inferring it with the `id` anyway.

hw/xbox/xid-sdl.c
JayFoxRox: The haptic subsystem might be optional (we'll have to check the SDL docs), so we should probably initialize that separately to get meaningful error messages + disable haptic support if necessary.

Before changing anything, consult the SDL docs.

hw/xbox/xid-sdl.c
```diff
@@ -512,14 +512,16 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
 #endif
 
 #ifdef FORCE_FEEDBACK
-    s->sdl_haptic = SDL_HapticOpenFromJoystick(s->sdl_gamepad);
+    s->sdl_haptic = SDL_HapticOpen(s->device_index);
     if (s->sdl_haptic == NULL) {
         fprintf(stderr, "Joystick doesn't support haptic\n");
     } else {
         if ((SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
             fprintf(stderr, "Joystick doesn't support necessary haptic effect\n");
```
JayFoxRox: Error message should be more specific, so users can debug this issue + so we don't have to rename it when adding additional checks `[...] effect (SDL_HAPTIC_LEFTRIGHT)` or something

hw/xbox/xid-sdl.c
JayFoxRox: `SDL_HapticEffect` should remain on the stack, because it's a stateless design: we just write to it and never read. We never have to do partial updates. It's a small structure, too.

The `id` should remain part of the state (we have to read it often), but be initialized once.

hw/xbox/xid-sdl.c
DiscoStarslayer: Without `SDL_INIT_HAPTIC`, `SDL_HapticOpen` always returns -1 and `SDL_GetError` reports that the device does not support haptic feedback, regardless of feedback support. Without `SDL_HapticOpen` I can't even sniff for haptic support with `SDL_HapticQuery`, so I think it is required here.

Docs are fairly thin on this matter: https://wiki.libsdl.org/CategoryForceFeedback

DiscoStarslayer: To be clear, I am for a separate init to handle the error gracefully though.

JayFoxRox: Right - you still have to init it obviously. But it should probably be done like this:

```c
if (SDL_InitSubSystem(SDL_INIT_GAMECONTROLLER)) { /* handle errors for gamecontroller [abort?] */ }
if (SDL_InitSubSystem(SDL_INIT_HAPTIC)) { /* handle errors for haptic [disable / skip force-feedback] */ }
```

Because only that way we can cleanly differentiate between errors of the *required* gamecontroller subsystem, and the *optional* haptic subsystem.

*(Or whatever SDL recommens / what other projects do that ran into similar issues)*

DiscoStarslayer: Makes sense, I miss-understood what you meant by optional. I think we are both in agreement here. Good idea.

hw/xbox/xid-sdl.c
```diff
@@ -106,10 +105,9 @@ typedef struct USBXIDState {
 
     uint8_t device_index;
     SDL_GameController *sdl_gamepad;
```
DiscoStarslayer: There is a lot of properties to this `struct` that are prefixed with `sdl`. Maybe I should add a new struct like `SDLDeviceState`(name not final) and have that as a property of `USBXIDState`?

hw/xbox/xid-sdl.c
```diff
@@ -195,35 +193,14 @@ static const XIDDesc desc_xid_xbox_gamepad = {
 #define BUTTON_MASK(button) (1 << ((button) - GAMEPAD_DPAD_UP))
 
 static void update_output(USBXIDState *s) {
-#ifdef FORCE_FEEDBACK
     if (s->sdl_haptic == NULL) { return; }
+    if (s->sdl_haptic_effect == NULL) { return; }
```
DiscoStarslayer: Maybe overkill?

hw/xbox/xid-sdl.c
```diff
@@ -195,35 +193,14 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
-        if (effect_id == -1) {
-            fprintf(stderr, "Failed to upload haptic effect!\n");
-            SDL_HapticClose(s->sdl_haptic);
-            s->sdl_haptic = NULL;
-            return;
-        }
-        SDL_HapticRunEffect(s->sdl_haptic, effect_id, 1);
-
-        s->sdl_haptic_effect_id = effect_id;
+    s->sdl_haptic_effect->type = SDL_HAPTIC_LEFTRIGHT;
```
DiscoStarslayer: I have to re-set the type each time we update the controller. Something within SDL is changing this value to `NULL`. Haven't narrowed it down yet, but SDL manipulating this memory feels wrong, so maybe this isn't the recommended way of using the API?

I will keep digging on this.

hw/xbox/xid-sdl.c
```diff
@@ -511,18 +486,49 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
...
+                };
+
+                int effect_id = SDL_HapticNewEffect(s->sdl_haptic, &effect);
+                if (effect_id == -1) {
+                    const char* error = SDL_GetError();
+                    fprintf(stderr, "Failed to create haptic effect! SDL Error: %s", error);
+                    SDL_HapticClose(s->sdl_haptic);
+                    s->sdl_haptic = NULL;
+                    s->sdl_haptic_effect = NULL;
+                    return;
```
DiscoStarslayer: return here is a crutch. It's convenient because this is the last branch in the method, but if more logic is added later this may cause tricky bugs. I will rework.

hw/xbox/xid-sdl.c
```diff
@@ -511,18 +486,49 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
...
-        if ((SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
-            fprintf(stderr, "Joystick doesn't support necessary haptic effect\n");
-            SDL_HapticClose(s->sdl_haptic);
-            s->sdl_haptic = NULL;
+    if (!SDL_HapticOpened(s->device_index)) {
+        s->sdl_haptic = SDL_HapticOpen(s->device_index);
+        if (s->sdl_haptic == NULL) {
+            fprintf(stderr, "Joystick doesn't support haptic effects\n");
+        } else {
+            if (SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) {
```
DiscoStarslayer: flipped the `if` statement to allow adding `else if` later for other fallback haptic modes in future commit.

hw/xbox/xid-sdl.c
```diff
@@ -511,18 +486,49 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
...
+                    const char* error = SDL_GetError();
+                    fprintf(stderr, "Failed to run haptic effect! SDL Error: %s\n", error);
+                    SDL_HapticClose(s->sdl_haptic);
+                    s->sdl_haptic = NULL;
+                    s->sdl_haptic_effect = NULL;
+                    return;
+                }
+
+                s->sdl_haptic_effect_id = effect_id;
+                s->sdl_haptic_effect = &effect;
```
JayFoxRox: `effect` is on the stack. This pointer is invalid once you leave the function.

hw/xbox/xid-sdl.c
JayFoxRox: See https://github.com/xqemu/xqemu/pull/197/files#r255324191

I still think you can just have a new local variable here. If you feel like duplicating the init of 2 fields is too much, you could just move it into a setter for the SDL structure like `SetLeftRightHapticEffect(&effect, left_strength, right_strength)` (maybe SDL already has macros / functions for this?)

hw/xbox/xid-sdl.c
JayFoxRox: Just group them visually for now. We'll probably worry about this when doing #59

hw/xbox/xid-sdl.c
JayFoxRox: Such things typically indicate you should be using functions

JayFoxRox:
Use an existing prefix for the commit title, `ff` isn't exactly descriptive when looking at `git log`; I assume this will be `xid-sdl`

hw/xbox/xid-sdl.c
JayFoxRox: Noted. Not worth mentioning as the intention is clear.

*Generally, unless there's an issue, don't bother adding a GitHub comment (to avoid unnecessary discussions or many "pings" to maintainers). If something is unclear, the reviewers will leave a comment anyway. You can then state the intention (which is often asked to be moved into code comments, if it's hard to understand).*

hw/xbox/xid-sdl.c
```diff
@@ -195,35 +193,21 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
+    }
 
-        s->sdl_haptic_effect_id = effect_id;
 
-    } else {
-        SDL_HapticUpdateEffect(s->sdl_haptic, s->sdl_haptic_effect_id,
-                               (SDL_HapticEffect*)&effect);
-    }
-#endif
+    SDL_HapticUpdateEffect(s->sdl_haptic, s->sdl_haptic_effect_id, s->sdl_haptic_effect);
```
DiscoStarslayer: With my testing, I've found just having this in the `update_output` function causes serious performance regression.

Maybe there can be improvements with only updating the effect when the actuator value changes? Or maybe run this in a non-blocking manner?

DiscoStarslayer: False alarm on the regression. After further testing and capturing timing on these functions, turns out the performance drop was from other causes, and I just had bad luck when I turned force feedback off making me feel like I introduced a regression.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +446,81 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
     uc->handle_attach  = usb_desc_attach;
 }
 
+static int apply_haptic_effect(USBXIDState *s, SDL_HapticEffect *effect) {
```
JayFoxRox: `Apply` doesn't really show that this is more complicated than `Update`.

`initialize` or `create` are probably better names.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +446,81 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+
+    s->sdl_haptic = SDL_HapticOpen(s->device_index);
+    if (s->sdl_haptic == NULL) {
+        fprintf(stderr, "Joystick doesn't support haptic effects\n");
+        return -1;
+    }
+
+    SDL_HapticEffect *effect = g_malloc(sizeof(SDL_HapticEffect));
+    int result = haptic_effect_init(s, effect);
+    if (result == -1) { g_free(effect); }
```
JayFoxRox: `haptic_effect_init` re-initializes the effect struct  from scratch anyway.

There's no reason to `malloc` it. Just use a local variable here (unless SDL requires it to remain in memory - which is unlikely).

hw/xbox/xid-sdl.c
```diff
@@ -195,35 +193,21 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
-        if (effect_id == -1) {
-            fprintf(stderr, "Failed to upload haptic effect!\n");
-            SDL_HapticClose(s->sdl_haptic);
-            s->sdl_haptic = NULL;
-            return;
-        }
-        SDL_HapticRunEffect(s->sdl_haptic, effect_id, 1);
+    if (s->sdl_haptic_effect->type == SDL_HAPTIC_SINE) {
+        s->sdl_haptic_effect->periodic.magnitude =
+            (s->out_state.right_actuator_strength + s->out_state.left_actuator_strength) >> 1;
```
JayFoxRox: `/ 2` (the generated code will likely be the same, but shows intent)

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +446,81 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+
+        effect->periodic = (SDL_HapticPeriodic){
+            .type = SDL_HAPTIC_SINE,
+            .period = 1000,
+            .magnitude = 0,
+            .length = SDL_HAPTIC_INFINITY,
+            .attack_length = 0,
+            .fade_length = 0,
+            .delay = 0,
+        };
```
JayFoxRox: If we are using sine, the frequency (`period`?) should probably be influenced by the left/right strength because those motors *do* run different frequencies.

Due to the added complexity, I'd probably avoid adding this fallback for now (split it into a second commit possibly, so we can apply it later if necessary). I'd assume that almost every device will support Left/Right anyway.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +446,81 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+        s->sdl_haptic_effect = NULL;
+        return -1;
+    }
+
+    s->sdl_haptic_effect = effect;
+    s->sdl_haptic_effect_id = effect_id;
+    return 0;
+}
+
+static int haptic_effect_init(USBXIDState *s, SDL_HapticEffect *effect) {
```
JayFoxRox: If this function took the `strength` / `magnitude` arguments it would be a lot more powerful and reusable.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +446,81 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+    if (SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) {
+        effect->leftright = (SDL_HapticLeftRight){
+            .type = SDL_HAPTIC_LEFTRIGHT,
+            .length = SDL_HAPTIC_INFINITY,
+            .large_magnitude = 0x0000,
+            .small_magnitude = 0x0000,
+        };
+
+        return apply_haptic_effect(s, effect);
```
JayFoxRox: As your current `apply_haptic_effect` creates a new effect, I'd advise to do `apply_haptic_effect` in the caller (which knows `s` and `effect` anyway). It would make this function *much* more reusable as it could be used to construct an haptic effect for either `SDL_HapticCreateEffect` or `SDL_HapticUpdateEffect`.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +446,81 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+        return -1;
+    }
+
+    s->sdl_haptic_effect = effect;
+    s->sdl_haptic_effect_id = effect_id;
+    return 0;
+}
+
+static int haptic_effect_init(USBXIDState *s, SDL_HapticEffect *effect) {
+    if (SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) {
```
JayFoxRox: We should probably have some `EmulationMode` enum (bad name) which picks which method to use for our effects on startup. That way, we only have to do the query once.

We'd then store the chosen emulation mode in `s` to generate the proper effect structure.

This is more extensible and can also be used in the future to take a user-override for devices with broken force feedback.

hw/xbox/xid-sdl.c
DiscoStarslayer: My assumption was that sine rumble will never truly be fully "accurate" and is just an approximation for users with older controllers. The period here is taken from SDL's implementation of `SDL_HapticRumble`: https://github.com/spurious/SDL-mirror/blob/master/src/haptic/SDL_haptic.c#L773. Testing on my local controller feels fine with this mode.

I warn the user with a print to stderr when initializing the fallback, maybe I can make it more clear that the fallback implementation isn't considered accurate?

I guess that goes back to the xqemu's philosophy though, should we even have this implementation if it will never truly be accurate? I have no problem removing the code here if that is preferred.

hw/xbox/xid-sdl.c
DiscoStarslayer: I think with that method we can also avoid allocating on the heap. I like it :+1:

hw/xbox/xid-sdl.c
JayFoxRox: Of course it's an approximation, but imagine the following cases:

- 100% of the large motor, 0% small motor: ~ sine at amplitude 100 at a very specific period (low frequency, because large weight spins slow).
- 100% of the small motor, 0% large motor: ~ sine at amplitude 100 at a very specific period (high frequency, because small weight spins fast).

The current fallback would always use the same frequency - So even approximations can be good or bad.

Yes, we should still have it, *if it's necessary*. However, I'd assume that most devices will have Left/Right anyway for XInput, possibly using similar approximation as implied here to drive a different combination of motors (to be compatible with most games).

So while we should ensure our code can possibly handle it in the future, we shouldn't have to worry about it for now (it's not a problem yet, so why spend time solving it?).

If we encounter many broken devices or XInput changes / new force feeback standards arrive, then we might have to revisit this.
And if we do, we'll probably want it as accurate as possible (not necessarily because of philosophy, but simply because users will expect it).

DiscoStarslayer: All very fair points. Thanks for spending the time for the breakdown.

hw/xbox/xid-sdl.c
JayFoxRox: Actually SDL also keeps their stuff on the heap.. https://github.com/spurious/SDL-mirror/blob/master/src/haptic/SDL_haptic.c#L821

Just find a working solution which is easy to read (I still prefer the heapless variant)

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +195,42 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
-
-    SDL_HapticLeftRight effect = {
-        .type = SDL_HAPTIC_LEFTRIGHT,
-        .length = SDL_HAPTIC_INFINITY,
-        /* FIXME: Might be left/right inverted */
-        .large_magnitude = s->out_state.right_actuator_strength,
-        .small_magnitude = s->out_state.left_actuator_strength
-    };
+static SDL_HapticEffect haptic_effect_init(USBXIDState *s, float large_magnitude_percent, float small_magnitude_percent) {
+    if (large_magnitude_percent > 1) {
```
DiscoStarslayer: Should I even clamp these? and if so, should I clamp numbers below `0` too?

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +195,40 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
+        effect.leftright = (SDL_HapticLeftRight){
+            .type = SDL_HAPTIC_LEFTRIGHT,
+            .length = SDL_HAPTIC_INFINITY,
+            .large_magnitude = floor(large_magnitude_percent * SDL_MAX_HAPTIC_VALUE),
+            .small_magnitude = floor(small_magnitude_percent * SDL_MAX_HAPTIC_VALUE),
+        };
     } else {
-        SDL_HapticUpdateEffect(s->sdl_haptic, s->sdl_haptic_effect_id,
-                               (SDL_HapticEffect*)&effect);
+        fprintf(stderr, "Unknown haptic effect type\n");
```
DiscoStarslayer: This shouldn't happen, we don't even call this function unless we query an effect in `haptic_init`. Should I just not branch here? maybe an assert?

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +195,40 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
+
+    return effect;
+}
+
+static void update_output(USBXIDState *s) {
+    if (s->sdl_haptic == NULL) { return; }
+
+    SDL_HapticEffect effect = haptic_effect_init(
+        s,
+        s->out_state.right_actuator_strength / (float)XID_MAX_HAPTIC_VALUE,
```
DiscoStarslayer: Wasn't sure if having `haptic_effect_init` take a percent here is better, or just assuming `haptic_effect_init` will receive a value with the range of `0x0000 -> 0xFFFF` and doing the math in there. What do you think?

hw/xbox/xid-sdl.c
```diff
@@ -54,6 +53,9 @@
 #define TYPE_USB_XID_SDL "usb-xbox-gamepad-sdl"
 #define USB_XID(obj) OBJECT_CHECK(USBXIDState, (obj), TYPE_USB_XID_SDL)
 
+#define SDL_MAX_HAPTIC_VALUE 0x7FFF
+#define XID_MAX_HAPTIC_VALUE 0xFFFF
```
DiscoStarslayer: Confirmed we get a max of `0xFFFF` on both left + right actuator by firing a shotgun in halo

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +195,40 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
+}
+
+static void update_output(USBXIDState *s) {
+    if (s->sdl_haptic == NULL) { return; }
+
+    SDL_HapticEffect effect = haptic_effect_init(
+        s,
+        s->out_state.right_actuator_strength / (float)XID_MAX_HAPTIC_VALUE,
+        s->out_state.left_actuator_strength / (float)XID_MAX_HAPTIC_VALUE
+    );
```
JayFoxRox: nit: Why bother with floats?

These are obviously integers using their full range; so to go from XID to SDL values, we just have to do `actuator_strength >> 1`.

*(We do the same for input, but it might actually change there because input must be more flexible - for output I don't see this issue)*

hw/xbox/xid-sdl.c
JayFoxRox: If so, clamp in both directions, probably using MIN / MAX functions or something (I'm sure QEMU has something).

However, I wouldn't bother with it; maybe `assert((strength >= 0.0) && (strength <= 1.0))`?

hw/xbox/xid-sdl.c
JayFoxRox: - Report the error, but also `assert(false);`
- Also print the effect type so we can debug user reports
- Turn this into a `switch` / `case`

The nv2a should contain a lot of similar code where you can copy the style from.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +467,49 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
     uc->handle_attach  = usb_desc_attach;
 }
 
+
+static int haptic_init(USBXIDState *s) {
+    if (SDL_HapticOpened(s-> device_index)) { return 0; }
+
+    s->sdl_haptic = SDL_HapticOpen(s->device_index);
```
JayFoxRox: Use `OpenFromJoystick` please.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +467,49 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
     uc->handle_attach  = usb_desc_attach;
 }
 
+
+static int haptic_init(USBXIDState *s) {
+    if (SDL_HapticOpened(s-> device_index)) { return 0; }
```
JayFoxRox: Space between `s->` and `device_index`

JayFoxRox: What is this check supposed to do? Shouldn't it report an error message somehow?

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +467,49 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+    if (SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) {
+        s->sdl_haptic_effect_type = SDL_HAPTIC_LEFTRIGHT;
+    } else {
+        SDL_HapticClose(s->sdl_haptic);
+        s->sdl_haptic = NULL;
+        fprintf(stderr, "Joystick doesn't support left/right haptic feedback\n");
+        return -1;
+    }
+
+    SDL_HapticEffect effect = haptic_effect_init(s, 0, 0);
```
JayFoxRox: If you use floats, use floats: `0.0f`

hw/xbox/xid-sdl.c
```diff
@@ -106,10 +108,9 @@ typedef struct USBXIDState {
 
     uint8_t device_index;
     SDL_GameController *sdl_gamepad;
-#ifdef FORCE_FEEDBACK
     SDL_Haptic *sdl_haptic;
+    int sdl_haptic_effect_type;
```
JayFoxRox: I'd use our own enum for this.

hw/xbox/xid-sdl.c
```diff
@@ -511,18 +557,11 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
...
-        fprintf(stderr, "Joystick doesn't support haptic\n");
-    } else {
-        if ((SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
-            fprintf(stderr, "Joystick doesn't support necessary haptic effect\n");
-            SDL_HapticClose(s->sdl_haptic);
-            s->sdl_haptic = NULL;
-        }
+    if (SDL_InitSubSystem(SDL_INIT_HAPTIC)) {
+        fprintf(stderr, "SDL failed to initialize haptic feedback subsystem\n");
     }
```
JayFoxRox: This should probably be in the class-init, so it's only called once. Not for every instance.
The same is true for the other SDL init though.

We should probably create an issue to review these things and do a follow-up PR, and refer developers to https://people.redhat.com/~thuth/blog/qemu/2018/09/10/instance-init-realize.html

hw/xbox/xid-sdl.c
DiscoStarslayer: `haptic_init` is called in `usb_xbox_gamepad_realize`. In `usb_xbox_gamepad_realize` it is mentioned that poorly programmed games may make this call multiple times. The `SDL_HapticOpened` check short-circuits `haptic_init` in the case it is called repeatedly on an already initialized device.

I don't mind adding a fprintf here, or an assert if we want to see if any games actually initialize controllers multiple times.

Not sure how SDL handles multiple opened haptic devices with multiple haptic effects, but I'd rather not find out.

DiscoStarslayer: Thinking about it some more, maybe this check makes more sense to live in `usb_xbox_gamepad_realize` as that method has more context. Then I can keep the check in `haptic_init` but make it a printf + assert to enforce only calling it once.

hw/xbox/xid-sdl.c
DiscoStarslayer: Floats felt more generic to me, `actuator_strength >> 1` relies on xbox xid out_state specifics. But I see your point on output not needing to be flexible and I like the implementation much more without the floats, so I'll follow your advice here.

hw/xbox/xid-sdl.c
JayFoxRox: Wouldn't an already initialized device have `sdl_haptic` non-`NULL`? This check sounds more like it checks for *another* user (like another program blocking the device, or maybe a second gamepad with the same ID).

To be honest, the documentation for this function is very vague, so I might be wrong.

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +201,35 @@ static const XIDDesc desc_xid_xbox_gamepad = {
 
 #define BUTTON_MASK(button) (1 << ((button) - GAMEPAD_DPAD_UP))
 
-static void update_output(USBXIDState *s) {
-#ifdef FORCE_FEEDBACK
-    if (s->sdl_haptic == NULL) { return; }
+static SDL_HapticEffect haptic_effect_init(HapticEmulationMode mode, uint large_magnitude, uint small_magnitude) {
+    assert((large_magnitude <= XID_MAX_HAPTIC_VALUE) && (small_magnitude <= XID_MAX_HAPTIC_VALUE));
```
JayFoxRox: Just use `uint16_t` for `_magnitude`? that way, you don't even need the `assert` - it implies the full range is used.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +466,59 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
     uc->handle_attach  = usb_desc_attach;
 }
 
+static HapticEmulationMode detect_haptic_emulation_mode(SDL_Haptic *haptic) {
```
JayFoxRox: nit: This doesn't just detect, it *picks* a mode from the detected modes

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +466,59 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+    }
+
+    return EMU_UNSUPPORTED;
+}
+
+static int haptic_init(USBXIDState *s) {
+    if (SDL_HapticOpened(s->device_index)) { return 0; }
+
+    s->sdl_haptic = SDL_HapticOpenFromJoystick(
+        SDL_GameControllerGetJoystick(s->sdl_gamepad)
```
JayFoxRox: nit: call in call is not ideal. A temporary variable (`sdl_joystick` for example) would make this more readable.

hw/xbox/xid-sdl.c
DiscoStarslayer: From what I gathered, all this function checks for is if the `haptic_reference_count` is `> 0`. Every time `SDL_HapticOpen[FromJoystick]` is called on a device, `haptic_reference_count` is incremented. Conversely it is decremented when `SDL_HapticClose` is called. I would _think_ this is on a per-application basis.

JayFoxRox: >  In usb_xbox_gamepad_realize it is mentioned that poorly programmed games may make this call multiple times

This is wrong.

The comment is about something else (when to update the input, which happens later - so to prevent hammering the device we do a one-time-init to prevent it). Check the QOM documentation as to when `_realize` is called by QEMU (I've put a link in another PR comment during this review).

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +466,59 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+static HapticEmulationMode detect_haptic_emulation_mode(SDL_Haptic *haptic) {
+    if (SDL_HapticQuery(haptic) & SDL_HAPTIC_LEFTRIGHT) {
+        return EMU_HAPTIC_LEFT_RIGHT;
+    }
+
+    return EMU_UNSUPPORTED;
+}
+
+static int haptic_init(USBXIDState *s) {
+    if (SDL_HapticOpened(s->device_index)) { return 0; }
```
JayFoxRox: Given https://github.com/xqemu/xqemu/pull/197#discussion_r255591397, this still implies we should return an error along the lines of "Haptic device is already in use" (or some more user-friendly equivalent).

DiscoStarslayer: Good call, and thanks for the clarification on https://github.com/xqemu/xqemu/pull/197#discussion_r255591397, I clearly miss-understood what the comment was mentioning, and now I can read up on some QOM docs!

hw/xbox/xid-sdl.c
DiscoStarslayer: Good call, created ticket here: https://github.com/xqemu/xqemu/issues/210

hw/xbox/xid-sdl.c
```diff
@@ -511,18 +564,11 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
...
-            fprintf(stderr, "Joystick doesn't support necessary haptic effect\n");
-            SDL_HapticClose(s->sdl_haptic);
-            s->sdl_haptic = NULL;
-        }
+    if (SDL_InitSubSystem(SDL_INIT_HAPTIC)) {
+        fprintf(stderr, "SDL failed to initialize haptic feedback subsystem\n");
     }
-#endif
+
+    haptic_init(s);
```
mborgerson: If `SDL_InitSubSystem(SDL_INIT_HAPTIC)` fails, we still do `haptic_init` anyway?

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +462,61 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+        s->sdl_haptic = NULL;
+        fprintf(stderr, "Joystick doesn't support necissary haptic effects\n");
+        return -1;
+    }
+
+    SDL_HapticEffect effect;
+    haptic_effect_init(&effect, s->haptic_emulation_mode, 0, 0);
+
+    int effect_id = SDL_HapticNewEffect(s->sdl_haptic, &effect);
+    if (effect_id == -1) {
```
mborgerson: According to [the spec](https://wiki.libsdl.org/SDL_HapticNewEffect), `SDL_HapticNewEffect` *"Returns the ID of the effect on success or a negative error code on failure."* This condition should be changed to `if (effect_id < 0)`.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +462,61 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+        const char* error = SDL_GetError();
+        fprintf(stderr, "Failed to create haptic effect! SDL Error: %s\n", error);
+        SDL_HapticClose(s->sdl_haptic);
+        s->sdl_haptic = NULL;
+        return -1;
+    }
+
+    int run_result = SDL_HapticRunEffect(s->sdl_haptic, effect_id, 1);
+    if (run_result == -1) {
```
mborgerson: Same comment as above w.r.t. negative error codes.

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +199,31 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
-static void update_output(USBXIDState *s) {
-#ifdef FORCE_FEEDBACK
-    if (s->sdl_haptic == NULL) { return; }
+static int haptic_effect_init(SDL_HapticEffect *effect, HapticEmulationMode mode, uint16_t large_magnitude, uint16_t small_magnitude) {
+    switch (mode) {
+    case EMU_HAPTIC_LEFT_RIGHT:
+        effect->leftright = (SDL_HapticLeftRight){
+            .type = SDL_HAPTIC_LEFTRIGHT,
+            .length = SDL_HAPTIC_INFINITY,
+            .large_magnitude = large_magnitude >> 1,
```
mborgerson: Probably worth mentioning somewhere in a comment that the range Xbox magnitude is in range [0-`0xffff`] (according to your research) while SDL apparently has a [max value of `0x7fff`](https://wiki.libsdl.org/SDL_HapticEffect#Remarks), hence the shift.

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +199,31 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
-            s->sdl_haptic = NULL;
-            return;
-        }
-        SDL_HapticRunEffect(s->sdl_haptic, effect_id, 1);
+static void update_output(USBXIDState *s) {
+    if (s->sdl_haptic == NULL) { return; }
 
-        s->sdl_haptic_effect_id = effect_id;
+    SDL_HapticEffect effect;
+    haptic_effect_init(&effect, s->haptic_emulation_mode, s->out_state.right_actuator_strength, s->out_state.left_actuator_strength);
```
mborgerson: Return value ignored here. Did you intend to do some error checking?

hw/xbox/xid-sdl.c
```diff
@@ -105,11 +109,12 @@ typedef struct USBXIDState {
     XIDGamepadOutputReport out_state_capabilities;
 
     uint8_t device_index;
+
+    HapticEmulationMode haptic_emulation_mode;
```
mborgerson: nit: Let's drop the blank lines around this new field

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +199,31 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
-                                            (SDL_HapticEffect*)&effect);
-        if (effect_id == -1) {
-            fprintf(stderr, "Failed to upload haptic effect!\n");
-            SDL_HapticClose(s->sdl_haptic);
-            s->sdl_haptic = NULL;
-            return;
-        }
-        SDL_HapticRunEffect(s->sdl_haptic, effect_id, 1);
+static void update_output(USBXIDState *s) {
+    if (s->sdl_haptic == NULL) { return; }
```
mborgerson: nit: Trailing statement should be on next line (according to checkpatch.pl)

hw/xbox/xid-sdl.c
DiscoStarslayer: We should probably not run `haptic_init` after the init fails. Technically the first call (`SDL_HapticOpened`) will always fail without the subsystem init, but skipping over this line is a good idea.

I return the error handling from `haptic_init` but I don't think handling anything at this level is really that useful, if something goes wrong inside `haptic_init` we log enough to `stderr` that a user should understand what happened. I return the error values just as a standard practice if we ever want to build some extra logic when an error occurs in the future.

If it is misleading though, I can just rip it out have have `haptic_init` return void.

hw/xbox/xid-sdl.c
DiscoStarslayer: nice catch :+1:

hw/xbox/xid-sdl.c
DiscoStarslayer: I felt like the other fields were more grouped by `sdl_` whereas this is a unique thing from that which is why I added the spacing. I'm not against moving it down though, I was on the fence about it anyway.

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +198,48 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
+                              uint16_t large_magnitude,
+                              uint16_t small_magnitude)
+{
+    switch (mode) {
+    case EMU_HAPTIC_LEFT_RIGHT:
+        effect->leftright = (SDL_HapticLeftRight){
+            .type = SDL_HAPTIC_LEFTRIGHT,
+            .length = SDL_HAPTIC_INFINITY,
+            /* Xbox rumble returns in a range of 0x0000 -> 0xFFFF,
+               SDL requires the range to be within 0x0000 -> 0x7FFF */
```
JayFoxRox: nit: I believe our comment style has leading * in multi-line comments.

hw/xbox/xid-sdl.c
```diff
@@ -511,18 +582,11 @@ static void usb_xbox_gamepad_realize(USBDevice *dev, Error **errp)
...
-        fprintf(stderr, "Joystick doesn't support haptic\n");
+    if (SDL_InitSubSystem(SDL_INIT_HAPTIC)) {
+        fprintf(stderr, "SDL failed to initialize haptic feedback subsystem\n");
     } else {
-        if ((SDL_HapticQuery(s->sdl_haptic) & SDL_HAPTIC_LEFTRIGHT) == 0) {
-            fprintf(stderr, "Joystick doesn't support necessary haptic effect\n");
-            SDL_HapticClose(s->sdl_haptic);
-            s->sdl_haptic = NULL;
-        }
+        haptic_init(s);
```
JayFoxRox: nit: I think we usually do: `if (noterror) { do stuff } else { handle-error }`.

I might be wrong though, but it would certainly make this bit of code easier to follow.

hw/xbox/xid-sdl.c
```diff
@@ -60,6 +59,11 @@ enum {
     STR_SERIALNUMBER,
 };
 
+typedef enum HapticEmulationMode {
+    EMU_HAPTIC_LEFT_RIGHT,
+    EMU_UNSUPPORTED
```
JayFoxRox: nit: I'd call this `EMU_NONE` or `EMU_DISABLED`, and put it first, so it's index 0.

If we ever allow user-overwrite for the mode (broken capabilities, or different modes supported) we don't have to rename stuff then. The suggested names are also still factually correct.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +478,63 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
     uc->handle_attach  = usb_desc_attach;
 }
 
+static HapticEmulationMode pick_haptic_emulation_mode(SDL_Haptic *haptic)
+{
+    if (SDL_HapticQuery(haptic) & SDL_HAPTIC_LEFTRIGHT) {
+        return EMU_HAPTIC_LEFT_RIGHT;
+    }
+
```
JayFoxRox: nit: For debug / user-support purposes, we could print the query result as hex value, so we can quickly decide on a plan of action if users report broken haptic for their device.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +478,63 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
     uc->handle_attach  = usb_desc_attach;
 }
 
+static HapticEmulationMode pick_haptic_emulation_mode(SDL_Haptic *haptic)
+{
+    if (SDL_HapticQuery(haptic) & SDL_HAPTIC_LEFTRIGHT) {
```
JayFoxRox: nit: Only query once, so if we add more modes, we don't have to re-query.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +478,63 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+        return EMU_HAPTIC_LEFT_RIGHT;
+    }
+
+    return EMU_UNSUPPORTED;
+}
+
+static int haptic_init(USBXIDState *s)
+{
+    if (SDL_HapticOpened(s->device_index)) {
+        fprintf(stderr, "Joystick haptic effect already in use, haptic effect disabled\n");
```
JayFoxRox: This isn't about the "haptic effect" being "in use", this is about the "haptic device" being "already opened".

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +478,63 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+        fprintf(stderr, "Joystick doesn't support haptic effects\n");
+        return -1;
+    }
+
+    s->haptic_emulation_mode = pick_haptic_emulation_mode(s->sdl_haptic);
+
+    if (s->haptic_emulation_mode == EMU_UNSUPPORTED) {
+        SDL_HapticClose(s->sdl_haptic);
+        s->sdl_haptic = NULL;
+        fprintf(stderr, "Joystick doesn't support necissary haptic effects\n");
```
JayFoxRox: necissary &rarr; necessary

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +479,66 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
     uc->handle_attach  = usb_desc_attach;
 }
 
+static HapticEmulationMode select_haptic_emulation_mode(SDL_Haptic *haptic)
+{
+    uint query_result = SDL_HapticQuery(haptic);
```
mborgerson: The type `uint` is undefined and causing compilation failures:

```C:/projects/xqemu/hw/xbox/xid-sdl.c:484:5: error: unknown type name 'uint'; did you mean 'guint'?```

We should probably follow [the docs](https://wiki.libsdl.org/SDL_HapticQuery) and just use `unsigned int` here.

File: hw/xbox/xid-sdl.c [hunk 52]
```diff
@@ -194,36 +198,49 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
+static void update_output(USBXIDState *s)
+{
+    if (s->sdl_haptic == NULL) {
+        return;
+    }
 
-        s->sdl_haptic_effect_id = effect_id;
+    SDL_HapticEffect effect;
+    int result = haptic_effect_prepare(&effect,
+                                    s->haptic_emulation_mode,
```
mborgerson: nit: Alignment

hw/xbox/xid-sdl.c
```diff
@@ -194,36 +198,49 @@ static const XIDDesc desc_xid_xbox_gamepad = {
...
+{
+    switch (mode) {
+    case EMU_HAPTIC_LEFT_RIGHT:
+        effect->leftright = (SDL_HapticLeftRight){
+            .type = SDL_HAPTIC_LEFTRIGHT,
+            .length = SDL_HAPTIC_INFINITY,
+            /* Xbox rumble returns in a range of 0x0000 -> 0xFFFF,
+             * SDL requires the range to be within 0x0000 -> 0x7FFF
+             */
+            .large_magnitude = large_magnitude >> 1,
```
mborgerson: After testing this PR, I don't think these `.large_magnitude` and `.small_magnitude` fields actually max out at `0x7FFF` as we had previously thought. Testing the haptic effects myself revealed that using `0x7FFF` has the rumble motors running well below full power. When I set the magnitude to `0xFFFF` however, the motors run at (what *feels like*) full power--certainly a noticeable difference over `0x7FFF`. So, I think we should probably pass the values here as-is and remove the comment/shift.

JayFoxRox: You are probably running an old version. This upstream bug was only fixed in May 2018, so the fix might not have landed yet in all distros:

https://github.com/spurious/SDL-mirror/commit/6066a1cf5fb6862f15cedaf6b3a81826203f78a1

We should implement features as documented and report issues upstream. Eventually all distros will be fixed.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +479,66 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
 }
 
+static HapticEmulationMode select_haptic_emulation_mode(SDL_Haptic *haptic)
+{
+    uint query_result = SDL_HapticQuery(haptic);
+    if (query_result & SDL_HAPTIC_LEFTRIGHT) {
+        return EMU_HAPTIC_LEFT_RIGHT;
+    }
+
+    printf("SDL_HapticQuery result: 0x%x\n", query_result);
```
JayFoxRox: nit: As this is a debug feature (although meant for users) we should either use a debug printing macro, or, at least describe the issue: `"No supported haptic mode found (SDL_HapticQuery result: 0x%x). Haptic disabled."` (or something like this)

DiscoStarslayer: I'll stick with your second suggestion here, as this is a user facing problem.

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +479,66 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
 }
 
+static HapticEmulationMode select_haptic_emulation_mode(SDL_Haptic *haptic)
+{
+    unsigned int query_result = SDL_HapticQuery(haptic);
+    if (query_result & SDL_HAPTIC_LEFTRIGHT) {
+        return EMU_HAPTIC_LEFT_RIGHT;
+    }
+
+    fprintf(stderr, "No supported haptic mode found (SDL_HapticQuery result: 0x%x). Haptic Disabled\n", query_result);
```
JayFoxRox: nit: Disabled &rarr; disabled (why is it capital D in the middle of statement?)

hw/xbox/xid-sdl.c
```diff
@@ -464,6 +479,66 @@ static void usb_xid_class_initfn(ObjectClass *klass, void *data)
...
+
+    fprintf(stderr, "No supported haptic mode found (SDL_HapticQuery result: 0x%x). Haptic disabled\n", query_result);
+
+    return EMU_NONE;
+}
+
+static int haptic_init(USBXIDState *s)
+{
+    if (SDL_HapticOpened(s->device_index)) {
+        fprintf(stderr, "Joystick haptic effect already opened, haptic effect disabled\n");
```
JayFoxRox: nit: This says "haptic effect disabled". We have a similar case for when no mode was found [a couple lines above] that uses "haptic disabled". I think the second case without effect is right, because no effect is involved here.

JayFoxRox:
I'll probably do a final review pass in the next days.
Most of my review in the past days focused on code quality and consistent messages. I did not review the design in the final iteration.

I also haven't tested this code yet either, and intend to do that soon.

However, final review should be by @mborgerson anyway (or other members of the community), because I've done early review commits and provided SDL example code - so the code is largely written around my ideas and based on my decisions.

mborgerson:
Thank you @DiscoStarslayer! Enabling force feedback is an exciting change that I'm sure our users will appreciate. I know I do!

PR #198, opened 2019-02-11
chihiro: Fix property names by avoiding illegal symbols

0DaveX:
Trying to run chihiro currently fails with `qemu-system-i386: Property '.mediaboard-rom' not found
`

As @JayFoxRox pointed out that's due to underscores '_' being kind of not allowed.
(They are replaced by hyphens '-' )

https://github.com/xqemu/xqemu/blob/960698a24ed7cd69fe2e452a28a3bb64e395a0fe/vl.c#L2706-L2711

### DIFF
--- a/hw/xbox/chihiro.c
+++ b/hw/xbox/chihiro.c
@@ -321,9 +321,9 @@ static const uint8_t eeprom[] = {
 static void chihiro_init(MachineState *machine)
 {
     const char *mediaboard_rom_file = object_property_get_str(
-        qdev_get_machine(), "mediaboard_rom", NULL);
+        qdev_get_machine(), "mediaboard-rom", NULL);
     const char *mediaboard_filesystem_file = object_property_get_str(
-        qdev_get_machine(), "mediaboard_filesystem", NULL);
+        qdev_get_machine(), "mediaboard-filesystem", NULL);
     chihiro_ide_interface_init(mediaboard_rom_file,
                                mediaboard_filesystem_file);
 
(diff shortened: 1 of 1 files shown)


JayFoxRox:
What about these 2 occurences: https://github.com/xqemu/xqemu/pull/198/files#diff-d955b87b6d9ffb6f8825a9bc1ca5ca3fR323

0DaveX:
thx,
changed the other occurrences, too

hw/xbox/chihiro.c
```diff
@@ -388,13 +388,13 @@ static void machine_set_mediaboard_filesystem(Object *obj, const char *value,
 
 static inline void chihiro_machine_initfn(Object *obj)
 {
-    object_property_add_str(obj, "mediaboard_rom",
+    object_property_add_str(obj, "mediaboard-rom",
                             machine_get_mediaboard_rom,
                             machine_set_mediaboard_rom, NULL);
     object_property_set_description(obj, "mediaboard_rom",
```
JayFoxRox: What about these? I missed them in my first pass. Please grep / "Find" all of these strings please, to make absolutely sure you got them all.

It might be worth to even check other components in /hw/xbox by grepping for `object_property_`: `grep -n "object_property_" -R | grep _`

0DaveX: Thanks!
I don't find any more occurrences (by us)

mborgerson:
Thanks!

PR #199, opened 2019-02-11
Final combiner G input is a scalar

JayFoxRox:
Minor changes to our PSH code.

G should be an alpha component, as specified in [NV_register_combiners](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_register_combiners.txt):

> The variable G is an alpha scalar.

I've changed the types to make this safer, this includes the pseudo-registers:

> The alpha component of E_TIMES_F_NV and SPARE0_PLUS_SECONDARY_COLOR_NV is always zero.

There's still a lot of hacky code left, such as casts to `vec3`, so this *does not* solve #202, yet.
That will be solved in a more complicated follow-up.

---

I did only test this with the dashboard and Smashing Drive. The final version I'm sending as PR **already contains some last-minute fixups and was never tested**; I'll run tests during review.

**Should be tested with a handful of games to make sure nothing breaks due to type changes.**

### DIFF
--- a/hw/xbox/nv2a/nv2a_psh.c
+++ b/hw/xbox/nv2a/nv2a_psh.c
@@ -296,24 +296,30 @@ static QString* get_input_var(struct PixelShader *ps, struct InputInfo in, bool 
 {
     QString *reg = get_var(ps, in.reg, false);
 
-    switch (in.chan) {
-    case PS_CHANNEL_RGB:
-        if (is_alpha) {
-            qstring_append(reg, ".b");
-        } else {
+    if (!is_alpha) {
+        switch (in.chan) {
+        case PS_CHANNEL_RGB:
             qstring_append(reg, ".rgb");
-        }
-        break;
-    case PS_CHANNEL_ALPHA:
-        if (is_alpha) {
-            qstring_append(reg, ".a");
-        } else {
+            break;
+        case PS_CHANNEL_ALPHA:
             qstring_append(reg, ".aaa");
+            break;
+        default:
+            assert(false);
+            break;
         }
-        break;
-    default:
-        assert(false);
-        break;
+    } else {
+        switch (in.chan) {
+        case PS_CHANNEL_BLUE:
+            qstring_append(reg, ".b");
+            break;
+        case PS_CHANNEL_ALPHA:
+            qstring_append(reg, ".a");
+            break;
+        default:
+            assert(false);
+            break;
+        }
     }
 
     QString *res;
--- a/hw/xbox/nv2a/nv2a_psh.c
+++ b/hw/xbox/nv2a/nv2a_psh.c
@@ -235,7 +235,7 @@ static QString* get_var(struct PixelShader *ps, int reg, bool is_dest)
         if (is_dest) {
             return qstring_from_str("");
         } else {
-            return qstring_from_str("0.0");
+            return qstring_from_str("vec4(0.0)");
         }
         break;
     case PS_REGISTER_C0:
@@ -280,9 +280,10 @@ static QString* get_var(struct PixelShader *ps, int reg, bool is_dest)
         return qstring_from_str("r1");
     case PS_REGISTER_V1R0_SUM:
         add_var_ref(ps, "r0");
-        return qstring_from_str("(v1 + r0)");
+        return qstring_from_str("vec4(v1.rgb + r0.rgb, 0.0)");
     case PS_REGISTER_EF_PROD:
-        return qstring_from_fmt("(%s * %s)", qstring_get_str(ps->varE),
+        return qstring_from_fmt("vec4(%s * %s, 0.0)",
+                                qstring_get_str(ps->varE),
                                 qstring_get_str(ps->varF));
     default:
         assert(false);
@@ -295,28 +296,24 @@ static QString* get_input_var(struct PixelShader *ps, struct InputInfo in, bool 
 {
     QString *reg = get_var(ps, in.reg, false);
 
-    if (strcmp(qstring_get_str(reg), "0.0") != 0
-        && (in.reg != PS_REGISTER_EF_PROD
-            || strstr(qstring_get_str(reg), ".a") == NULL)) {
-        switch (in.chan) {
-        case PS_CHANNEL_RGB:
-            if (is_alpha) {
-                qstring_append(reg, ".b");
-            } else {
-                qstring_append(reg, ".rgb");
-            }
-            break;
-        case PS_CHANNEL_ALPHA:
-            if (is_alpha) {
-                qstring_append(reg, ".a");
-            } else {
-                qstring_append(reg, ".aaa");
-            }
-            break;
-        default:
-            assert(false);
-            break;
+    switch (in.chan) {
+    case PS_CHANNEL_RGB:
+        if (is_alpha) {
+            qstring_append(reg, ".b");
+        } else {
+            qstring_append(reg, ".rgb");
         }
+        break;
+    case PS_CHANNEL_ALPHA:
+        if (is_alpha) {
+            qstring_append(reg, ".a");
+        } else {
+            qstring_append(reg, ".aaa");
+        }
+        break;
+    default:
+        assert(false);
+        break;
     }
 
     QString *res;
--- a/hw/xbox/nv2a/nv2a_psh.c
+++ b/hw/xbox/nv2a/nv2a_psh.c
@@ -307,7 +307,11 @@ static QString* get_input_var(struct PixelShader *ps, struct InputInfo in, bool 
             }
             break;
         case PS_CHANNEL_ALPHA:
-            qstring_append(reg, ".a");
+            if (is_alpha) {
+                qstring_append(reg, ".a");
+            } else {
+                qstring_append(reg, ".aaa");
+            }
             break;
         default:
             assert(false);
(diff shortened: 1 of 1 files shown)


Ale100000:
This PR fix  ```nv2a: fragment shader compilation failed: 0(64) : error C7011: implicit cast from "float" to "vec3" ```crash  in Stars Wars Battlefront 2.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -280,9 +280,10 @@ static QString* get_var(struct PixelShader *ps, int reg, bool is_dest)
         return qstring_from_str("r1");
     case PS_REGISTER_V1R0_SUM:
         add_var_ref(ps, "r0");
-        return qstring_from_str("(v1 + r0)");
+        return qstring_from_str("vec4(v1.xyz + r0.xyz, 0.0)");
```
JayFoxRox: nit: this should be `.rgb`

PatrickvL: `v1.rgb` or `r0.rgb` or both?

JayFoxRox: Given that this belongs to register combiners: everything.

The register combiners only deal with colors, in particular only RGB [`.rgb`], or only ALPHA [`.a`].
It makes no sense to ever use `.xyzw` in this code section; everything should be `.rgba` (ideally split: `.rgb` + `.a`)

PatrickvL:
PR title could use a `psh: ` prefix

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -485,13 +486,13 @@ static void add_final_stage_code(struct PixelShader *ps, struct FCInputInfo fina
     QString *b = get_input_var(ps, final.b, false);
     QString *c = get_input_var(ps, final.c, false);
     QString *d = get_input_var(ps, final.d, false);
-    QString *g = get_input_var(ps, final.g, false);
+    QString *g = get_input_var(ps, final.g, true);
 
     qstring_append_fmt(ps->code, "fragColor.rgb = %s + mix(vec3(%s), vec3(%s), vec3(%s));\n",
                        qstring_get_str(d), qstring_get_str(c),
                        qstring_get_str(b), qstring_get_str(a));
     /* FIXME: Is .x correctly here? */
```
mborgerson: If we drop the `vec3().x` construct, we should remove this comment too.

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -280,9 +280,10 @@ static QString* get_var(struct PixelShader *ps, int reg, bool is_dest)
         return qstring_from_str("r1");
     case PS_REGISTER_V1R0_SUM:
         add_var_ref(ps, "r0");
-        return qstring_from_str("(v1 + r0)");
+        return qstring_from_str("vec4(v1.rgb + r0.rgb, 0.0)");
```
mborgerson: Good catch!

*Note to other reviewers: `PS_REGISTER_V1R0_SUM` refers to the pseudo register listed in the spec as `SPARE0_PLUS_SECONDARY_COLOR_NV` in which "alpha component ... is always zero." [[§3.8.12.3](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_register_combiners.txt)] For historic reasons this file uses Xbox DirectX PSH terminology (mentioned in comment at head of file by espes). Eventually we may want to switch this over to the NVIDIA spec terminology, or at least have some comments describing the name translation, to make cross-referencing with the spec easier.*

The spec also mentions:
> If csc, the color sum clamp, is non-FALSE, the value of SPARE0_PLUS_SECONDARY_COLOR_NV is first clamped to the range [0,1].

And it seems like the code has already been set up for the flag (`ps.final_input.clamp_sum`), it's just not being used:
```
PS_FINALCOMBINERSETTING_CLAMP_SUM=     0x80, // V1+R0 sum clamped to [0,1]
...
ps.final_input.clamp_sum = flags & PS_FINALCOMBINERSETTING_CLAMP_SUM;
```

Do you think we should we also apply clamping at this point while we are here?

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -295,24 +296,24 @@ static QString* get_input_var(struct PixelShader *ps, struct InputInfo in, bool
...
+        break;
+    case PS_CHANNEL_ALPHA:
+        if (is_alpha) {
             qstring_append(reg, ".a");
-            break;
-        default:
-            assert(false);
-            break;
+        } else {
+            qstring_append(reg, ".aaa");
```
mborgerson: Makes sense:
```
     while an ALPHA <componentUsage> parameter indicates that the
     alpha portion of the indicated register should be replicated across
     the RGB portion of the combiner input variable.
```

hw/xbox/nv2a/nv2a_psh.c
```diff
@@ -295,24 +296,24 @@ static QString* get_input_var(struct PixelShader *ps, struct InputInfo in, bool
...
-        case PS_CHANNEL_RGB:
-            if (is_alpha) {
-                qstring_append(reg, ".b");
-            } else {
-                qstring_append(reg, ".rgb");
-            }
-            break;
-        case PS_CHANNEL_ALPHA:
+    switch (in.chan) {
+    case PS_CHANNEL_RGB:
```
mborgerson: nit: While you're here, for readability (the values are the same, so it makes no functional difference...), would you please add `case PS_CHANNEL_BLUE:` here to show that we are handling that with the same logic?

hw/xbox/nv2a/nv2a_psh.c
JayFoxRox: Oops! Will fix this.

hw/xbox/nv2a/nv2a_psh.c
JayFoxRox: I believe that would trigger a duplicated-case error? Restructuring the code to check `is_alpha` first (and having 2 switch-statements) is kind of dumb, because I want to separate the `PS_CHANNEL_RGB` and `PS_CHANNEL_ALPHA` more strictly in the future (as done in the spec).

We should just remove `PS_CHANNEL_BLUE` in my opinion.
We don't have `PS_CHANNEL_AAA` either.

It's just fact that `RGB` mapped to `A`, is using the `B` channel. I'm not even sure if adding a comment ("// Only B component is used when RGB is accessed as alpha") would add any readability.

hw/xbox/nv2a/nv2a_psh.c
JayFoxRox: The old code also did this, but using ugly `vec3` casts when accessing the values (which still exist, and will be cleaned up in the future).

See #202

hw/xbox/nv2a/nv2a_psh.c
JayFoxRox: Yes, see #207 - I have some code which I'll send in the future, it needs some other fixes first.

mborgerson: Cool! We can follow up on this with subsequent PR then.

hw/xbox/nv2a/nv2a_psh.c
mborgerson: Agree, the old `vec3` casting approach was ugly. This method is much more readable and clearly indicates desired action/intent at this point. Thanks!

hw/xbox/nv2a/nv2a_psh.c
mborgerson: I'm not sure about the duplicate case (haven't tried it out yet), maybe it would cause compilation error. If that's the case, we could probably add a comment at least. The reason I mentioned documenting it here, and why we may want to keep `PS_CHANNEL_BLUE` around (*though a full NVIDIA spec rename overhaul at some point would be nice...*), is because the spec specifically mentions it:
```
     The <componentUsage> parameter may be one of RGB, ALPHA, or BLUE.

     When the <portion> parameter is RGB, a <componentUsage> parameter
     of RGB indicates that the RGB portion of the indicated register
     should be assigned to the RGB portion of the combiner input variable,
     while an ALPHA <componentUsage> parameter indicates that the
     alpha portion of the indicated register should be replicated across
     the RGB portion of the combiner input variable.

     When the <portion> parameter is ALPHA, the <componentUsage>
     parameter of ALPHA indicates that the alpha portion of the indicated
     register should be assigned to the alpha portion of the combiner
     input variable, while a BLUE <componentUsage> parameter indicates
     that the blue component of the indicated register should be assigned
     to the alpha portion of the combiner input variable.
```

JayFoxRox: Very good comment!

I didn't remember the spec also had this.

I also noticed that my previous explanation was wrong: I want to restructure code based on `is_alpha`, not `in.channel` (`PS_CHANNEL_RGB` and `PS_CHANNEL_ALPHA`).

\- I'll change my code and split this into 2 separate switch cases or equivalent.

JayFoxRox: Added a new commit to do this.

I did not test this change myself.

hw/xbox/nv2a/nv2a_psh.c
JayFoxRox: Done

mborgerson:
Thank you!

PR #208, opened 2019-02-11
psh: Use input mapping from NV_register_combiners GL extension

JayFoxRox:
[This should be closer to NV_register_combiners](https://www.khronos.org/registry/OpenGL/extensions/NV/NV_register_combiners.txt).

I've had this code for a while and never bothered to fully test or PR it.

So this needs a bit of testing - but I assume it's correct.


mborgerson:
Thanks! Makes sense to me based on §3.8.12.2, Table NV_register_combiners.4.

PR #211, opened 2019-02-18
Fix vertex-blending (Skinning)

JayFoxRox:
This does 3 separate, but closely related things, to fix skinning. The motivation is an issue in ["Dead or Alive Xtreme Beach Volleyball"](https://www.youtube.com/watch?v=6Ufbyr33hKQ) (also [see this timestamp](https://youtu.be/euoh-Jlvthg?t=27)).

When applying this PR, it will look correct:

Fixed

The changes are detailed below.

#### Fix skinning mode order

This commit makes the code select the correct skinning mode, but the emulation is still limited by old skinning code (only 2 matrices allowed).

This was broken since the [original PR to espes repo in 2015](https://github.com/espes/xqemu/pull/14). I somehow re-arranged the enum, possibly for readability, so it didn't match the NV2A order anymore.

We'd ideally solve this with a mapping function between NV2A and our shader generator, but none of the other existing code (foggen / texgen / ..)  does this, so we should clean it up later.

#### Support generated weights in skinning

This commit allows more than 2 matrices / blend weights.

This implements the vertex blending as described in the following documents:

* https://docs.microsoft.com/en-us/windows/desktop/direct3d9/geometry-blending
* https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_vertex_blend.txt

To implement this, I decided to start with `weight_n = 1.0` which will contain the weight of the final n-th matrix weight.
Each other matrix will then use the current weight from vertex data (`weight_i = weight.%c`) and subtract it from the final matrix weight.
For the final n-th matrix, only `weight_i = weight_n` remains, so the sum of all weights is `1.0`.
An uncontrolled `weight.%c` is not accessed with this algorithm.

I also looked at, but could not find any references which hint existence of these features:

* https://docs.microsoft.com/en-us/windows/desktop/direct3d9/vertex-tweening
* https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_matrix_palette.txt

I've therefore changed the comment which referred to this mode "tweening".

#### Minor improvements to individually weighted skinning 

This fixes the comment in the code that is probably the non-generated case mentioned in the ARB_vertex_blend specification.

I also felt the new style of doing the blending last (after swizzle) was nicer, because it does weight on fewer components, so I copied that style from the other change.

The `assert(false)` remains, because this code still wasn't tested, but is assumed to work.

---

**Notes:**

With this new code, we could also support `SKINNING_OFF` in the auto-generated code path with `count = 1`. We can even handle the generated weight and individual weight case using the same code.
However, I didn't do any of that as I felt the intend of the code would be lost. Maybe we can do it in the future if we cleanup/refactor and find a nicer way to write it.

I only tested with "Dead or Alive Xtreme Beach Volleyball", so further tests in other games would be nice.


hw/xbox/nv2a/nv2a_shaders.c
```diff
@@ -246,26 +246,31 @@ static void append_skinning_code(QString* str, bool mix,
     } else {
         qstring_append_fmt(str, "%s %s = %s(0.0);\n", type, output, type);
         if (mix) {
```
JayFoxRox: nit: This `mix` name isn't good anymore. I could rename it to `sum_unity` or something.

Ale100000:
Fixed Character Models in Chicago Enforcer.
<details><summary>(Click to unfold screenshots)</summary><br>

**Master**
a
**PR**
b




</details>

JayFoxRox:
I have made a list of possibly affected games (from John GodGames test list) - because of the wrong order, I'm not sure if these games use auto-weights or not:

* Disney's The Haunted Mansion
* Tron 2.0: Killer App
* Zathura
* Dead or Alive: Xtreme Beach Volleyball
* Knight's Apprentice: Memorick's Adventures
* Crazy Taxi 3: High Roller
* Barbarian
* Playboy: The Mansion
* Red Faction II
* Ski Racing 2006 Featuring Hermann Maier
* Sonic Riders
* Stacked with Daniel Negreanu
* Tennis Masters Series 2003
* The Punisher
* Van Helsing

We should re-test as many games as possible, so we can also fix individual skinning in a follow-up (or even in this PR if we can quickly find one).

Ernegien:
This has a noticeable improvement in The Punisher as well.

<details><summary>(Click to unfold screenshots)</summary>
<br/>

**Master**
image

**PR**
image

</details>

JohnGodgames:
**Zathura**
Master:
zathura1
Pr:
zathura2

**Crazy Taxi 3: High Roller**
Master:
crazytaxi 3 1
Pr:
crazytaxi 3 2

**Barbarian**
Master:
barbarian 1
Pr:
barbarian 2

**Playboy: The Mansion**
Master:
playboy 1
Pr:
playboy 2

**Red Faction II**
Master:
red faction 2 1
Pr:
red faction 2 2

**Tennis Masters Series 2003**
Master:
tennis 1
Pr:
tennis 2

**Van Helsing**
Master:
van helsing 1
Pr:
van helsing 2

PR #216, opened 2019-02-23
Support PGRAPH RDI

JayFoxRox:
This implements the PGRAPH RDI.
RDI is an interface to access the RAM of PGRAPH. We might also PFB RDI in the future, but this is not implemented here.

On hardware, the GPU has to be idle before RDI can be used (apparently RDI shares some address lines with other GPU features). For now, this restriction will not be emulated in XQEMU. We can always make it more accurate in the future.

PGRAPH RDI is irrelevant for game emulation, because D3D drivers use it for features we probably don't need to emulate (optimization, debugging, ..). However, nv2a-trace will depend on this functionality, as some states can not be accessed through other interfaces.

I chose to silently ignore write, unless debug mode is used. The fact that D3D uses this prevents us from `assert(false)` on *writes*.
I chose to `assert('false')` on *reads*, because it would put emulation at risk if something bad is returned. Fortunately D3D doesn't seem to read either.

I've added code for `select` 0x17 and 0xCC as example, which access vertex program constants. They mostly seem to access the same data (probably for 2 separate transform-engines) and I doubt we'll have to emulate these banks going out of sync.

After merge, I'll create issues about other `select` values which will be required for this feature to be useful. Issues will only focus on those features we need, not completeness.

**This should be tested to avoid accidental `assert`. I've only done little testing with the MS dashboard and nothing else.**

### DIFF
--- a/hw/xbox/nv2a/nv2a_pgraph.c
+++ b/hw/xbox/nv2a/nv2a_pgraph.c
@@ -302,23 +302,24 @@ static uint64_t fast_hash(const uint8_t *data, size_t len, unsigned int samples)
 uint64_t pgraph_read(void *opaque, hwaddr addr, unsigned int size)
 {
     NV2AState *d = (NV2AState *)opaque;
+    PGRAPHState *pg = &d->pgraph;
 
-    qemu_mutex_lock(&d->pgraph.lock);
+    qemu_mutex_lock(&pg->lock);
 
     uint64_t r = 0;
     switch (addr) {
     case NV_PGRAPH_INTR:
-        r = d->pgraph.pending_interrupts;
+        r = pg->pending_interrupts;
         break;
     case NV_PGRAPH_INTR_EN:
-        r = d->pgraph.enabled_interrupts;
+        r = pg->enabled_interrupts;
         break;
     default:
-        r = d->pgraph.regs[addr];
+        r = pg->regs[addr];
         break;
     }
 
-    qemu_mutex_unlock(&d->pgraph.lock);
+    qemu_mutex_unlock(&pg->lock);
 
     reg_log_read(NV_PGRAPH, addr, r);
     return r;
@@ -327,37 +328,39 @@ uint64_t pgraph_read(void *opaque, hwaddr addr, unsigned int size)
 void pgraph_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size)
 {
     NV2AState *d = (NV2AState *)opaque;
+    PGRAPHState *pg = &d->pgraph;
 
     reg_log_write(NV_PGRAPH, addr, val);
 
-    qemu_mutex_lock(&d->pgraph.lock);
+    qemu_mutex_lock(&pg->lock);
 
     switch (addr) {
     case NV_PGRAPH_INTR:
-        d->pgraph.pending_interrupts &= ~val;
-        qemu_cond_broadcast(&d->pgraph.interrupt_cond);
+        pg->pending_interrupts &= ~val;
+        qemu_cond_broadcast(&pg->interrupt_cond);
         break;
     case NV_PGRAPH_INTR_EN:
-        d->pgraph.enabled_interrupts = val;
+        pg->enabled_interrupts = val;
         break;
     case NV_PGRAPH_INCREMENT:
         if (val & NV_PGRAPH_INCREMENT_READ_3D) {
-            SET_MASK(d->pgraph.regs[NV_PGRAPH_SURFACE],
+            SET_MASK(pg->regs[NV_PGRAPH_SURFACE],
                      NV_PGRAPH_SURFACE_READ_3D,
-                     (GET_MASK(d->pgraph.regs[NV_PGRAPH_SURFACE],
+                     (GET_MASK(pg->regs[NV_PGRAPH_SURFACE],
                               NV_PGRAPH_SURFACE_READ_3D)+1)
-                        % GET_MASK(d->pgraph.regs[NV_PGRAPH_SURFACE],
+                        % GET_MASK(pg->regs[NV_PGRAPH_SURFACE],
                                    NV_PGRAPH_SURFACE_MODULO_3D) );
-            qemu_cond_broadcast(&d->pgraph.flip_3d);
+            qemu_cond_broadcast(&pg->flip_3d);
         }
         break;
     case NV_PGRAPH_CHANNEL_CTX_TRIGGER: {
         hwaddr context_address =
-            GET_MASK(d->pgraph.regs[NV_PGRAPH_CHANNEL_CTX_POINTER], NV_PGRAPH_CHANNEL_CTX_POINTER_INST) << 4;
+            GET_MASK(pg->regs[NV_PGRAPH_CHANNEL_CTX_POINTER],
+                     NV_PGRAPH_CHANNEL_CTX_POINTER_INST) << 4;
 
         if (val & NV_PGRAPH_CHANNEL_CTX_TRIGGER_READ_IN) {
             unsigned pgraph_channel_id =
-                GET_MASK(d->pgraph.regs[NV_PGRAPH_CTX_USER], NV_PGRAPH_CTX_USER_CHID);
+                GET_MASK(pg->regs[NV_PGRAPH_CTX_USER], NV_PGRAPH_CTX_USER_CHID);
 
             NV2A_DPRINTF("PGRAPH: read channel %d context from %" HWADDR_PRIx "\n",
                          pgraph_channel_id, context_address);
@@ -369,7 +372,7 @@ void pgraph_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size)
 
             NV2A_DPRINTF("    - CTX_USER = 0x%x\n", context_user);
 
-            d->pgraph.regs[NV_PGRAPH_CTX_USER] = context_user;
+            pg->regs[NV_PGRAPH_CTX_USER] = context_user;
             // pgraph_set_context_user(d, context_user);
         }
         if (val & NV_PGRAPH_CHANNEL_CTX_TRIGGER_WRITE_OUT) {
@@ -379,18 +382,18 @@ void pgraph_write(void *opaque, hwaddr addr, uint64_t val, unsigned int size)
         break;
     }
     default:
-        d->pgraph.regs[addr] = val;
+        pg->regs[addr] = val;
         break;
     }
 
     // events
     switch (addr) {
     case NV_PGRAPH_FIFO:
-        qemu_cond_broadcast(&d->pgraph.fifo_access_cond);
+        qemu_cond_broadcast(&pg->fifo_access_cond);
         break;
     }
 
-    qemu_mutex_unlock(&d->pgraph.lock);
+    qemu_mutex_unlock(&pg->lock);
 }
 
 static void pgraph_method(NV2AState *d,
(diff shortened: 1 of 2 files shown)


hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -299,26 +299,81 @@ static uint64_t fnv_hash(const uint8_t *data, size_t len);
...
+        unsigned int select = GET_MASK(pg->regs[NV_PGRAPH_RDI_INDEX],
+                                       NV_PGRAPH_RDI_INDEX_SELECT);
+        unsigned int address = GET_MASK(pg->regs[NV_PGRAPH_RDI_INDEX],
+                                        NV_PGRAPH_RDI_INDEX_ADDRESS);
+
+        r = pgraph_rdi_read(pg, select, address);
+
+        /* FIXME: Overflow into select? */
+        assert(address < GET_MASK(NV_PGRAPH_RDI_INDEX_ADDRESS,
+                                  NV_PGRAPH_RDI_INDEX_ADDRESS));
```
JayFoxRox: Untested; I might run a unit test for this in the next days.

(Other instances too)

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -299,26 +299,81 @@ static uint64_t fnv_hash(const uint8_t *data, size_t len);
...
+
+static uint32_t pgraph_rdi_read(PGRAPHState *pg,
+                                unsigned int select, unsigned int address)
+{
+    uint32_t r = 0;
+    switch(select) {
+    case RDI_INDEX_VTX_CONSTANTS0:
+    case RDI_INDEX_VTX_CONSTANTS1:
+        assert((address / 4) < NV2A_VERTEXSHADER_CONSTANTS);
+        r = pg->vsh_constants[address / 4][3 - address % 4];
```
JayFoxRox: If RDI always sends 4 words in reverse order, maybe do: `address ^= 0x3` in the beginning, or even `unsigned int part = address % 4; address /= 4`?

I might run a unit test for this in the next days.

hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -1282,6 +1286,9 @@
 #define NV_IGRAPH_XF_LTC1_L6                         0x12
 #define NV_IGRAPH_XF_LTC1_L7                         0x13
 
+#define RDI_INDEX_VTX_CONSTANTS0                     0x17
+#define RDI_INDEX_VTX_CONSTANTS1                     0xCC
```
JayFoxRox: These appear to be names given by MS.

nvidia seems to refer to `RDI_INDEX_VTX_CONSTANTS0` by `RDI_RAMSEL_XL_XFCTX`. However, `RDI_INDEX_VTX_CONSTANTS1` doesn't seem to have an nvidia name.

I suggest we update these names when they are named in nouveau.

We can still use nvidia names (if we find them somewhere) in other instances - but I'd rather not mix MS and nvidia names for closely related registers.

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -299,26 +299,81 @@ static uint64_t fnv_hash(const uint8_t *data, size_t len);
...
+    case RDI_INDEX_VTX_CONSTANTS1:
+        assert(false); /* Untested */
+        assert((address / 4) < NV2A_VERTEXSHADER_CONSTANTS);
+        pg->vsh_constants_dirty[address / 4] |=
+            (val != pg->vsh_constants[address / 4][3 - address % 4]);
+        pg->vsh_constants[address / 4][3 - address % 4] = val;
+        break;
+    default:
+        fprintf(stderr, "nv2a: unknown rdi write select 0x%x, address 0x%x\n", select, address);
+        assert(false);
```
DiscoStarslayer: Should we drop the assert here? From your PR comments:
"The fact that D3D uses this prevents us from `assert(false)` on writes."
"After merge, I'll create issues about other `select` values which will be required for this feature to be useful. Issues will only focus on those features we need, not completeness."

You seem to suggest we don't want to assert here, should we remove `assert(false)` for non-implemented rdi write selects?

JayFoxRox: Good catch! I had fixes that I forgot to push.

hw/xbox/nv2a/nv2a_pgraph.c
JayFoxRox: Got no time for this at the moment. Code can remain as-is and improved later.

hw/xbox/nv2a/nv2a_pgraph.c
JayFoxRox: Got no time for this at the moment. Code can remain as-is and improved later; it will trigger an assert before unknown behaviour.

dracc:
I've tested with some nxdk applications (NevoX, the mesh sample and httpd_bsd), Disney's Cars, Forza Motorsport and Blinx - The Time Sweeper. No new assertions were hit. :+1:

hw/xbox/nv2a/nv2a_regs.h
mborgerson: If we mix name conventions here, we should probably leave a comment here indicating what/why this was done and what the alternative naming is. Also see comment above with regard to `CONSTANTS1`.

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -299,26 +299,82 @@ static uint64_t fnv_hash(const uint8_t *data, size_t len);
...
 
 /* PGRAPH - accelerated 2d/3d drawing engine */
+
+static uint32_t pgraph_rdi_read(PGRAPHState *pg,
+                                unsigned int select, unsigned int address)
+{
+    uint32_t r = 0;
+    switch(select) {
+    case RDI_INDEX_VTX_CONSTANTS0:
+    case RDI_INDEX_VTX_CONSTANTS1:
```
mborgerson: Are we confident `RDI_INDEX_VTX_CONSTANTS1` is valid? Have we confirmed that `0xCC` select works on real hardware? If they address the same data, why would you use `CONSTANTS1` instead of `CONSTANTS0`?

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -299,26 +299,82 @@ static uint64_t fnv_hash(const uint8_t *data, size_t len);
...
+    case NV_PGRAPH_RDI_DATA: {
+        unsigned int select = GET_MASK(pg->regs[NV_PGRAPH_RDI_INDEX],
+                                       NV_PGRAPH_RDI_INDEX_SELECT);
+        unsigned int address = GET_MASK(pg->regs[NV_PGRAPH_RDI_INDEX],
+                                        NV_PGRAPH_RDI_INDEX_ADDRESS);
+
+        r = pgraph_rdi_read(pg, select, address);
+
+        /* FIXME: Overflow into select? */
+        assert(address < GET_MASK(NV_PGRAPH_RDI_INDEX_ADDRESS,
```
mborgerson: Should we check range of `address` *before* `pgraph_rdi_read`?

hw/xbox/nv2a/nv2a_pgraph.c
```diff
@@ -327,37 +383,54 @@ uint64_t pgraph_read(void *opaque, hwaddr addr, unsigned int size)
...
+    case NV_PGRAPH_RDI_DATA: {
+        unsigned int select = GET_MASK(pg->regs[NV_PGRAPH_RDI_INDEX],
+                                       NV_PGRAPH_RDI_INDEX_SELECT);
+        unsigned int address = GET_MASK(pg->regs[NV_PGRAPH_RDI_INDEX],
+                                        NV_PGRAPH_RDI_INDEX_ADDRESS);
+
+        pgraph_rdi_write(pg, select, address, val);
+
+        /* FIXME: Overflow into select? */
+        assert(address < GET_MASK(NV_PGRAPH_RDI_INDEX_ADDRESS,
```
mborgerson: Should we check range of `address` *before* `pgraph_rdi_write`?

hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -256,6 +256,10 @@
 #   define NV_PGRAPH_INCREMENT_READ_3D                          (1 << 1)
 #define NV_PGRAPH_FIFO                                   0x00000720
 #   define NV_PGRAPH_FIFO_ACCESS                                (1 << 0)
+#define NV_PGRAPH_RDI_INDEX                              0x00000750
+#   define NV_PGRAPH_RDI_INDEX_ADDRESS                        0x00000FFC
```
mborgerson: Are we sure this isn't `0x1FFC`? MMIO region is 0x2000, so it looks suspicious

hw/xbox/nv2a/nv2a_regs.h
```diff
@@ -256,6 +256,10 @@
...
+#   define NV_PGRAPH_RDI_INDEX_SELECT                         0x00FF0000
```
mborgerson: Are we sure this is an 8 bit field?

hw/xbox/nv2a/nv2a_pgraph.c
JayFoxRox: > If they address the same data, why would you use CONSTANTS1 instead of CONSTANTS0?

Described in PR description:

> I've added code for select 0x17 and 0xCC as example, which access vertex program constants. **They mostly seem to access the same data (probably for 2 separate transform-engines) and I doubt we'll have to emulate these banks going out of sync**.

---

> Are we confident RDI_INDEX_VTX_CONSTANTS1 is valid? Have we confirmed that 0xCC select works on real hardware?

I believe I did, but I don't fully remember anymore (this code is based on work from 2017; and I haven't  touched it again since February 2019 - when this PR was created - either).

I checked existing code, and found this:

- [nv2a-re only seems to use 0x17](https://github.com/JayFoxRox/nv2a-re/blob/37a425ceb29f863372037c6f6539d1c45f11bd24/main.c#L159-L161).
- [nv2a-trace (a development branch) seems to use both](https://github.com/JayFoxRox/nv2a-trace/blob/b38f2e55473366380a1040658ace68f4169f1dfd/Trace.py#L249-L250).

I can't seem to find any dumps I've made with nv2a-trace though, I'm still somewhat confident it's almost always the same data.

hw/xbox/nv2a/nv2a_regs.h
JayFoxRox: The mask is what D3D debug libs use.

What MMIO region? RDI probably exists because there's no MMIO to this GPU RAM.

This is an index into the selected bank (`NV_PGRAPH_RDI_INDEX_SELECT` to pick bank of a feature). So there can be up-to 1024 words per bank, but probably not all of those are used for each feature.

If you pick the bank for the constant space bank for example, it probably only has 192 vec4 constants = 768 valid indices.
The program space bank should only have 136 * 4 words per instruction = 544 valid indices.
\- So the exact size of each bank could be different.

I'm not sure what happens at higher indices, but we shouldn't have to worry about it yet; I have added an `assert` to guard against higher indices in the constant banks.

hw/xbox/nv2a/nv2a_regs.h
JayFoxRox: I'll move my comment into the code then, although we have other instances were it appears that non-nvidia names have been chosen, with no hints (like most `NV2A_*` defines right below).

mborgerson:
Thanks!

PR #217, opened 2019-03-03
Merge QEMU tag 'v3.1.0'

mborgerson:
# Don't merge this
I'll merge it manually when its ready.

Still debugging HVF issues. I've got it working in a local branch but need to reconcile upstream changes.

Other things should be smoke tested.


mborgerson:
Filing new PR with 4.0.0 update instead. Closed.

PR #218, opened 2019-03-26
nvnet: Guest MAC address should not be initialized by hardware

LukeUsher:
This is set by the running application, so the default value remains unused.

See https://github.com/LukeUsher/Cxbx-Reloaded/issues/15


mborgerson:
I don't think initializing the MAC addr registers is negatively impacting anything since they are overwritten later on, but I would like to know that the register value is out of hw reset--we can look into that at a later time. It's correct that all nvnet init code (that I've tested anyhow) will program this register--this MAC init section follows the style of other ethernet hw drivers, but it's not really applicable in this case. 

Anyway, I'm ok with this change as we don't know what the value *actually should be* anyway (it probably doesn't matter really) and it simplifies the code a little bit. Thanks!

PR #219, opened 2019-04-24
build.sh: Disable GlusterFS support

thrimbor:
Building GlusterFS support has been broken for a while now (at least on Arch, error log [here](https://pastebin.com/qfxdgUkh)).

xqemu doesn't require it anyway, so this disables GlusterFS in build.sh.


mborgerson:
I'm not familiar with GlusterFS., but this seems like an error due to API mismatch. This should be addressed upstream, so let's just keep it off for now. Thanks!

PR #222, opened 2019-05-25
ci: Fix MSYS2 upgrade process for AppVeyor

### LINKED ISSUE #221: Building on Appveyor is broken
While testing something on my own fork I noticed that the build system for appveyor is broken. I've narrowed this down to an issue with doing a `sysupgrade` to the MSYS environment. This is because to do a `sysupgrade` you're required to close the msys shell and rerun the command after doing it once.

This problem has been there since the project began using appveyor, it just went unnoticed as everything is very verbose and only just now became a real problem as base packages are starting to update upstream.

Some solutions for this are:

1. manually updating the single broken package `libreadline`
(isn't the best option as we'll face the same issue in the future with other packages)
2. figuring out a way to relaunch the msys shell after the first `sysupgrade`
(likely hard)
3. just updating packages we deem necessary to be up-to-date
(likely the best option)

GXTX:
As stated in #221 the build script for appveyor wasn't completing the upgrade process for MSYS2 which caused a lib to become out of sync and not update. This PR fixes this by following the upgrade process recommended on MSYS2 website.

Fixes #221.

### DIFF
--- a/appveyor.bat
+++ b/appveyor.bat
@@ -9,8 +9,8 @@ rem Set the paths appropriately
 PATH C:\msys64\%MSYSTEM%\bin;C:\msys64\usr\bin;%PATH%
 
 rem Upgrade the MSYS2 platform
-bash -lc "pacman --noconfirm --sync --refresh --refresh pacman"
-bash -lc "pacman --noconfirm --sync --refresh --refresh --sysupgrade --sysupgrade"
+bash -lc "pacman --noconfirm -Syu"
+bash -lc "pacman --noconfirm -Su"
 
 rem Install required tools
 bash -xlc "pacman --noconfirm -S --needed base-devel"


JayFoxRox:
This seems to work - "better than master", so I think we should accept it, even though I don't understand it. This is also the recommended way of MSYS, so I'm fine with it.

No idea why we never did it like this in the first place.

---

**I believe the `master`-variant is:**

```
pacman -Syy pacman
pacman -Syyuu
```

* `pacman -Syy pacman` forces a package list update and then upgrades the package manager.
    * I believe this might be necessary sometimes where pacman  is unable to handle some new packages, so it must be handled with priority.
* `pacman -Syyuu` forces another package list update (why?) and then does a system upgrade, but also allowing downgrades (why?).
    * I assume the package list update is done because `man pacman` recommends it, probably to avoid pacman trying to fetch old packages which give 404; but in this use case might not be necessary

I don't understand what / why this caused issues on master? Was `libreadline` being downgraded?

**The new proposal is:**

```
pacman -Syu"
pacman -Su"
```

* `pacman -Syu` does a package list update (not forced if it thinks the list is up-to-date) and then upgrades all packages (without allowing downgrades)
    * I worry this might fail if the package format / keyring changes and old pacman doesn't know how to deal with it?
* `pacman -Su` forces another upgrade to all packages (without allowing downgrades) (why?).
    * I don't understand why this is done; maybe in case the first update only updated pacman

---

Eitherway, we can always refine it later - our current CI is broken, so this is still an improvement.

mborgerson:
Thanks for taking the time to look into this!

PR #224, opened 2019-05-27
ci: Fix compiling with macOS on travis-ci

GXTX:
macOS has been broken for some time due to updated packages, we should be using what travis-ci gives us via similar scripts as to what we currently use for linux with the same provider.

I also updated the python version to 3.7, however I'm not sure this actually effects anything as both versions build fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ os:
   - osx
 
 python:
-  - "2.6"
+  - "3.7"
 
 compiler:
   - gcc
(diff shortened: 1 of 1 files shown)


JayFoxRox:
Thanks!

This looks very good to me. Can be merged.

nit: "fix" &rarr; "Fix" / "macos" &rarr "macOS" in commit title etc.; same for other PR. Always try to match existing commit style (including capitalization)

mborgerson:
Cool, thanks for keeping our CI alive!

PR #231, opened 2019-05-29
ci: Stop updating the MSYS2 platform on AppVeyor

### LINKED ISSUE #221: Building on Appveyor is broken

GXTX:
#222 was supposed to fix #221 however it seems that during the time it took between the PR and merge MSYS2 updated *something* which broke building again. So from now on we wont update MSYS and let AppVeyor handle keeping packages up-to-date like we currently do with travis.

### DIFF
--- a/appveyor.bat
+++ b/appveyor.bat
@@ -8,10 +8,6 @@ rem    MSYSTEM:  MINGW64 or MINGW32
 rem Set the paths appropriately
 PATH C:\msys64\%MSYSTEM%\bin;C:\msys64\usr\bin;%PATH%
 
-rem Upgrade the MSYS2 platform
-bash -lc "pacman --noconfirm -Syu"
-bash -lc "pacman --noconfirm -Su"
-
 rem Install required tools
 bash -xlc "pacman --noconfirm -S --needed base-devel"
 


mborgerson:
I'm honestly not sure what's wrong with it (tbf I didn't try to debug...), but the build failed due to msys-readline8.dll not being found. It seems like an issue with the update process and possibly a conflict with built-in libs. Anyway, this solution gets the build working again, so I think it should be good for now. Thanks!

PR #234, opened 2019-05-30
eeprom: Implement writeback

thrimbor:
This is my own take on EEPROM writeback, based on the review comments on #122.
I see two remaining issues with my code:
1. It leaks the file descriptor (afaik this was also an issue in #122), because I couldn't spot a good place where I could put a `close()` (there seems to be no "destructor" for `MachineState`).
2. Saving the EEPROM is still done in a global function in xbox.c (criticized by @JayFoxRox). I intended to solve this by passing the file descriptor to the EEPROM device, but doing this properly would require work that is far over my head (this is the first time I'm touching qemu internals)

### DIFF
--- a/hw/i2c/smbus_eeprom.c
+++ b/hw/i2c/smbus_eeprom.c
@@ -26,6 +26,7 @@
 #include "hw/hw.h"
 #include "hw/i2c/i2c.h"
 #include "hw/i2c/smbus.h"
+#include "hw/xbox/xbox.h"
 
 //#define DEBUG
 
@@ -84,6 +85,8 @@ static void eeprom_write_data(SMBusDevice *dev, uint8_t cmd, uint8_t *buf, int l
     len -= n;
     if (len)
         memcpy(eeprom->data, buf + n, len);
+
+    save_eeprom(eeprom->data);
 }
 
 static uint8_t eeprom_read_data(SMBusDevice *dev, uint8_t cmd, int n)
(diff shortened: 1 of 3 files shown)


mborgerson:
Thanks for looking into this! I wish we could have talked before hand to share some thoughts I had on the matter; though tbf they probably should have been written down somewhere :).

Apparently, the direction the QEMU project wants to take with this is to introduce a blockdev backend that handles sync with the stored contents of the EEPROM. Eventually, to keep our lives easier, we will probably want to align with that approach. Even though it does feel like a *little* overkill for such a small amount of data, there are good reasons for doing it that way. I had planned on doing this and submitting the patch upstream. If you would like to take the helm of committing to doing this instead, and submitting the patch upstream, that would be good visibility for you!

The other, less burdensome (but unlikely to be adopted upstream) approach I had considered was tacking on a simple string property to the object and having the eeprom code do all of its own read/writing (this decouples it from our Xbox machine which as this PR is written--EEPROM breaks for every other machine). But again, I would prefer we align with upstream whenever possible, and it seems like they too want to solve this problem too, just in a way that fits their needs better, but which we also benefit from.

Naturally, others may disagree with my opinion here, so let's talk about it before acting on this PR.

mborgerson:
Closed for now as we have discussed in Discord to approach this another way with a follow up PR.

JayFoxRox:
There was no such discussion bridged to XQEMU IRC. Please relay.

PR #235, opened 2019-06-01
dsp: fix running on big-endian PowerPC

LukeUsher:
This change is enough to get the MS kernel to boot on big-endian PowerPC hardware. 

Execution reaches far enough for the nv2a emulation to hit glDrawRangeElements but unable to test further due to broken Mesa drivers. 

Cromwell does work.

### DIFF
--- a/hw/xbox/dsp/dsp_cpu.c
+++ b/hw/xbox/dsp/dsp_cpu.c
@@ -26,6 +26,8 @@
 #include <string.h>
 #include <assert.h>
 
+#include "qemu/osdep.h"
+#include "qemu/bswap.h"
 #include "dsp_cpu.h"
 
 #define TRACE_DSP_DISASM 0
@@ -957,7 +959,7 @@ static uint32_t read_memory_p(dsp_core_t* dsp, uint32_t address)
 {
     assert((address & 0xFF000000) == 0);
     assert(address < DSP_PRAM_SIZE);
-    uint32_t r = dsp->pram[address];
+    uint32_t r = ldl_le_p(&dsp->pram[address]);
     assert((r & 0xFF000000) == 0);
     return r;
 }
@@ -1019,7 +1021,7 @@ static void write_memory_raw(dsp_core_t* dsp, int space, uint32_t address, uint3
         dsp->yram[address] = value;
     } else if (space == DSP_SPACE_P) {
         assert(address < DSP_PRAM_SIZE);
-        dsp->pram[address] = value;
+        stl_le_p(&dsp->pram[address], value);
     } else {
         assert(false);
     }


hw/xbox/dsp/dsp_cpu.c
```diff
@@ -26,6 +26,8 @@
 #include <string.h>
 #include <assert.h>
 
+#include "qemu/osdep.h"
+#include "qemu/bswap.h"
```
JayFoxRox: We probably want our own endianess swap in this file, because it belongs to Hatari, not QEMU.

JayFoxRox:
nit: The commit author information is incorrect

hw/xbox/dsp/dsp_cpu.c
mborgerson: I think in general we should prefer to avoid having duplicate platform abstraction functions, this code lives in XQEMU after all. That said, it would be nice if upstream could benefit improvements we make (we want to be good open-source citizens)--in this case I would be okay with something like that.

hw/xbox/dsp/dsp_cpu.c
```diff
@@ -1019,7 +1021,7 @@ static void write_memory_raw(dsp_core_t* dsp, int space, uint32_t address, uint3
         dsp->yram[address] = value;
```
mborgerson: What about here too?

LukeUsher: It looked to me like these areas are only accessed by the dsp code as 32bit words and so are not effected by endianness issues: since it is never written or read by anything else. (But do correct me if I’m wrong) 

pram, however, is binary code transferred from the cpu, so we need to respect endianness here.

JayFoxRox: Generally P, X, Y - banks are all the same: 24 bit words.

The main means of accessing them with data is through MMIO regions (mapping directly to DSP RAM) or DMA (controlled by DSP code).

I'm not sure how the Hatari code accesses each of these banks. Our DSP DMA seems to use `dsp56k_read_memory` / `dsp56k_write_memory` to access each of these banks.

There's also scratch space, which isn't 24 bit words (and probably poorly accessed):

https://github.com/xqemu/xqemu/blob/787012e012af96d6d69ebac3f32a63356ede93c2/hw/xbox/dsp/dsp_dma.c#L271-L288

mborgerson:
@LukeUsher do you plan to address the remaining endianness issues with this code?

LukeUsher:
Eventually yes, but it's a little hard to test on my current PowerPC hardware, since it takes over 30 minutes for the first frame of the splash screen to show

mborgerson:
> it's a little hard to test on my current PowerPC hardware

I think we need not test on PPC hardware (though it would be nice). I just wanted to handle the cases where we were loading/storing to memory using the correct endianness handling functions.

For testing the changes, an alternative approach could also be to spin up a Linux machine on QEMU. I'm not sure what the performance would be like, but perhaps it would be a little faster.

mborgerson:
I have amended this commit to have correct author information based on previous commits and will merge once CI completes.

PR #243, opened 2019-06-28
Fix for building on macOS

### LINKED ISSUE #239: Error when building on macOS
Mojave v 10.14.5
```
[adhir@bigmac:~/src/xqemu] sh build.sh 
Compiling for MacOS…
readlink: illegal option -- f
usage: readlink [-n] [file ...]
+ ./configure '--extra-cflags=-DXBOX=1 -O0 -g -march=native ' --enable-debug --disable-cocoa --target-list=i386-softmmu --enable-trace-backends=nop --enable-sdl --with-sdlabi=2.0 --disable-curl --disable-vnc --disable-vnc-sasl --disable-docs --disable-tools --disable-guest-agent --disable-tpm --disable-live-block-migration --disable-rdma --disable-replication --disable-capstone --disable-fdt --disable-libiscsi --disable-spice --disable-user --disable-stack-protector --disable-glusterfs --disable-bluez --disable-gtk --disable-curses --disable-gnutls --disable-nettle --disable-gcrypt --disable-crypto-afalg --disable-virglrenderer --disable-vhost-net --disable-vhost-crypto --disable-vhost-vsock --disable-vhost-user --disable-virtfs --disable-libssh2 --disable-snappy --disable-bzip2 --disable-vde --disable-libxml2 --disable-seccomp --disable-numa --disable-lzo --disable-smartcard --disable-usb-redir

ERROR: glib-2.40 gobject-2.0 is required to compile QEMU
```

thrimbor:
The problems we currently experience when building on macOS are caused by pkg-config not finding `libffi.pc`, which is a requirement for `pkg-config` to find `gobject-2.0`. libffi is not symlinked into `/usr/local` by homebrew on purpose, so we must add the path to `libffi.pc` to the `PKG_CONFIG_PATH` environment variable (this is the solution recommended by homebrew when installing libffi).

It seems that XQEMU doesn't actually require libffi when building, otherwise we'd have to add `/usr/local/opt/libffi/lib` to the libary search directories as well.

Building tested on macOS, building and running tested on Arch.

Closes #239.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ addons:
       - libsdl2-dev
       - libepoxy-dev
   homebrew:
+    update: true
     packages:
       - glib
       - pixman
--- a/build.sh
+++ b/build.sh
@@ -48,6 +48,8 @@ case "$(uname -s)" in # adjust compilation option based on platform
         echo 'Compiling for MacOS…'
         sys_cflags='-march=native'
         sys_opts='--disable-cocoa'
+        # necessary to find libffi, which is required by gobject
+        export PKG_CONFIG_PATH="${PKG_CONFIG_PATH}/usr/local/opt/libffi/lib/pkgconfig"
         ;;
     CYGWIN*|MINGW*|MSYS*)
         echo 'Compiling for Windows…'


JayFoxRox:
Thanks for looking into this. Are there some docs by homebrew developers about this we could link from this discussion?

Also please use prefix style for your commits.

thrimbor:
> Are there some docs by homebrew developers about this we could link from this discussion?

I'm only aware of https://docs.brew.sh/FAQ#what-does-keg-only-mean, and the place where it's specified is here: https://github.com/Homebrew/homebrew-core/blob/master/Formula/libffi.rb#L24

Issue #192, opened 2019-02-03
EP DSP emulation throws "0x0c1e84 Illegal instruction" which is "lsl #ii, D"


JayFoxRox:
Instruction is missing for EP emulation, so it only happens when enabling this line:

See https://github.com/JayFoxRox/xqemu-espes/pull/24/commits/02e90ee4bd445f245ec151b14bdb1c01a15ae9d6 for an old experimental commit I made.

This needs to be tested on hardware and then submitted (possibly with minor style changes, such as lower-case commit prefix).

Possibly interacts with #191 if we use available helpers (although lsl / asl is less likely to be affected).

[For datasheet and tool links check XboxDevWiki](http://xboxdevwiki.net/DSP).

Issue #161, opened 2018-12-26
DSP DMA control bit 4 ignored


JayFoxRox:
We ignore unknown control bit 4, needs research.

JayFoxRox:
Looks like it's related to circular buffers. It seems to be used for buffer 0xE (what we assume to be "circular" scratch-space).

I think this might modify the DMA block to writeback the cursor, because despite being a circular buffer, the offset remains at 0 in our emulation.

Issue #160, opened 2018-12-26
DSP DMA control bit 0 ignored


JayFoxRox:
We ignore unknown control bit 0, needs research.

Issue #151, opened 2018-12-26
APU GP DSP isn't running enough cycles per frame


JayFoxRox:
The APU GP DSP only runs 1000 cycles per APU frame (0.666...ms) in current master revision.

The GP on hardware probably runs at 160MHz, so it should run up to 0.666...ms*(160MHz) ~ 106666.56 cycles per frame.

Desynchronization of APU frames / DSP code can lead to unpredictable results, so if we run another APU frame before the previous DSP frame has finished, it can break emulation.

---

We can probably get away with less than 106k cycles. To check if the DSP has reached its goal, we can look for the following instruction, which is used in the DirectSound GP DSP code:

```
p:000b 0a8581 00000a (02 cyc) jclr #1,x:$ffffc5,p:$000a
```

This instruction waits for a new frame start, and requires APU intervention - meaning the DSP will be stuck there. We can finish DSP emulation for the current frame at that point.

JayFoxRox:
To show that this issue is valid, I have hooked the instruction I mentioned above by adding the following `printf` code; I'm documenting this here for the future, and for people not experienced with the DSP (which might struggle to find the relevant DSP code).

This is in hw/xbox/dsp/dsp_emu.inl:

```c
static void emu_jclr_pp(dsp_core_t* dsp)
{
    uint32_t memspace, addr, value, numbit, newaddr;
    
    memspace = (dsp->cur_inst>>6) & 1;
    value = (dsp->cur_inst>>8) & BITMASK(6);
    numbit = dsp->cur_inst & BITMASK(5);
    addr = 0xffffc0 + value;
    value = dsp56k_read_memory(dsp, memspace, addr);
    newaddr = read_memory_p(dsp, dsp->pc+1);

    dsp->instr_cycle += 4;

    // Add this
    printf("                      PC at 0x%X, wants to jump to 0x%X after checking %d, 0x%X\n", dsp->pc, newaddr, numbit, addr);

    if ((value & (1<<numbit))==0) {
        dsp->pc = newaddr;
        dsp->cur_inst_len = 0;
        return;
    } 
    ++dsp->cur_inst_len;
}
```

You also want to disable the MCPX "frame ping" message in hw/xbox/mcpx_apu.c.
The frame ping marks a frame start, and the DSP message signals that the frame has finished (= waiting for new frame).
If you don't see the DSP message before a new "frame ping", the emulation can break in various ways.

---

[...]

PR #251, opened 2019-07-12
Audio PR #3: Connect VP output directly into system audio

mborgerson:
Note: This PR is part of a series of PRs coming to add basic audio.

---

This patch gets *very* basic audio working in XQEMU by feeding the
output of the VP directly into system audio through the QEMU sound
architecture.

Note: the audio coming out of the VP is not fully mixed, and this patch
mixes all voices down to a single channel to keep things simple for now.
Currently, only buffers are played, streamed audio is not yet supported.
This may cause some games to hang. Furthermore, the DSP is not involved
here yet, so in addition to poor mixing, some audio may be horribly
broken.

Processing of individual voices is turned off in this patch to avoid regressions with games
until I follow up with the streaming PR. To test, set the `PROCESS_VOICES`
macro in mcpx_apu.c to 1.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -42,19 +42,19 @@ case "$(uname -s)" in # adjust compilation option based on platform
     Linux)
         echo 'Compiling for Linux…'
         sys_cflags='-march=native -Wno-error=redundant-decls -Wno-error=unused-but-set-variable'
-        sys_opts='--enable-kvm --disable-xen --disable-werror'
+        sys_opts='--enable-kvm --disable-xen --disable-werror --audio-drv-list=alsa'
         ;;
     Darwin)
         echo 'Compiling for MacOS…'
         sys_cflags='-march=native'
-        sys_opts='--disable-cocoa'
+        sys_opts='--disable-cocoa --audio-drv-list=coreaudio'
         # necessary to find libffi, which is required by gobject
         export PKG_CONFIG_PATH="${PKG_CONFIG_PATH}/usr/local/opt/libffi/lib/pkgconfig"
         ;;
     CYGWIN*|MINGW*|MSYS*)
         echo 'Compiling for Windows…'
         sys_cflags='-Wno-error'
-        sys_opts='--python=python3 --disable-cocoa --disable-opengl'
+        sys_opts='--python=python3 --disable-cocoa --disable-opengl --audio-drv-list=dsound'
         postbuild='package_windows' # set the above function to be called after build
         ;;
     *)
--- a/hw/xbox/mcpx_apu.c
+++ b/hw/xbox/mcpx_apu.c
@@ -25,10 +25,12 @@
 #include "hw/pci/pci.h"
 #include "cpu.h"
 #include "hw/xbox/dsp/dsp.h"
+#include "audio/audio.h"
+#include "qemu/fifo8.h"
 #include <math.h>
 
 #define USE_APU_THREAD 1
-#define PROCESS_VOICES 0
+#define PROCESS_VOICES 1
 
 #define NUM_SAMPLES_PER_FRAME 32
 #define NUM_MIXBINS 32
@@ -319,6 +321,11 @@ typedef struct MCPXAPUState {
     PCIDevice dev;
     bool exiting;
     bool set_irq;
+
+    QEMUSoundCard qemu_audio_card;
+    SWVoiceOut *qemu_audio_voice_out;
+    Fifo8 vp_out_buf;
+    QemuSpin vp_out_buf_lock;
 
     MemoryRegion *ram;
     uint8_t *ram_ptr;
@@ -1435,6 +1442,7 @@ static void process_voice(MCPXAPUState *d,
     }
 
     //FIXME: Decode voice volume and bins
+#if 0
     int bin[8] = {
       voice_get_mask(d, v, NV_PAVS_VOICE_CFG_VBIN, NV_PAVS_VOICE_CFG_VBIN_V0BIN),
       voice_get_mask(d, v, NV_PAVS_VOICE_CFG_VBIN, NV_PAVS_VOICE_CFG_VBIN_V1BIN),
@@ -1445,6 +1453,7 @@ static void process_voice(MCPXAPUState *d,
       voice_get_mask(d, v, NV_PAVS_VOICE_CFG_FMT, NV_PAVS_VOICE_CFG_FMT_V6BIN),
       voice_get_mask(d, v, NV_PAVS_VOICE_CFG_FMT, NV_PAVS_VOICE_CFG_FMT_V7BIN)
     };
+#endif
     uint16_t vol[8] = {
       voice_get_mask(d, v, NV_PAVS_VOICE_TAR_VOLA, NV_PAVS_VOICE_TAR_VOLA_VOLUME0),
       voice_get_mask(d, v, NV_PAVS_VOICE_TAR_VOLA, NV_PAVS_VOICE_TAR_VOLA_VOLUME1),
@@ -1468,6 +1477,7 @@ static void process_voice(MCPXAPUState *d,
       voice_set_mask(d, v, NV_PAVS_VOICE_PAR_OFFSET, NV_PAVS_VOICE_PAR_OFFSET_CBO, cbo);
     }
 
+#if 0
     // Apply the amplitude envelope
     //FIXME: Figure out when exactly and how exactly this is actually done
     for(unsigned int j = 0; j < channels; j++) {
@@ -1475,10 +1485,13 @@ static void process_voice(MCPXAPUState *d,
             samples[j][i] *= ea_value;
         }
     }
+#endif
 
     //FIXME: If phase negations means to flip the signal upside down
     //       we should modify volume of bin6 and bin7 here.
 
+
+#if 0
     // Mix samples into voice bins
     for(unsigned int j = 0; j < 8; j++) {
         MCPX_DPRINTF("Adding voice 0x%04X to bin %d [Rate %.2f, Volume 0x%03X] sample %d at %d [%.2fs]\n", v, bin[j], rate, vol[j], samples[0], cbo, cbo / (rate * 48000.0f));
@@ -1487,6 +1500,22 @@ static void process_voice(MCPXAPUState *d,
             //FIXME: What happens to the other channel? Is this behaviour correct?
             mixbins[bin[j]][i] += (0xFFF - vol[j]) * samples[j % channels][i] / 0xFFF;
         }
+    }
+#endif
+
+    /*
+     * FIXME: For now, simply mix samples into a single bin for basic playback
+     */
+
+    // Find the bin with maximal volume and just use that to scale volume
+    int mix_vol = 0xfff;
+    for (int j = 0; j < 8; j++) {
+        mix_vol = MIN(mix_vol,vol[j]);
+    }
+
+    for (int i = 0; i < 0x20; i++) {
+        mixbins[0][i] += ((0xfff - mix_vol) * (samples[0][i] / 0xfff) / 4);
+        // mixbins[0][i] += ((0xfff - mix_vol) * (samples[1][i] / 0xfff) / 4) / 2;
     }
 #endif
 }
@@ -1503,6 +1532,14 @@ static void se_frame(void *opaque)
     timer_mod(d->se.frame_timer, qemu_clock_get_us(QEMU_CLOCK_VIRTUAL) + 600);
 #endif
     MCPX_DPRINTF("mcpx frame ping\n");
+
+    // Check for free space in the output buffer
+    qemu_spin_lock(&d->vp_out_buf_lock);
+    int avail = fifo8_num_free(&d->vp_out_buf);
+    qemu_spin_unlock(&d->vp_out_buf_lock);
+    if (avail < (NUM_SAMPLES_PER_FRAME * 4)) {
+        return;
+    }
 
     /* Buffer for all mixbins for this frame */
     int32_t mixbins[NUM_MIXBINS][NUM_SAMPLES_PER_FRAME] = { 0 };
@@ -1560,6 +1597,11 @@ static void se_frame(void *opaque)
     }
 #endif
 
+    // Temporarily provide samples just from mixbin 0
+    qemu_spin_lock(&d->vp_out_buf_lock);
+    fifo8_push_all(&d->vp_out_buf, (uint8_t*)&mixbins[0][0], sizeof(mixbins[0]));
+    qemu_spin_unlock(&d->vp_out_buf_lock);
+
     /* Write VP results to the GP DSP MIXBUF */
     for (mixbin = 0; mixbin < NUM_MIXBINS; mixbin++) {
         for (sample = 0; sample < NUM_SAMPLES_PER_FRAME; sample++) {
@@ -1583,6 +1625,32 @@ static void se_frame(void *opaque)
 
         // hax
         // dsp_run(d->ep.dsp, 1000);
+    }
+}
+
+static void mcpx_vp_out_cb(void *opaque, int free_b)
+{
+    MCPXAPUState *s = MCPX_APU_DEVICE(opaque);
+    if (free_b <= 0) return;
+
+    int avail = fifo8_num_used(&s->vp_out_buf);
+    int temp = audio_MIN(free_b, avail);
+
+    while (temp > 0) {
+        // Get samples from output buffer
+        qemu_spin_lock(&s->vp_out_buf_lock);
+        uint32_t to_copy;
+        const uint8_t *samples = fifo8_pop_buf(&s->vp_out_buf, temp, &to_copy);
+
+        // Provide samples to audio card
+        int copied = AUD_write(s->qemu_audio_voice_out, (void*)samples, to_copy);
+
+        qemu_spin_unlock(&s->vp_out_buf_lock);
+
+        if (!copied) {
+            break;
+        }
+        temp -= copied;
     }
 }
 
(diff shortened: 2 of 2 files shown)


build.sh
```diff
@@ -42,18 +42,18 @@ case "$(uname -s)" in # adjust compilation option based on platform
     Linux)
         echo 'Compiling for Linux…'
         sys_cflags='-march=native -Wno-error=redundant-decls -Wno-error=unused-but-set-variable'
-        sys_opts='--enable-kvm --disable-xen --disable-werror'
+        sys_opts='--enable-kvm --disable-xen --disable-werror --audio-drv-list=alsa'
         ;;
     Darwin)
         echo 'Compiling for MacOS…'
-        sys_cflags='-march=native'
+        sys_cflags='-march=native --audio-drv-list=coreaudio'
```
thrimbor: `--audio-drv-list=coreaudio` needs to go into `sys_opts` instead of `sys_cflags` (same for Windows).

mborgerson: Duh, thx

hw/xbox/mcpx_apu.c
```diff
@@ -1394,13 +1422,13 @@ static void process_voice(MCPXAPUState *d,
...
-                    samples[channel][i] = ldub_phys(&address_space_memory, addr);
+                    samples[channel][i] = ldub_phys(&address_space_memory, addr) << 24;
                     break;
                 case NV_PAVS_VOICE_CFG_FMT_SAMPLE_SIZE_S16:
-                    samples[channel][i] = (int16_t)lduw_le_phys(&address_space_memory, addr);
+                    samples[channel][i] = (int16_t)lduw_le_phys(&address_space_memory, addr) << 16;
                     break;
                 case NV_PAVS_VOICE_CFG_FMT_SAMPLE_SIZE_S24:
-                    samples[channel][i] = (int32_t)(ldl_le_phys(&address_space_memory, addr) << 8) >> 8;
+                    samples[channel][i] = (int32_t)(ldl_le_phys(&address_space_memory, addr) ) << 8;
```
JayFoxRox: I assume I've used 24 bit here, because that's what hardware tests suggested. The DSP also receives 24 bit samples - I'm not sure how it expects them to be packaged.. but I assume in the lower 24 bit.

JayFoxRox:
There's https://github.com/JayFoxRox/xqemu/pull/25 and https://github.com/JayFoxRox/xqemu/pull/28 which do a lot of the same (except they are lacking the FIFO). 

Why are we even considering a different implementation now, that wasn't tested against hardware? 

This lacks GP / EP (HLE) functionality like ITU downmix, but instead does naive channel picking and combines it with hardware functionality (VP mixing).

hw/xbox/mcpx_apu.c
```diff
@@ -1612,6 +1721,21 @@ static void mcpx_apu_register(void)
...
 
+#if USE_APU_THREAD
+static void *mcpx_apu_frame_thread(void *arg)
+{
+    volatile MCPXAPUState *d = MCPX_APU_DEVICE(arg);
+
+    while (!d->exiting) {
+        int should_run = (d->regs[NV_PAPU_SECTL] & NV_PAPU_SECTL_XCNTMODE) >> 3;
+        if (should_run == NV_PAPU_SECTL_XCNTMODE_OFF) continue; // FIXME: use a proper sync method
+        se_frame((void*)d);
```
JayFoxRox: There should at least be a warning somewhere in `se_frame`, that this runs at the wrong frequency (host dependent, because it depends on the FIFO being empty).

hw/xbox/mcpx_apu.c
```diff
@@ -1424,6 +1453,7 @@ static void process_voice(MCPXAPUState *d,
       voice_get_mask(d, v, NV_PAVS_VOICE_CFG_FMT, NV_PAVS_VOICE_CFG_FMT_V6BIN),
       voice_get_mask(d, v, NV_PAVS_VOICE_CFG_FMT, NV_PAVS_VOICE_CFG_FMT_V7BIN)
     };
+#endif
```
JayFoxRox: Why is this being disabled? As far as I remember, this was confirmed against hardware.

hw/xbox/mcpx_apu.c
```diff
@@ -1447,17 +1477,21 @@ static void process_voice(MCPXAPUState *d,
...
 
+#if 0
     // Apply the amplitude envelope
     //FIXME: Figure out when exactly and how exactly this is actually done
     for(unsigned int j = 0; j < channels; j++) {
         for(unsigned int i = 0; i < 0x20; i++) {
             samples[j][i] *= ea_value;
         }
     }
+#endif
```
JayFoxRox: Why is this being disabled? While the exact place isn't known (especially for feedback-loops, 3D audio, effects, ..) the code itself should be fine.

If you want to disable it, you can simply stub the envelope.

hw/xbox/mcpx_apu.c
```diff
@@ -1468,6 +1502,22 @@ static void process_voice(MCPXAPUState *d,
...
+
+    /*
+     * FIXME: For now, simply mix samples into a single bin for basic playback
+     */
+
+    // Find the bin with maximal volume and just use that to scale volume
+    int mix_vol = 0xfff;
+    for (int j = 0; j < 8; j++) {
+        mix_vol = MIN(mix_vol,vol[j]);
+    }
```
JayFoxRox: This makes no sense. Why is this being done?

hw/xbox/mcpx_apu.c
```diff
@@ -1620,4 +1744,34 @@ void mcpx_apu_init(PCIBus *bus, int devfn, MemoryRegion *ram)
...
+    struct audsettings as = {
+        .freq = 48000,
+        .nchannels = 1,
+        .fmt = AUDIO_FORMAT_S32,
+        .endianness = 0,
+    };
+
+    d->qemu_audio_voice_out = AUD_open_out(&d->qemu_audio_card,
+                                           d->qemu_audio_voice_out,
+            "XboxAPU.out", d, mcpx_vp_out_cb, &as);
```
JayFoxRox: This name is not QEMU style (as far as I remember).
The QEMU names appear to be lowercase.

I chose "mcpx-apu.vp_bins" in my code for similar bins. It should also be clearly marked as debug output, as no such output exists in the hardware. APU is an audio processor, not capable of driving audio output directly (that's what the ACI does).

hw/xbox/mcpx_apu.c
```diff
@@ -24,8 +25,13 @@
 #include "hw/pci/pci.h"
 #include "cpu.h"
 #include "hw/xbox/dsp/dsp.h"
+#include "audio/audio.h"
+#include "qemu/fifo8.h"
 #include <math.h>
 
+#define USE_APU_THREAD 1
```
JayFoxRox: Why is this an option?

We should do threading, or we shouldn't do it - I don't think there's a benefit of having it be an option; much less a compile-time option.

Giving the option just increases complexity and makes it harder to trace problems (if someone decides to mess with this).

hw/xbox/mcpx_apu.c
```diff
@@ -1511,6 +1571,17 @@ static void se_frame(void *opaque)
...
+    if (d->set_irq) {
+#if USE_APU_THREAD
+        qemu_mutex_lock_iothread();
+        update_irq(d);
+        qemu_mutex_unlock_iothread();
+#else
+        update_irq(d);
+#endif
+        d->set_irq = false;
+    }
```
JayFoxRox: Shouldn't this happen after the debug `GENERATE_MIXBIN_BEEP` to ensure (and hint for other developers) that there are valid samples when the IRQ handler runs?

hw/xbox/mcpx_apu.c
```diff
@@ -1526,6 +1597,11 @@ static void se_frame(void *opaque)
     }
 #endif
 
+    // Temporarily provide samples just from mixbin 0
+    qemu_spin_lock(&d->vp_out_buf_lock);
+    fifo8_push_all(&d->vp_out_buf, (uint8_t*)&mixbins[0][0], sizeof(mixbins[0]));
```
JayFoxRox: Why not provide all samples here, so the recipient can run a proper GP / EP / ACI equivalent mixing?

hw/xbox/mcpx_apu.c
```diff
@@ -1478,9 +1528,19 @@ static void se_frame(void *opaque)
...
     MCPX_DPRINTF("mcpx frame ping\n");
 
+    // Check for free space in the output buffer
+    qemu_spin_lock(&d->vp_out_buf_lock);
+    int avail = fifo8_num_free(&d->vp_out_buf);
+    qemu_spin_unlock(&d->vp_out_buf_lock);
+    if (avail < (NUM_SAMPLES_PER_FRAME * 4)) {
+        return;
+    }
+
```
JayFoxRox: We probably want a debug marker here, to ensure that this code path is taken at 1500Hz in the virtual-time.

It requires a comment (or an extension to what's already in line 1523), because the logic implies it will be running at 1500Hz in host-time... which is wrong / leads to problems.

It's not even 1500Hz in host-time, but only average, as there is variable frequency, depending on how large the host audio driver buffer sizes are - so it's possible that it will suddenly consume 20 frames at once, then nothing for a while.
Meanwhile, the virtual-time will still advance, leading to timing issues and various forms of desynchronization (VP / GP, CPU / GP, GP / GP, GP / CPU, GP / EP, EP / EP, EP / ACI - and maybe others I didn't even think of).

The proper solution is to run it at 1500Hz in virtual-time, and [stretch the frames](https://en.wikipedia.org/wiki/Audio_time_stretching_and_pitch_scaling) to 1500Hz in real-time.

Also see discussion in https://github.com/xqemu/xqemu/pull/250#discussion_r303221485

dracc:
I've tested this PR, gave me some audio (:heartbeat:) and then I hit an assertion due to streaming not being implemented (as expected).

hw/xbox/mcpx_apu.c
mborgerson: This comment should be on Audio PR#2 as this branch is based on it, but I'll respond here.

This is a holdover from my development so I could revert back quickly to test if there was an issue with the independent thread. I left it in so others could also help test the changes, but the threading approach is the way to go and it's very stable after my local testing over the past couple months. I'm going to clean up Audio PR#2 and remove timer model altogether.

hw/xbox/mcpx_apu.c
mborgerson: `se_frame` will run at ~1500Hz, give or take a little. This is throttled by the system audio sink. Agree, that could be more clear here, so I'll add a comment about it in Audio PR#2.

hw/xbox/mcpx_apu.c
mborgerson: `mcpx-apu.vp_bins` is better, I'll use that. Thanks!

hw/xbox/mcpx_apu.c
mborgerson: I mixed all samples into mixbin 0 here, so they are provided but this code path should be cleaned up

hw/xbox/mcpx_apu.c
mborgerson: Agree, we should stick to 24b here. I scaled them up here for 32-bit system audio output, but we should do that instead when we mix down for the VP audio sink.

hw/xbox/mcpx_apu.c
mborgerson: This was disabled as I disabled mixing to bins below. It should be re-enabled however before this gets merged

hw/xbox/mcpx_apu.c
mborgerson: If I recall correctly, there seemed to be an issue with this amplitude scaling. We should probably understand this better before we apply it to the samples. Agree we can stub in envelope generator.

hw/xbox/mcpx_apu.c
mborgerson: Done here because I'm mixing everything into one bin

hw/xbox/mcpx_apu.c
mborgerson: Agree, will add a comment about how this throttles to ~1500Hz due to FIFO. At the moment, I'm unsure of how we should handle running at 1500Hz in virtual-time, but it probably doesn't need to be so precise. Once we set up the pipeline to output audio from EP, I think we may not need to worry too much about desync

hw/xbox/mcpx_apu.c
mborgerson: To be honest I'm not sure exactly when the hardware will assert an IRQ--we need to do some more research there. This is added here to handle stream notifications. I think whether after or before the mixbin beep doesn't matter too much as the samples stored in the temporary buffer will still be the same. I can move it down though as it looks a little cleaner.

mborgerson:
> There's JayFoxRox#25 and JayFoxRox#28 which do a lot of the same (except they are lacking the FIFO).

Ideally, 25 and 28 should have been PR'd upstream to avoid this duplication of work. Nevertheless, we can trivially integrate the best of all three of these branches.

> Why are we even considering a different implementation now, that wasn't tested against hardware?

How shall we test dumping output from the VP? Is it even necessary? This approach works well in the interim until we get the full GP-EP pipeline working.

> This lacks GP / EP (HLE) functionality like ITU downmix, but instead does naive channel picking and combines it with hardware functionality (VP mixing).

Correct. It's sub-optimal, but works well enough for now. Let us not be hindered by perfectionism here, we can improve this as we go.

hw/xbox/mcpx_apu.c
JayFoxRox: Did you try my mixer (disabled code right above = `// Mix samples into voice bins `) and:

- accumulating the results of the 32 mixbins (at 1/32 of the volume; to get a hacky downmix - will contain multi-pass garbage though)
- or: assume a ITU downmix of the respective mixbins (basically what default GP / EP would do if I remember correctly)

For reference (ITU downmix approach):

[This code should be doing readback of the data that GP sends to EP (which is post-processed VP mixbins)](https://github.com/JayFoxRox/xqemu/blob/c14829135269952d81729d8d30d54b18d9f3c87c/hw/xbox/mcpx_apu.c#L882-L889).
I don't fully remember, but [I believe this code was also based on RE and does what EP does](https://github.com/JayFoxRox/xqemu/blob/c14829135269952d81729d8d30d54b18d9f3c87c/hw/xbox/mcpx_apu.c#L912-L933).

This kind of emulation basically skips GP effects processing and EP AC3 encoding, but still mimics the 5.1 to stereo conversion and VP mixbin layout suggested by DirectSound.

hw/xbox/mcpx_apu.c
JayFoxRox: I worry a lot about de-synchronization. There's many things which can go wrong (VP = GP = EP = ACI = Host Audio timing). If any component is too slow (or fast) this will degrade audio quality and potentially affect game logic (GP is in direct contact with CPU and they potentially run in lockstep).

We likely can't guarantee that Host Audio has the same behavior as ACI should have, so we also can't guarantee anything for components feeding ACI (VP / GP / EP). However, at least GP frame timing would potentially affect the CPU = game logic suddenly depends on Host Audio.
This is something we'd want to avoid - Host Audio timing (and effects on logic) is a very prominent issue in most modern emulators.

The easiest solution is to decouple host and guest timing by using time-stretching. That guarantees the same logic for all users, independent from their audio drivers.

It also avoids ugly hacks.

*(Time stretching can be done later of course; just wanted to elaborate as to why I worry about this)*

> Once we set up the pipeline to output audio from EP

Debug audio output from EP or GP doesn't solve the timing issues: you need isochronous 1500Hz, but if EP is hooked up to Host Audio, then you might get something different (Remember VP = GP = EP = ACI = Host Audio timing).

EP outputting audio is also only a temporary hack until we have good ACI.

JayFoxRox:
> Ideally, 25 and 28 should have been PR'd upstream to avoid this duplication of work.

I had discussed them in public before, and they were still work-in-progress (also indicated by labels, which I'm going to remove soon).

>> Why are we even considering a different implementation now, that wasn't tested against hardware?
>
> How shall we test dumping output from the VP? Is it even necessary?

I mostly meant the underlying hacks to make output possible (see review comments: disabling certain features, changing from 24 bits to 32 bits, timing is changed from actual hardware values to random values, picking samples into a single mixbin).

\- Hacking audio output into the codebase should not modify the audio generation itself.

If code is necessary to make VP output to speakers work, then it should happen outside of the code which emulates the actual hardware (so it should only read the finished MIXBINs).
If code is causing bad MIXBIN output, it should be removed; marked as broken / disabled; or the bugs should be tolerated temporarily. Replacing it with hacks, or modifying values which were not confirmed on hardware, should not be an option.

Doing anything else makes it hard to understand how the APU actually works and will also mess up GP inputs (making any further work very hard).


The proposed VP routing also has no relation to how the hardware works, but isn't marked as debug output (so readers get confused about what the APU is; reminder: APU has no audio output, that's ACIs job - any APU audio output is a hack for debugging).

Routing VP MIXBIN directly to speakers just isn't a good idea. At least it should be kept separate from the code that feeds the MIXBINs and use the algorithms that GP / EP would use (based on RE and hardware testing, so it sounds okish) - see linked branches / comments.

Routing VP MIXBIN to disk is a better idea, so we can continue to fix VP by comparing against hardware, [using tools like this](https://github.com/JayFoxRox/xbox-tools/blob/master/python-scripts/trace_apu_mixbuf.py).

> This approach works well in the interim until we get the full GP-EP pipeline working.

VP to GP, and GP to EP should already be working I think?
I definitely upstreamed a lot of the DMA transfer stuff; so I'd assume that the transfers work except for minor issues (#160 / #161).

My debug output branch has had 2-3 options I believe:
- Dumping VP output
- Dumping GP output (sniffing "GP to EP" DMA source)
- Dumping EP/ACI output (sniffing "EP to RAM (ACI buffer)" DMA source or destination)

The EP output isn't interesting because it's only encoding, and EP is blocked by #192. GP output should work (no DMA involved) but is probably affected by #151 for a lot of code. - I tested these things in isolation (without full VP), so I only ever got the beep working.

>> This lacks GP / EP (HLE) functionality like ITU downmix, but instead does naive channel picking and combines it with hardware functionality (VP mixing).
>
> Correct. It's sub-optimal, but works well enough for now. Let us not be hindered by perfectionism here, we can improve this as we go.

Fair enough - but again: hacks should be marked as such, and kept to a minimum, so integration of existing routing code will be easier.

PR #253, opened 2019-09-03
smbus: Add feature to specify video output type

binarymaster:
This adds `avpack` parameter with options `scart`, `svideo`, `vga`, `rfu`, `hdtv`, `none`, and `composite` (by default).

Also improved constant definitions based on https://xboxdevwiki.net/AV_Cables.

### DIFF
--- a/hw/xbox/smbus_xbox_smc.c
+++ b/hw/xbox/smbus_xbox_smc.c
@@ -28,6 +28,7 @@
 #include "hw/i2c/i2c.h"
 #include "hw/i2c/smbus_slave.h"
 #include "qemu/config-file.h"
+#include "qapi/error.h"
 #include "sysemu/sysemu.h"
 #include "smbus.h"
 
@@ -55,10 +56,11 @@
 #define SMC_REG_AVPACK              0x04
 #define     SMC_REG_AVPACK_SCART        0x00
 #define     SMC_REG_AVPACK_HDTV         0x01
-#define     SMC_REG_AVPACK_VGA_SOG      0x02
+#define     SMC_REG_AVPACK_VGA          0x02
+#define     SMC_REG_AVPACK_RFU          0x03
 #define     SMC_REG_AVPACK_SVIDEO       0x04
 #define     SMC_REG_AVPACK_COMPOSITE    0x06
-#define     SMC_REG_AVPACK_VGA          0x07
+#define     SMC_REG_AVPACK_NONE         0x07
 #define SMC_REG_FANMODE             0x05
 #define SMC_REG_FANSPEED            0x06
 #define SMC_REG_LEDMODE             0x07
@@ -86,6 +88,7 @@ typedef struct SMBusSMCDevice {
     SMBusDevice smbusdev;
     int version_string_index;
     uint8_t cmd;
+    uint8_t avpack_reg;
     uint8_t scratch_reg;
 } SMBusSMCDevice;
 
@@ -154,8 +157,7 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
             smc->version_string_index++ % (sizeof(smc_version_string) - 1)];
 
     case SMC_REG_AVPACK:
-        /* pretend to have a composite av pack plugged in */
-        return SMC_REG_AVPACK_COMPOSITE;
+        return smc->avpack_reg;
 
     case SMC_REG_SCRATCH:
         return smc->scratch_reg;
@@ -178,16 +180,61 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
     return 0;
 }
 
+bool xbox_smc_avpack_to_reg(const char *avpack, uint8_t *value)
+{
+    uint8_t r;
+
+    if (!strcmp(avpack, "composite")) {
+        r = SMC_REG_AVPACK_COMPOSITE;
+    } else if (!strcmp(avpack, "scart")) {
+        r = SMC_REG_AVPACK_SCART;
+    } else if (!strcmp(avpack, "svideo")) {
+        r = SMC_REG_AVPACK_SVIDEO;
+    } else if (!strcmp(avpack, "vga")) {
+        r = SMC_REG_AVPACK_VGA;
+    } else if (!strcmp(avpack, "rfu")) {
+        r = SMC_REG_AVPACK_RFU;
+    } else if (!strcmp(avpack, "hdtv")) {
+        r = SMC_REG_AVPACK_HDTV;
+    } else if (!strcmp(avpack, "none")) {
+        r = SMC_REG_AVPACK_NONE;
+    } else {
+        return false;
+    }
+
+    if (value) {
+        *value = r;
+    }
+    return true;
+}
+
+void xbox_smc_append_avpack_hint(Error **errp)
+{
+    error_append_hint(errp, "Valid options are: composite (default), scart, svideo, vga, rfu, hdtv, none\n");
+}
+
 static void smbus_smc_realize(DeviceState *dev, Error **errp)
 {
     SMBusSMCDevice *smc = XBOX_SMC(dev);
+    char *avpack;
 
     smc->version_string_index = 0;
+    smc->avpack_reg = 0; /* Default value for Chihiro machine */
     smc->scratch_reg = 0;
     smc->cmd = 0;
 
     if (object_property_get_bool(qdev_get_machine(), "short-animation", NULL)) {
         smc->scratch_reg = SMC_REG_SCRATCH_SHORT_ANIMATION;
+    }
+
+    avpack = object_property_get_str(qdev_get_machine(), "avpack", NULL);
+    if (avpack) {
+        if (!xbox_smc_avpack_to_reg(avpack, &smc->avpack_reg)) {
+            error_setg(errp, "Unsupported avpack option '%s'", avpack);
+            xbox_smc_append_avpack_hint(errp);
+        }
+
+        g_free(avpack);
     }
 }
 
--- a/hw/xbox/xbox.c
+++ b/hw/xbox/xbox.c
@@ -476,6 +476,28 @@ static void machine_set_eeprom(Object *obj, const char *value,
     ms->eeprom = g_strdup(value);
 }
 
+static char *machine_get_avpack(Object *obj, Error **errp)
+{
+    XboxMachineState *ms = XBOX_MACHINE(obj);
+
+    return g_strdup(ms->avpack);
+}
+
+static void machine_set_avpack(Object *obj, const char *value,
+                               Error **errp)
+{
+    XboxMachineState *ms = XBOX_MACHINE(obj);
+
+    if (!xbox_smc_avpack_to_reg(value, NULL)) {
+        error_setg(errp, "-machine avpack=%s: unsupported option", value);
+        xbox_smc_append_avpack_hint(errp);
+        return;
+    }
+
+    g_free(ms->avpack);
+    ms->avpack = g_strdup(value);
+}
+
 static void machine_set_short_animation(Object *obj, bool value,
                                         Error **errp)
 {
@@ -501,6 +523,12 @@ static inline void xbox_machine_initfn(Object *obj)
                             machine_set_eeprom, NULL);
     object_property_set_description(obj, "eeprom",
                                     "Xbox EEPROM file", NULL);
+
+    object_property_add_str(obj, "avpack", machine_get_avpack,
+                            machine_set_avpack, NULL);
+    object_property_set_description(obj, "avpack",
+                                    "Xbox video connector: composite (default), scart, svideo, vga, rfu, hdtv, none", NULL);
+    object_property_set_str(obj, "composite", "avpack", NULL);
 
     object_property_add_bool(obj, "short-animation",
                              machine_get_short_animation,
--- a/hw/xbox/smbus.h
+++ b/hw/xbox/smbus.h
@@ -29,4 +29,7 @@ void smbus_xbox_smc_init(I2CBus *smbus, int address);
 void smbus_cx25871_init(I2CBus *smbus, int address);
 void smbus_adm1032_init(I2CBus *smbus, int address);
 
+bool xbox_smc_avpack_to_reg(const char *avpack, uint8_t *value);
+void xbox_smc_append_avpack_hint(Error **errp);
+
 #endif
--- a/hw/xbox/xbox.h
+++ b/hw/xbox/xbox.h
@@ -47,6 +47,7 @@ typedef struct XboxMachineState {
     /*< public >*/
     char *bootrom;
     char *eeprom;
+    char *avpack;
     bool short_animation;
 } XboxMachineState;
 


hw/xbox/smbus_xbox_smc.c
```diff
@@ -181,14 +181,34 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
...
+        smc->avpack_reg = SMC_REG_AVPACK_SCART;
+    }
+    else if (strcmp(avpack, "svideo") == 0) {
+        smc->avpack_reg = SMC_REG_AVPACK_SVIDEO;
+    }
+    else if (strcmp(avpack, "vga") == 0) {
+        smc->avpack_reg = SMC_REG_AVPACK_VGA;
+    }
+    else if (strcmp(avpack, "vga-sog") == 0) {
+        smc->avpack_reg = SMC_REG_AVPACK_VGA_SOG;
```
JayFoxRox: What does SOG stand for?

mark-riley: Sync on green

binarymaster: > https://xboxdevwiki.net/Video_Encoder

Sync-on-Green maybe.

hw/xbox/smbus_xbox_smc.c
```diff
@@ -181,14 +181,34 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
...
+    }
+    else if (strcmp(avpack, "hdtv") == 0) {
+        smc->avpack_reg = SMC_REG_AVPACK_HDTV;
+    }
```
JayFoxRox: Move this code for mapping name to constant into a function. It will be cleaner and can be re-used when we eventually add a monitor command to unplug and replug cables.

hw/xbox/smbus_xbox_smc.c
JayFoxRox: These constants are weird, see https://xboxdevwiki.net/AV_Cables for the MS names in the respective AV code + more notes.

hw/xbox/xbox.c
```diff
@@ -502,6 +518,11 @@ static inline void xbox_machine_initfn(Object *obj)
     object_property_set_description(obj, "eeprom",
                                     "Xbox EEPROM file", NULL);
 
+    object_property_add_str(obj, "avpack", machine_get_avpack,
+                            machine_set_avpack, NULL);
+    object_property_set_description(obj, "avpack",
+                                    "Xbox video connector: composite (default), scart, svideo, vga, vga-sog, hdtv", NULL);
```
JayFoxRox: Missing a `none` option (for when no cable is plugged in). Also see my comment regarding MS names for cable types.

binarymaster: @JayFoxRox there is no definition for "none" in the xqemu code, at least I didn't found it.

Used sources:
- https://github.com/xqemu/xqemu/blob/3b29f0888a2f942b9ad91907e4e12d47a72049f8/hw/xbox/smbus_xbox_smc.c#L55
- https://github.com/XboxDev/cromwell/blob/5b06e83cd1e2b0f7587e06946ffcc0f587536cc0/drivers/video/VideoInitialization.h#L15

I'm suspecting `0xFF` value, but there are no sources confirming that.

JayFoxRox: The 3 mode pin bits will be 1 when no cable is plugged in. So the SMC constant is 7. See link from mentioned comment: https://xboxdevwiki.net/AV_Cables

Ditch the current names (fabricated / guessed or by homebew scene / xbox-linux) and use the MS names.

MS never even released a VGA cable for Xbox, let alone the VGA-SOG stuff. Some modded bios added support for these modes, but there's no official adapters. We should focus on official / licensed original Xbox stuff, not some home-made cables which require hacks.

The only unknown cable type is SMC constant 5, which the kernel treats largely the same as mode 7 (no cable plugged). We don't know what it might have been used for (and don't need to expose it for now).

Generally trust XboxDevWiki over xbox-linux stuff (we've mostly absorbed their work and improved on it).

hw/xbox/smbus_xbox_smc.c
```diff
@@ -56,7 +56,9 @@
 #define     SMC_REG_AVPACK_SCART        0x00
 #define     SMC_REG_AVPACK_HDTV         0x01
 #define     SMC_REG_AVPACK_VGA_SOG      0x02
+#define     SMC_REG_AVPACK_RFU          0x03
 #define     SMC_REG_AVPACK_SVIDEO       0x04
+#define     SMC_REG_AVPACK_MODE5        0x05
```
JayFoxRox: I wouldn't even bother exposing this for now. I doubt we'll *ever* need it. And if someone required it, they could just add it themselves.

There's no good way to name it as-is and we *know* that some idiot will pick it and complain that it doesn't work :P

hw/xbox/smbus_xbox_smc.c
```diff
@@ -178,17 +180,53 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
...
+        return SMC_REG_AVPACK_SVIDEO;
+    }
+    if (strcmp(avpack, "mode5") == 0) {
+        return SMC_REG_AVPACK_5;
+    }
+    if (strcmp(avpack, "standard") == 0) {
+        return SMC_REG_AVPACK_COMPOSITE;
+    }
+    if (strcmp(avpack, "none") == 0) {
+        return SMC_REG_AVPACK_VGA;
```
JayFoxRox: "none" = VGA?

Please fix by adopting MS suffix. e.g `SMC_REG_AVPACK_NONE`

hw/xbox/xbox.c
haxar: I'm using the `hdtv` option for true VGA out & digital audio out on Xbox hardware, as a pin configured on the AVIP connector. The `vga` & `vga-sog` options are not officially supported. Sync-on-green is a hack to get `vsync` & `hsync`, but you can tap those signals directly from the GPU, amplify those signals to 5v & not need sync-on-green at all, which is a hardware mod for VGA out.

hw/xbox/smbus_xbox_smc.c
```diff
@@ -178,17 +179,50 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
...
+}
+
 static void smbus_smc_realize(DeviceState *dev, Error **errp)
 {
     SMBusSMCDevice *smc = XBOX_SMC(dev);
+    char *avpack;
 
     smc->version_string_index = 0;
+    /* pretend to have a composite av pack plugged in by default */
+    smc->avpack_reg = SMC_REG_AVPACK_COMPOSITE;
```
JayFoxRox: nit: For code clarity and localness, why not move this into an `else` for the `if (avpack) {` check? Then it's clear that we'll pick either what was provided OR this one.

*(Has pros and cons, I'll leave final decision to others or you)*

binarymaster: Let it be so. :slightly_smiling_face:

hw/xbox/smbus_xbox_smc.c
```diff
@@ -55,10 +55,11 @@
 #define SMC_REG_AVPACK              0x04
 #define     SMC_REG_AVPACK_SCART        0x00
 #define     SMC_REG_AVPACK_HDTV         0x01
-#define     SMC_REG_AVPACK_VGA_SOG      0x02
+#define     SMC_REG_AVPACK_VGA          0x02
+#define     SMC_REG_AVPACK_RFU          0x03
 #define     SMC_REG_AVPACK_SVIDEO       0x04
 #define     SMC_REG_AVPACK_COMPOSITE    0x06
```
JayFoxRox: Is called `_STANDARD` in kernel (and also as your "avpack" option, creating an inconsistency).

I wonder if we should make an exception here, and call it "composite" (including "avpack" option) across our codebase as to not confuse the reader with the MS name. "standard" is very vague - it was fitting at the time, but by today's standards, we'd consider "standard"=HDMI.

hw/xbox/xbox.c
```diff
@@ -502,6 +518,11 @@ static inline void xbox_machine_initfn(Object *obj)
     object_property_set_description(obj, "eeprom",
                                     "Xbox EEPROM file", NULL);
 
+    object_property_add_str(obj, "avpack", machine_get_avpack,
+                            machine_set_avpack, NULL);
+    object_property_set_description(obj, "avpack",
+                                    "Xbox video connector: standard (default), none, scart, svideo, vga, rfu, hdtv", NULL);
```
JayFoxRox: Consider ordering these by something meaningful. I'd put "none" first or last.

nit: Consider adopting same order in your mapping function.

hw/xbox/smbus_xbox_smc.c
binarymaster: I'd leave it as composite because it describes it better. Do you know other places where it needs to be fixed up for consistence?

JayFoxRox: Only your "avpack" option name probably.

hw/xbox/smbus_xbox_smc.c
```diff
@@ -189,6 +216,14 @@ static void smbus_smc_realize(DeviceState *dev, Error **errp)
     if (object_property_get_bool(qdev_get_machine(), "short-animation", NULL)) {
         smc->scratch_reg = SMC_REG_SCRATCH_SHORT_ANIMATION;
     }
+
+    avpack = object_property_get_str(qdev_get_machine(), "avpack", NULL);
```
mborgerson: `object_property_get_str` gives us ownership of the string returned, so we should `g_free` it if we no longer need it. Side note: we should also fix this in the other places we call `object_property_get_str` as I'm pretty sure we leak it in exactly the same way.

hw/xbox/smbus_xbox_smc.c
```diff
@@ -178,9 +179,35 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
...
+    }
+    if (strcmp(avpack, "rfu") == 0) {
+        return SMC_REG_AVPACK_RFU;
+    }
+    if (strcmp(avpack, "hdtv") == 0) {
+        return SMC_REG_AVPACK_HDTV;
+    }
+    if (strcmp(avpack, "none") == 0) {
+        return SMC_REG_AVPACK_NONE;
+    }
```
mborgerson: We fail to return a value here if none of the above strings match. We should error out in this case.

hw/xbox/xbox.c
```diff
@@ -476,6 +476,22 @@ static void machine_set_eeprom(Object *obj, const char *value,
...
+{
+    XboxMachineState *ms = XBOX_MACHINE(obj);
+
+    return g_strdup(ms->avpack);
+}
+
+static void machine_set_avpack(Object *obj, const char *value,
+                               Error **errp)
+{
+    XboxMachineState *ms = XBOX_MACHINE(obj);
```
mborgerson: We should check the string here to make sure it is valid, and error out if it is not an option.

hw/xbox/smbus_xbox_smc.c
binarymaster: Addressed. :ok_hand:

LukeUsher:
Just a note for later: On Chihiro, the SMC Constant codes have different meanings than on a retail Xbox.

The Chihiro version of D3D8 makes the assumption that 0 (AV_PACK_SCART on retail systems) is AV_PACK_VGA, and most Chihiro games will fail to boot if any other av-pack type is used.

JayFoxRox:
@LukeUsher Are those just AV constants (which aren't relevant here), or did the SMC constants (which are important here) also change?
*(Basically I'm asking if the input to the kernel mapping function changed, or if the mapping function was modified)*

Someone should add this information to https://xboxdevwiki.net/AV_Cables#Supported_signals_.2F_AV_cables

We probably want to create an issue about this Chihiro mapping to not delay this PR much further (unless we know the full mapping).

hw/xbox/smbus_xbox_smc.c
```diff
@@ -178,9 +179,36 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
...
+    assert(0);
```
JayFoxRox: nit: I think we normally use: `fprintf` to `stderr`, followed by `assert(false)`. But I might misremember or we might have done different things in different files.

hw/xbox/xbox.c
```diff
@@ -476,6 +476,34 @@ static void machine_set_eeprom(Object *obj, const char *value,
...
+
+    if (strcmp(value, "composite") != 0 &&
+        strcmp(value, "scart") != 0 &&
+        strcmp(value, "svideo") != 0 &&
+        strcmp(value, "vga") != 0 &&
+        strcmp(value, "rfu") != 0 &&
+        strcmp(value, "hdtv") != 0 &&
+        strcmp(value, "none") != 0)
```
JayFoxRox: nit: Not a big fan of all the duplications in this code. It's easy to break stuff by just changing the strings in one function, or by forgetting to add one of the items.

Also having this list hardcoded here makes it harder to add a monitor command later (duplicating this list *again*).

There aren't many clean options I can think of myself. I'd *personally* probably abuse `smbus_avpack_to_reg` and have it return an `int`, where `-1` simply means: no valid mapping found. Then you can do the error handling based on that (as that is still unique to each caller).
There might be better options (using a second argument for `bool* valid` return-value etc.) so it could be helpful to look how other files in QEMU handle this.

Again: this is nitpicking to me (for now). Other maintainers might think differently.

LukeUsher:
> @LukeUsher Are those just AV constants (which aren't relevant here), or did the SMC constants (which are important here) also change?
> _(Basically I'm asking if the input to the kernel mapping function changed, or if the mapping function was modified)_
> 
> Someone should add this information to https://xboxdevwiki.net/AV_Cables#Supported_signals_.2F_AV_cables
> 
> We probably want to create an issue about this Chihiro mapping to not delay this PR much further (unless we know the full mapping).

Chihiro games expect the SMC video mode to be set to 0 (this can be configured via the dip-switches on the filterboard). Chihiro games fail to boot if this is not the case & it also expects the kernel to return AV_PACK_VGA, but I'm not sure where exactly that behavior is implemented

JayFoxRox:
Thanks for clarifying.

We probably want to expose the DIP-Switches for "chihiro" machine type then (and not the internals of the Xbox board / possibly disable "avpack" on Chihiro). We should consider this a non-issue for this PR. A separate issue should be created to describe this difference.

hw/xbox/xbox.c
binarymaster: Improved. :slightly_smiling_face:  I looked up how qemu handles errors in other realize functions and did it the same way.

hw/xbox/xbox.c
```diff
@@ -476,6 +476,28 @@ static void machine_set_eeprom(Object *obj, const char *value,
...
+
+    return g_strdup(ms->avpack);
+}
+
+static void machine_set_avpack(Object *obj, const char *value,
+                               Error **errp)
+{
+    XboxMachineState *ms = XBOX_MACHINE(obj);
+
+    if (!smbus_avpack_to_reg(value, NULL))
```
mborgerson: This is a statically defined function in smbus_xbox_smc.c. We should make it public if we are calling it here

binarymaster: Indeed, thanks.

mborgerson:
Minor nit about commit message: smbus prefix isn't really descriptive enough. The functional component should be "smc". Something more like:
```
smc: Make attached avpack configurable
```

hw/xbox/smbus_xbox_smc.c
```diff
@@ -178,9 +179,43 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
...
+    if (strcmp(avpack, "hdtv") == 0) {
+        if (value) *value = SMC_REG_AVPACK_HDTV;
+        return true;
+    }
+    if (strcmp(avpack, "none") == 0) {
+        if (value) *value = SMC_REG_AVPACK_NONE;
+        return true;
+    }
+    return false;
+}
```
JayFoxRox: nit: Possible alternative for your consideration (I personally consider this cleaner):

```c
bool smbus_avpack_to_reg(char *avpack, uint8_t *value)
{
    uint8_t r;
    if (strcmp(avpack, "composite") == 0) {
        r = SMC_REG_AVPACK_COMPOSITE;
    } else if (strcmp(avpack, "scart") == 0) {
        r = SMC_REG_AVPACK_SCART;
    } else if (strcmp(avpack, "svideo") == 0) {
        r = SMC_REG_AVPACK_SVIDEO;
    } else if (strcmp(avpack, "vga") == 0) {
        r = SMC_REG_AVPACK_VGA;
    } else if (strcmp(avpack, "rfu") == 0) {
        r = SMC_REG_AVPACK_RFU;
    } else if (strcmp(avpack, "hdtv") == 0) {
        r = SMC_REG_AVPACK_HDTV;
    } else if (strcmp(avpack, "none") == 0) {
        r = SMC_REG_AVPACK_NONE;
    } else {
        return false;
    }

    if (value) *value = r;
    return true;
}
```

*(I personally also prefer `!strcmp`; I'm not sure what we do in the rest of the codebase)*

mborgerson: Agree except `if (value) *value = r;` should have braces per qemu coding conventions

binarymaster: ^ it's actually GNU code style, and also it's a huge nitpick :stuck_out_tongue: 

I'd leave it as is, because style like this was also used on line 109. :grin:

dracc:
Current behaviour (as of 073776e) is that a value is required or else xqemu stops. Would having a fallback to composite make sense?

hw/xbox/smbus_xbox_smc.c
mborgerson: Two wrongs and all :)

hw/xbox/xbox.c
```diff
@@ -476,6 +476,28 @@ static void machine_set_eeprom(Object *obj, const char *value,
...
 }
 
+static char *machine_get_avpack(Object *obj, Error **errp)
+{
+    XboxMachineState *ms = XBOX_MACHINE(obj);
+
+    return g_strdup(ms->avpack);
+}
+
+static void machine_set_avpack(Object *obj, const char *value,
```
mborgerson: We should probably support accepting the empty string here as well.

binarymaster: Fixed and tested, now everything should be good to go. :ok_hand:

hw/xbox/smbus_xbox_smc.c
```diff
@@ -178,9 +180,41 @@ static uint8_t smc_receive_byte(SMBusDevice *dev)
     return 0;
 }
 
+bool smbus_avpack_to_reg(char *avpack, uint8_t *value)
+{
+    uint8_t r;
+
+    if (!strlen(avpack)) {
+        /* pretend to have a composite av pack plugged in by default */
+        r = SMC_REG_AVPACK_COMPOSITE;
```
JayFoxRox: This code path is not good, because it prevents re-using this function for the monitor. We probably don't want a monitor command with empty argument to switch to composite.

@mborgerson and myself discussed this on XQEMU IRC; I'll mark what I believe was the solution we had agree'd on instead.

hw/xbox/smbus_xbox_smc.c
```diff
@@ -189,6 +223,14 @@ static void smbus_smc_realize(DeviceState *dev, Error **errp)
     if (object_property_get_bool(qdev_get_machine(), "short-animation", NULL)) {
         smc->scratch_reg = SMC_REG_SCRATCH_SHORT_ANIMATION;
     }
+
+    avpack = object_property_get_str(qdev_get_machine(), "avpack", NULL);
```
JayFoxRox: nit: `assert(avpack != NULL);` to guard against `NULL` case (we don't think it can happen; but better safe than sorry).

hw/xbox/xbox.c
```diff
@@ -476,6 +476,28 @@ static void machine_set_eeprom(Object *obj, const char *value,
...
+                               Error **errp)
+{
+    XboxMachineState *ms = XBOX_MACHINE(obj);
+
```
JayFoxRox: Handle empty strings here (if possible?).

```c
if (strlen(value) == 0) {
    /* pretend to have a composite av pack plugged in by default */
    value = "composite";
}
```

hw/xbox/smbus_xbox_smc.c
binarymaster: Why in the world we need to reuse avpack function for the monitor, is there any specific reason for that?

hw/xbox/smbus_xbox_smc.c
binarymaster: Added.

hw/xbox/smbus_xbox_smc.c
JayFoxRox: Because you can physically unplug / replug an AV cable at runtime on a real console. The [SMC will actually handle this case explicitly and signal an interrupt](http://xboxdevwiki.net/PIC#Interrupt_reasons).

I'm not sure whose responsibility it is (kernel or XDK), but most programs will probably only reboot the Xbox. However, we should still support it, as it's an option that people have with a physical Xbox, too.

The monitor command will be added in a future PR, but we should still prepare for it. Hence this helper function in the first place.

binarymaster: Improved.

haxar: Just a minor nit: `if (!strlen(avpack))` can be optimized by checking the first character of `avpack` if it is `0` via `if (!avpack[0])` or `if (!*avpack)` (what I do) & save yourself a call to `strlen()`.

This can break if the `avpack` pointer is `NULL` & I also sanity check for that.

JayFoxRox: It also makes it harder to read in a non-hot code path, for a case that would likely be optimized anyway (regardless of how the check is written).

hw/xbox/smbus_xbox_smc.c
```diff
@@ -189,6 +220,18 @@ static void smbus_smc_realize(DeviceState *dev, Error **errp)
...
+    assert(avpack != NULL);
+    if (strlen(avpack) == 0) {
+        /* pretend to have a composite av pack plugged in by default */
+        smc->avpack_reg = SMC_REG_AVPACK_COMPOSITE;
```
JayFoxRox: Leaving this marker for @mborgerson : default value is hardcoded here + zero-check, but duplicated below.

Is this fine? I consider it bad-code, but don't have a good alternative myself, and we have bikeshedded this a lot already.

@binarymaster said on Discord:

>  `machine_set_avpack` is not called at all if no `avpack` parameter was set in command line
>  but the string value still contains an empty string in that case
>  that's why there are two checks

hw/xbox/xbox.c
```diff
@@ -476,6 +476,34 @@ static void machine_set_eeprom(Object *obj, const char *value,
...
+    if (strlen(value) == 0) {
+        /* pretend to have a composite av pack plugged in by default */
+        g_free(ms->avpack);
+        ms->avpack = g_strdup("composite");
+        return;
+    }
```
JayFoxRox: Leaving marker for @mborgerson : dupliaction of empty-check and hardcoded default is here.

*(Also duplication of the code that's at the end of this function - avoidable by modifying `value`*)

mborgerson:
Thanks!

PR #256, opened 2019-09-21
build: Use bash

dracc:
`sh` evaluation of expressions is wonky.
`! expression` evaluates to the same as `expression` in `sh` which makes the dirty-check ineffective on some systems.

Tested locally on Debian Linux, works fine.
Would appreciate if someone could test the AppVeyor artifact on Windows and of course building it locally on both Windows and macOS.


GXTX:
This doesn't seem to break anything. Tested locally, and through a couple of different CI environments.

PR #263, opened 2019-10-08
nv2a: Add some missing RAMDAC registers

binarymaster:
They are needed to obtain current screen resolution.

image


binarymaster:
And here is the result :slightly_smiling_face: https://twitter.com/x86corez/status/1181908765899153408

PR #266, opened 2019-11-16
GitHub Actions CI

mborgerson:
Replaces AppVeyor and Travis-CI with GitHub Actions workflows to handle continuous integration for all platforms. Additionally introduces ccache to speed up compilation times by caching object files to be opportunistically recycled in future builds saving *a lot* of time on these build runners.

* Ubuntu and macOS builds are now under 3 minutes.
* Windows builds are down from ~20min to ~6min.
* Times don't include time spent in queue. GitHub seems to start the jobs almost immediately (for now anyway).

~~Don't merge me yet.~~ I've disabled source fortification [because I thought it was broken on msys2](https://github.com/msys2/MINGW-packages/issues/5803) ~~but I might just need to enable `-fstack-protector-strong`~~. ~~Will look into it before merge.~~ This should be looked into later. For now I'm going to leave as-is.

Note:
* ~~Windows platform builds are accelerated by downloading a pre-built msys2 environment from Dropbox (about 2.5 minutes download+extraction), which I should provide some more information about. Creating the environment by full install takes quite a long time, experimentally I've seen it take anywhere from 8-30 minutes.~~
* Update: Windows builds are now using a toolchain created automatically from https://github.com/mborgerson/msys2-tools/. This saves about 4+ minutes (and possible MSYS2 reliability issues) from every Windows CI build by caching the MSYS2 environment. I'll move the msys2-tools repo over to xqemu org after merge.

TODO:
* Download links and badges on the website will need to be updated.


JayFoxRox:
I'm not sure if just speed alone is a good metric for CI.
My main issue is unreliability of Travis and AppVeyor. If I know it always takes half an hour, that's fine with me. But if it sometimes takes 3 hours to even start, and then it fails while installing packages, then it's a problem.

One benefit of having Travis and AppVeyor is that we always have at least *some* CI if one of them fails. Personally, I also dislike adding too strong of a dependency on GitHub alone. However, I can tolerate GitHub Actions if it actually works better than the third-party solutions.

> Windows platform builds are accelerated by downloading a pre-built msys2 environment from Dropbox (about 2.5 minutes download+extraction)

I have a strong dislike for Dropbox. I hosted some files on my Dropbox account and the URLs were almost always blocked for being over usage limits.
Also Dropbox has been attacked in the past, which lead to login leaks - if I remember correctly, Dropbox handled the situation poorly back then.

There's probably better alternatives (if need-be).

> msys2 environment [...] which I should provide some more information about.

Yes, if it's not publicly documented and reproducible, we shouldn't do it.

mborgerson:
> I'm not sure if just speed alone is a good metric for CI. My main issue is unreliability of Travis and AppVeyor. If I know it always takes half an hour, that's fine with me. But if it sometimes takes 3 hours to even start, and then it fails while installing packages, then it's a problem.

Though an important factor, build speed is not the reason I decided to switch over the CI to GitHub actions. In fact, it wasn't even part of my initial motivation; however I was happy with the performance improvement and opted to share the good news.

The primary reasons I decided to switch things over (some of which I thought were sort of obvious at the time) are:
* Uniform feature set across target platforms (including nice features like artifact storage for all platforms),
* Consolidation of services into one manageable location,
* Reduction of burden on forks in setting up CI for themselves (they don't need to create accounts on other platforms anymore),
* Common configuration formats for all build platforms,
* Possible integration with other GitHub features, and
* What appears to be more reliable (instant) build queue times, at least for the time being.

> One benefit of having Travis and AppVeyor is that we always have at least some CI if one of them fails.

I don't see this as much of a potential issue, but we could revisit it if GitHub Actions becomes unreliable. In the ultra-paranoid case one could have multiple CI services running the same job (e.g. both TravisCI and GH running Linux/macOS builds), but I think that's overkill for the time being.

> Personally, I also dislike adding too strong of a dependency on GitHub alone. However, I can tolerate GitHub Actions if it actually works better than the third-party solutions.

I'm not sure it's something that needs to be "tolerated." The service doesn't cost anything, and I would prefer one proprietary solution over three proprietary solutions. In any case, it's easy enough to migrate to another platform if GitHub becomes unreliable.

> I have a strong dislike for Dropbox. [...] There's probably better alternatives (if need-be).

What better alternatives do you suggest? Download speeds from Dropbox are very fast, and I've never had an issue with reliability of the service. Admittedly, I've never depended on it for a build system.

> Also Dropbox has been attacked in the past [...]

I suppose I could add a file integrity check (e.g. `md5sum`) in the workflow to verify if the toolchain is valid or corrupt (maliciously or otherwise) and if so fallback to the slow path.

> Yes, if it's not publicly documented and reproducible, we shouldn't do it.

It's documented in the Windows workflow file how exactly to reproduce the environment, I've just cached it for build performance. I'll probably automate the environment build at some point.

I think GitHub actions is a nice addition to GitHub and that it will serve the project very well, and I plan to use it on my fork for the time being.

mborgerson:
This PR has been updated to now:
* Use a [GitHub repository](https://github.com/mborgerson/msys2-tools/) to automatically create and host the MSYS2 environment, as a release. This will likely get moved to org control shortly after merge.
* Use [ccache](https://ccache.dev/) to drastically accelerate build times on all platforms by opportunistically (and carefully) recycling object files from previous builds. Linux and macOS complete in under 3 minutes. Windows builds are at about 6 minutes. A majority of the time is spent in runner setup--actual compilation times are about 30s, 40s, 2m22 respectively.

.github/workflows/build_macos.yml
```diff
@@ -0,0 +1,52 @@
...
+          sdl2 \
+          libepoxy \
+          coreutils \
+          pkg-config \
+          ccache
+    - name: Initialize ccache
+      id: cache
+      uses: actions/cache@v1
+      with:
+        path: /tmp/ccache
```
JayFoxRox: This is `/tmp/xqemu-ccache` on other platforms. Why inconsistent?

.github/workflows/build_macos.yml
```diff
@@ -0,0 +1,52 @@
...
+        key: cache-${{ runner.os }}-${{ matrix.configuration }}-${{ github.sha }}
+        restore-keys: cache-${{ runner.os }}-${{ matrix.configuration }}-
```
JayFoxRox: I'm not sure how any of this really works, but I'm a bit surprised to see this end in a dash symbol `-` (which implies more data follows, as for the `key` property, which is actually followed by the hash).

.github/workflows/build_windows.yml
```diff
@@ -0,0 +1,59 @@
...
+        restore-keys: cache-${{ runner.os }}-${{ matrix.configuration }}-
+    - name: Compile
+      run: |
+        C:\tools\msys64\usr\bin\bash.exe -lc @"
+        set -e
+        cd `$GITHUB_WORKSPACE
+        export CCACHE_DIR=/tmp/xqemu-ccache
+        export CCACHE_MAXSIZE=250M
+        export PATH="/usr/lib/ccache/bin:/mingw64/bin:`$PATH"
+        export PATH=":`$PATH"
```
JayFoxRox: What does this do?

.github/workflows/build_windows.yml
```diff
@@ -0,0 +1,59 @@
...
+        ./build.sh ${{ env.BUILD_PARAM }}
+        echo -e "\n** Compiler Cache Stats"
```
JayFoxRox: Our other `echo` don't seem to do `\n** ` prefix. Why does this one do it?

.github/workflows/build_macos.yml
```diff
@@ -0,0 +1,52 @@
...
+        brew update
+        brew install \
+          glib \
+          pixman \
+          sdl2 \
+          libepoxy \
+          coreutils \
+          pkg-config \
+          ccache
+    - name: Initialize ccache
```
JayFoxRox: Should probably be "Initialize Compiler Cache" to make use of `ccache` internal, but more consistently keeping "Compiler Cache" in the human readable parts.

include/qemu/osdep.h
```diff
@@ -85,17 +85,17 @@ extern int daemon(int, int);
 #endif
 #endif
 
+/* enable C99/POSIX format strings (needs mingw32-runtime 3.15 or later) */
+#ifdef __MINGW32__
+#define __USE_MINGW_ANSI_STDIO 1
+#endif
+
```
JayFoxRox: Why was this code-block moved?

.github/workflows/build_windows.yml
```diff
@@ -0,0 +1,59 @@
...
+          print('::set-env name=BUILD_PARAM::--release')
+          print('::set-env name=ARTIFACT_NAME::xqemu-release')
+    - name: Clone Tree
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+    - name: Install Dependencies
+      run: |
+        echo "Downloading MSYS2 environment..."
+        Invoke-WebRequest -Uri "https://github.com/mborgerson/msys2-tools/releases/latest/download/msys64.7z" -OutFile "msys64.7z"
```
JayFoxRox: I think it's foreseeable that this PR will be merged.
Please move `mborgerson/msys2-tools` to `xqemu/ci-environments` or something.

I'm not happy about either of those repo names; I'd like something extensible in case we also create environments for other toolchains. The release is already called msys64.7z anyway (which means we could also add other environments to that repository).

If we do one repository per environment, the repo name should probably be along the lines of `ci-environment-msys64` (or `ci-environment-msys2` if we decide to share it for 32 bit?)

.github/workflows/build_macos.yml
mborgerson: Ok

.github/workflows/build_windows.yml
mborgerson: Copy-paste error. Removed.

.github/workflows/build_macos.yml
mborgerson: This is to work around what is an oversight of [actions/cache](https://github.com/actions/cache) (see link for docs). The problem is that if there's a cache hit, the cache will *not* be updated after the build--which means we only get objects from the first time the cache is created.

Unfortunately, there's no way to force an update of the cache post-build. However, I found by correlating the cache key with the commit hash causing the event (which will always be a cache-miss), then providing a restore-key common *prefix* (which actions/cache will use as a prefix in a reverse-chronological search of the cache list), I can get the cache to (1) always get the latest cached objects and (2) always create an updated cache. Naturally, old cache entries will be evicted as time goes on.

AS for the terminating `-`, this is the style used in GitHub docs.

.github/workflows/build_macos.yml
mborgerson: Debug artifact. Fixed.

.github/workflows/build_macos.yml
JayFoxRox: I'd have to read up on this. So I'll just blindly trust your expertise for this PR.
Here's some naive thoughts that I still had.

> Unfortunately, there's no way to force an update of the cache post-build. However, I found *[hacky-logic]*

I feel like this should be an issue on actions/cache (about how to force this update; and wether *[hacky logic]* is a suitable / future-proof workaround).
I'm a bit worried that this behaviour will break.

> [...] then providing a restore-key common prefix (which actions/cache will use as a prefix in a reverse-chronological search of the cache list),
> As for the terminating -, this is the style used in GitHub docs.

I'm surprised to hear that this is a prefix.
The README of actions/cache does not suggest that at all.

This should be reported upstream so documentation can be improved.
I'm a bit worried that this behaviour will break (where they might change how to mark it as a prefix).

mborgerson: [Please see here for the docs](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows#matching-a-cache-key):
> When a key doesn't match directly, the action searches for keys prefixed with the restore key. If there are multiple partial matches for a restore key, the action returns the most recently created cache.

> I feel like this should be an issue on actions/cache
I agree. For now, this will have to do.

> I'm a bit worried that this behaviour will break.
I think this is intended to be a functional solution to the issue. It appears to be by-design how this works, but I think instead it would be better if they at least provided a way to just refresh the cache.

.github/workflows/build_macos.yml
JayFoxRox: This wasn't fixed / pushed?

mborgerson: It's fixed, GitHub UI is lying to us

.github/workflows/build_windows.yml
mborgerson: I like the prefix personally, but removed it for now to prevent bikeshed.

include/qemu/osdep.h
mborgerson: Noted in [commit message](https://github.com/xqemu/xqemu/pull/266/commits/25302ddd2bd8a392a0000605db67e4deeb19845d):
> This patch moves the definition of __USE_MINGW_ANSI_STDIO to precede
inclusion of any headers. This is done to prevent stdio.h from assigning
a default value which was previously conflicting with our unconditional
definition, resulting in numerous compiler warnings.

.github/workflows/build_windows.yml
mborgerson: I accept the `ci-environment-msys2` proposal. Repo has been moved, and the PR has been updated.

.github/workflows/build_windows.yml
JayFoxRox: I was just wondering why.

I'm not even sure where I could see one of the GitHub Action logs, so I have no idea how / where / why / what it puts in the output log.

mborgerson: Personally I like the prefix to help differentiate our build system flow vs tool output. But this would be the first instance for now. In the future we should probably clean up the build output to make it a tad more readable, but it's fine for now.

You can see the whole workflow [here](https://github.com/mborgerson/xqemu/commit/8cdeb3de529ba2bc6fd4ccb4ba1f849a919f9480/checks?check_suite_id=325961312).

A warning: don't have Firefox open watching logs with something that spits out a lot of text to stdout. It causes my Firefox to lock up...apparently it's a known issue.

README.md
```diff
@@ -7,7 +7,8 @@ and Linux. Please visit [xqemu.com](http://xqemu.com) to learn more.
...
-| Debug   | [![Build status](https://ci.appveyor.com/api/projects/status/i1m0keigjabxg170/branch/master?svg=true)](https://ci.appveyor.com/project/xqemu-bot/xqemu/branch/master) | [![Build status](https://travis-matrix-badges.herokuapp.com/repos/xqemu/xqemu/branches/master/3)](https://travis-ci.org/xqemu/xqemu) | [![Build status](https://travis-matrix-badges.herokuapp.com/repos/xqemu/xqemu/branches/master/1)](https://travis-ci.org/xqemu/xqemu) |
+| Platform | Build Status |
+|----------|--------------|
+| Windows | [![Build status](https://github.com/xqemu/xqemu/workflows/Build%20(Windows)/badge.svg?branch=master)](https://github.com/xqemu/xqemu/actions?query=branch%3Amaster) |
+| macOS | [![Build status](https://github.com/xqemu/xqemu/workflows/Build%20(macOS)/badge.svg?branch=master)](https://github.com/xqemu/xqemu/actions?query=branch%3Amaster) |
+| Ubuntu | [![Build status](https://github.com/xqemu/xqemu/workflows/Build%20(Ubuntu)/badge.svg?branch=master)](https://github.com/xqemu/xqemu/actions?query=branch%3Amaster) |
```
JayFoxRox: nit: "Linux (Ubuntu)" would be slightly better in my opinion

README.md
```diff
@@ -7,7 +7,8 @@ and Linux. Please visit [xqemu.com](http://xqemu.com) to learn more.
...
-| Release | [![Build status](https://ci.appveyor.com/api/projects/status/i1m0keigjabxg170/branch/master?svg=true)](https://ci.appveyor.com/project/xqemu-bot/xqemu/branch/master) | [![Build status](https://travis-matrix-badges.herokuapp.com/repos/xqemu/xqemu/branches/master/4)](https://travis-ci.org/xqemu/xqemu) | [![Build status](https://travis-matrix-badges.herokuapp.com/repos/xqemu/xqemu/branches/master/2)](https://travis-ci.org/xqemu/xqemu) |
-| Debug   | [![Build status](https://ci.appveyor.com/api/projects/status/i1m0keigjabxg170/branch/master?svg=true)](https://ci.appveyor.com/project/xqemu-bot/xqemu/branch/master) | [![Build status](https://travis-matrix-badges.herokuapp.com/repos/xqemu/xqemu/branches/master/3)](https://travis-ci.org/xqemu/xqemu) | [![Build status](https://travis-matrix-badges.herokuapp.com/repos/xqemu/xqemu/branches/master/1)](https://travis-ci.org/xqemu/xqemu) |
+| Platform | Build Status |
+|----------|--------------|
+| Windows | [![Build status](https://github.com/xqemu/xqemu/workflows/Build%20(Windows)/badge.svg?branch=master)](https://github.com/xqemu/xqemu/actions?query=branch%3Amaster) |
```
JayFoxRox: Is there any repo where I can see these badges? I checked mborgerson/xqemu but they weren't there.
I also would like to see if there are per-platform links that we can use (nit).

mborgerson: No, the badges won't work here because they have the xqemu URL. They look like the old badges for the most part

JayFoxRox: Nevermind, they are here: https://github.com/mborgerson/xqemu/workflows/Build%20(macOS)/badge.svg - this should work fine.

README.md
JayFoxRox: Looking over the code and how they are displayed, this is more consistent (the badge shows "Build (Ubuntu)", and "Build (Linux (Ubuntu))" would look stupid).
I'm fine with this, although this is something that could be improved slightly.

.github/workflows/build_windows.yml
JayFoxRox: Yes, prefix looks fine in log (and makes sense) - like I said: I was just asking as I had no idea how / where it appears in the log.
I'm fine eitherway: with or without prefix.

README.md
JayFoxRox: Do we know if https://github.com/xqemu/xqemu/workflows/Build%20(Windows)/badge.svg?branch=master includes PRs? The docs don't seem to explain it.
AppVeyor had this weird `?ispr=false` nonsense (for download URLs) because `?branch=master` would also show PRs to master; if GitHub Actions does something similar, our status badge would display wrong information when a PR fails.

.github/workflows/build_windows.yml
```diff
@@ -0,0 +1,57 @@
...
+    #      C:\tools\msys64\usr\bin\bash -xlc "pacman --noconfirm -S --needed base-devel git make autoconf automake-wrapper mingw-w64-x86_64-libtool mingw-w64-x86_64-pkg-config mingw-w64-x86_64-glib2 mingw-w64-x86_64-SDL2 mingw-w64-x86_64-pixman mingw-w64-x86_64-libepoxy"
+    # - name: Compile
+    #   shell: cmd
+    #   run: |
+    #      C:\tools\msys64\usr\bin\bash -xlc "cd $GITHUB_WORKSPACE; export PATH="/mingw64/bin:$PATH"; ./build.sh"
+    #
+    - name: Install Dependencies
+      run: |
+        echo "Downloading MSYS2 environment..."
+        Invoke-WebRequest -Uri "https://dl.dropboxusercontent.com/s/8lt6464xtk00t59/msys64.7z" -OutFile "msys64.7z"
```
JayFoxRox: The never-really-used dropbox URL remains in the history. Switch to https://github.com/xqemu/ci-environment-msys2 is only in ccache commit.

build.sh
```diff
@@ -125,6 +125,6 @@ set -x # Print commands from now on
     --disable-usb-redir \
     ${user_opts}
 
-time make -j"${job_count}" 2>&1 | tee build.log
+time make -j"${job_count}" subdir-i386-softmmu 2>&1 | tee build.log
```
JayFoxRox: The "subdir-" make-option sounds quite aggressive and I'm not sure how it really works. I think that should have been a separate change (unless removing the option adds more dependencies to CI); if the Windows CI builds work I'm fine with trusting this.

I don't expect this to do much, as we already restrict targets to i386-softmmu anyway (and we'll disable even more things in #267).

JayFoxRox:
Reviewed PR. I have limited knowledge about GitHub Actions / CI in general, but this looks fine.

A couple of things I noticed / thought:
- Has anybody tested the Windows artifacts? Should we have some people on XQEMU Discord be guinea-pigs?
- The [retention period for GitHub Artifacts is only 3 months (90 days)](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/persisting-workflow-data-using-artifacts#about-workflow-artifacts) (I think it was 6 months for AppVeyor?); this is probably fine, but we will have to pay attention to this in the future.
- What would we link on the website? I'm also having trouble to find docs for artifact URLs (I've skimmed the docs and the forum). This could be a big problem.
- How do we merge this? I assume we have to log in to Travis CI and AppVeyor to disable those, too? I'm not sure if I have the necessary credentials / rights for that.
- https://github.com/mborgerson/xqemu/actions is 404 for GitHub guests (when not logged in); this includes the links from the README (the badge images still work fine). We should report this issue to GitHub?

I'd like to finish migration of this by the end of the weekend unless we there's a problem with any of the things I've mentioned here.

.github/workflows/build_macos.yml
```diff
@@ -0,0 +1,52 @@
...
+          pkg-config \
+          ccache
+    - name: Initialize Compiler Cache
+      id: cache
+      uses: actions/cache@v1
+      with:
+        path: /tmp/xqemu-ccache
+        key: cache-${{ runner.os }}-${{ matrix.configuration }}-${{ github.sha }}
+        restore-keys: cache-${{ runner.os }}-${{ matrix.configuration }}-
+    - name: Compile
```
JayFoxRox: nit: I prefer "Build" over "Compile". Matches our build.sh and also what is proposed for nxdk CI right now.

mborgerson:
> Has anybody tested the Windows artifacts? Should we have some people on XQEMU Discord be guinea-pigs?

I have not tested the Windows artifacts, but I expect they are unchanged. We are building on Windows using the same process as before, with the msys2 toolchain. If there's an issue introduced with this PR, I would be surprised; nevertheless, it wouldn't hurt to ask someone to test it out.

> The retention period for GitHub Artifacts is only 3 months (90 days) (I think it was 6 months for AppVeyor?); this is probably fine, but we will have to pay attention to this in the future.

I think 3 month retention for artifacts is adequate. It allows people to use the bleeding edge build, and a large enough window to let people grab older builds to help with regression testing, etc.

> What would we link on the website? I'm having trouble to find docs for artifact URLs (I've skimmed the docs and the forum). This could be a big problem if there's no usable URL.

In my opinion, this is the biggest issue with GitHub's artifact storage. It [appears to be a known issue](https://github.com/actions/upload-artifact/issues/27), but I haven't seen any communication from GitHub on how they plan to address this need.

With GitHub releases, there is a simple URL that can be used to retrieve the latest release files--but it appears there is no such way to easily retrieve files from actions artifacts. For this reason, I've decided to use GitHub releases instead of artifacts for [ci-environment-msys2](https://github.com/xqemu/ci-environment-msys2/blob/master/.github/workflows/gen_msys2_env.yml) and also for [XboxDev/cromwell](https://github.com/XboxDev/cromwell/releases). Ideally GitHub will provide such an option for artifacts as well.

In the mean time, I think we can do one of a few things:
* Option 1: We can start using GitHub releases to host the latest version of XQEMU in a single release, overwriting the binary whenever the "master" branch is updated. Simple and easy to use, but older artifacts are hard to acquire. We could also have multiple artifacts be uploaded to a single release, giving the last 7 or so builds and prune older builds (e.g. [this action](https://github.com/marketplace/actions/deploy-nightly)).
* Option 2: We can start using GitHub releases and push a new release every time "master" branch is updated as I have demonstrated in links above, giving a nice linear history of releases but causing a bit of chaos with many release tags and GitHub notifications.
* Option 3: We can push build artifacts to a 3rd party service. [JFrog Bintray](https://bintray.com/signup/oss) offers a free service for open source projects giving 10GB storage and 1TB transfer with CDN. This service lets us simply make an HTTP PUT request (scriptable via `curl`) to upload build artifacts. Bintray also has a [simple, flat publicly accessible webpage with links](https://dl.bintray.com/mborgerson/xqemu-mborgerson), badges, and indirect links to download the latest version, e.g. [ Download ](https://bintray.com/mborgerson/xqemu-mborgerson/xqemu/_latestVersion). This would be easy to integrate with GitHub actions, but adds a bit of management complexity as it introduces another 3rd party service, and I'd need to script removal of old artifacts.

Honestly, I'm fine with any of the three options above. Option 2 is probably the simplest approach for now as I already have a ready-to-go solution, it doesn't introduce a 3rd party service into the mix, and makes older builds easily available to users through the releases tab. I dislike how disruptive it can be, but there is not so much activity on the repository for it to be a major issue (imo).

I'm curious to hear what others think. Let's agree on a direction and I'll update the PR.

 > How do we merge this? I assume we have to log in to Travis CI and AppVeyor to disable those, too? I'm not sure if I have the necessary credentials / rights for that.

It can be merged through GitHub UI. I'll disable AppVeyor and Travis when we are ready to merge.

> https://github.com/mborgerson/xqemu/actions is 404 for GitHub guests (when not logged in); this includes the links from the README (the badge images still work fine). We should report this issue to GitHub?

Good catch. I didn't notice that the Actions tab is missing when not logged in. I think it should be exposed for public view.

JayFoxRox:
Sorry for taking a long time to respond; I had hoped I'd get responses from Windows-users or GitHub (as I reached out after reading your comment) or that the issues would resolve themselves more quickly. However, I never got any useful responses so I kept procrastinating.

> I would be surprised; nevertheless, it wouldn't hurt to ask someone to test it out.

No response; so ping-ing Windows-users @GXTX @LukeUsher @Ernegien

> I think 3 month retention for artifacts is adequate. [...] a large enough window to let people grab older builds to help with regression testing, etc.

Fair enough; we'll just have to remember to re-push for CI if we don't actively work on XQEMU.
We probably had some periods with absolutely no new commit in 3 months in the past.
It's a minor issue that we can ignore though.

> > What would we link on the website? [...]
> In my opinion, this is the biggest issue with GitHub's artifact storage. It [appears to be a known issue](https://github.com/actions/upload-artifact/issues/27), but I haven't seen any communication from GitHub on how they plan to address this need.

I did ask on the linked issue but got very brief and rather uninformative replies (not even a rough estimate). With GitHub being so secretive about their plans it's hard to tell how long it will take.

> In the mean time, I think we can do one of a few things:
> 
>  * Option 1: We can start using GitHub releases to host the latest version of XQEMU in a single release, overwriting the binary whenever the "master" branch is updated. [...]
>  * Option 2: We can start using GitHub releases and push a new release every time "master" branch is updated [...]
>  * Option 3: We can push build artifacts to a 3rd party service. [...]

To me..

* Option 1 sounds okay; but adds complexity.
* Option 2 sounds okay; but I feel like we should remove those releases once we migrate to the actual solution. Ideally we'd push releases to a separate repo?
* Option 3 sounds bad; I don't want to add dependency on another third party service (for a temporary solution).

Additionally I'm proposing another option:

* Option 4: could be to just not advertise binaries temporarily. We could tell people to just manually fetch those files (or even update the website manually if there are major features). At current pace, I don't think this would be much trouble. Other than the Actions tab, the actual artifact URLs also work for non-logged in GitHub users.

I'm actually in strong favor of Option 4: we'd just do it manually and wait for GitHub to finish the API.
If the API still isn't done by mid 2020 (or updating manually is too much work), then we should consider other alternatives again.

> I didn't notice that the Actions tab is missing when not logged in. I think it should be exposed for public view.

[I asked on the forums](https://github.community/t5/GitHub-Actions/Make-the-quot-Actions-quot-tab-public/m-p/40451/highlight/true#M4241) but got no response.
There are multiple threads about this problem.

---

@mborgerson If we go with option 4 + get feedback from a Windows user, then you can go ahead with disabling old CI (and merging).

mborgerson:
@JayFoxRox Your proposal of Option 4 is fine with me for the time being. Hopefully GitHub adds this simple feature in the near future. I will merge this after someone with access to a Windows machine reports success.

mborgerson:
Windows build [smoke tested by CYRiXplaysHalo](https://discordapp.com/channels/428359196719972353/428359372842991616/657805550071054336) (thanks!).

PR #267, opened 2019-11-24
build: Disable some unnecessary features

mborgerson:
Disables support for more stuff not relevant to Xbox emulation. Also disables support for various image formats. Decent size savings of 2.8M in Windows binary size and eliminates 165 C files from being compiled.

```
$ grep "CC" build.log.before | wc -l
947
$ grep "CC" build.log.after | wc -l
782
$ ls -lh before/xqemu.* after/xqemu.*
-rw-r--r-- 1 mborgerson mborgerson 4.9M Nov 24  2019 after/xqemu.exe
-rw-r--r-- 1 mborgerson mborgerson 6.1M Nov 23 23:01 after/xqemu.zip
-rw-r--r-- 1 mborgerson mborgerson 7.7M Oct 18 21:18 before/xqemu.exe
-rw-r--r-- 1 mborgerson mborgerson 7.1M Oct 18 14:18 before/xqemu.zip
```

Note: The failing CI seems to be an issue with Homebrew's Python not having a proper hashlib. I don't plan to debug or fix that as I have a total overhaul of the entire CI system pending in #266 with which I have tested this commit previously and it built fine.


default-configs/i386-softmmu.mak
```diff
@@ -25,3 +25,4 @@
 CONFIG_ISAPC=y
```
mborgerson: Ideally these should be able to be disabled and only `CONFIG_XBOX` should need to be enabled, but there are some dependencies here that I haven't eliminated yet. Leaving as-is for now.

hw/i386/Kconfig
```diff
@@ -99,3 +99,8 @@ config VMPORT
 config VMMOUSE
     bool
     depends on VMPORT
+
+config XBOX
+    select PC
```
mborgerson: Probably incomplete list, but `USB_OHCI` and `AC97` are required for us when building (`--without-default-devices` disables them otherwise). Leaving as-is for now.

default-configs/i386-softmmu.mak
JayFoxRox: We should probably create an issue after merge.

It's probably a bit confusing to have a dedicated `XBOX` option that still depends on other options (when Kconfig is meant to take care of this).

JayFoxRox:
Thanks! Getting rid of unnecessary code is a noble goal - it improves compilation time and avoids waste of disk space; it will also be helpful if we ever want to switch build-systems or when we detach from upstream.

---

I have done a fresh clone of this and running `./build.sh` produced a working build.

The changes in build.sh make sense.

I'm not qualified to comment on the Kconfig but assume it's fine, too.
However, my understanding is that Kconfig is supposed to deal with dependencies, so we should create an issue to document that `CONFIG_XBOX` doesn't do that properly yet (already mentioned in https://github.com/xqemu/xqemu/pull/267#discussion_r352307297).

Only waiting for #266 so this can be re-ran on new CI, then merged.

build.sh
```diff
@@ -123,6 +123,17 @@ set -x # Print commands from now on
...
+    --disable-bochs \
+    --disable-cloop \
+    --disable-dmg \
+    --disable-vdi \
+    --disable-vvfat \
+    --disable-qcow1 \
+    --disable-qed \
+    --disable-parallels \
+    --disable-sheepdog \
+    --without-default-devices \
```
JayFoxRox: I just noticed that I didn't try testing any uncommon devices like usb-forwarding or usb-storage.
Could this be excluded by this option? We should check that.

hw/i386/Kconfig
JayFoxRox: Is this already necessary or just preparation for when we disable those things mentioned in [your other comment](https://github.com/xqemu/xqemu/pull/267/files#r349904622)?

build.sh
JayFoxRox: This is actually a problem:

qemu-system-i386: -device usb-storage,drive=xmu_1a_path,port=3.2: 'usb-storage' is not a valid device model name

This feature is exposed from XQEMU-Manager, so this would break it.

dracc: Using a Duke through the `usb-host` driver works like before.

mborgerson: Good catch. Will fix!

hw/i386/Kconfig
mborgerson: Yes this is necessary

build.sh
mborgerson: Fixed. usb-storage is enabled again. Requesting someone help test

dracc:
I've now tested commit 13827d2, both controller forwarding and virtual memory cards still works.

mborgerson:
Thanks for the test @dracc

PR #271, opened 2020-02-15
Create GitHub 'release' on push to master

mborgerson:
This adds a new job to the CI workflow to create a release upon push to master (i.e. merging a PR). We can consider alternative distribution methods for the *very latest build* (as opposed to point releases) in the future, but for now this should be a stable way to distribute XQEMU since we don't use point releases. More stable than GH artifacts anyway.

Preview here: https://github.com/mborgerson/xqemu/releases

Caveat: A new tag is created for every release created. Since the volume of merged PRs is low, I don't see this as a major issue.


.github/workflows/build.yml
```diff
@@ -0,0 +1,219 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Create Build Tag
+      run: |
+        export BUILD_TAG=build-$(date -u +'%Y%m%d%H%M')
+        echo "::set-env name=BUILD_TAG::$BUILD_TAG"
+        mkdir -p dist
+        echo -n $BUILD_TAG > dist/tag
+    - name: Upload Artifacts
+      uses: actions/upload-artifact@master
```
Margen67: https://github.com/actions/upload-artifact/issues/41

mborgerson:
I've merged workflows which means we won't have individual badge status for each platform. I will look for a solution, but I may just unify the badges

.github/workflows/build.yml
```diff
@@ -0,0 +1,219 @@
...
+          print('::set-env name=BUILD_PARAM::')
+          print('::set-env name=ARTIFACT_NAME::xqemu-win-debug.zip')
+        else:
+          print('Configuring for Release')
+          print('::set-env name=BUILD_PARAM::--release')
+          print('::set-env name=ARTIFACT_NAME::xqemu-win-release.zip')
+    - name: Clone Tree
+      uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
```
Margen67: v2 of this action (not sure about v1) has fetch-depth 1 by default.

.github/workflows/build.yml
```diff
@@ -0,0 +1,219 @@
...
+      uses: actions/download-artifact@v1
+      with:
+        name: dist
+    - name: Get Package Info
+      run: |
+        echo "::set-env name=BUILD_TAG::$(cat dist/tag)"
+    - name: Create Release
+      if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+      id: create_release
+      uses: actions/create-release@v1.0.0
```
Margen67: This version number can be shortened to just `v1`.

mborgerson: Thanks, copy paste from old documentation

.github/workflows/build.yml
mborgerson: Cool, didn't know they updated to v2. Thanks!

.github/workflows/build.yml
mborgerson: Thanks for the link. Copy-paste from GitHub's own old docs

JayFoxRox:
> Caveat: A new tag is created for every release created. Since the volume of merged PRs is low, I don't see this as a major issue.

I don't see this as a *major* issue, but still an issue. I'd rather not have this on the main repository, but pushing it elsewhere also adds a lot of complexity. This is really just a temporary fallback to me.

> I've merged workflows which means we won't have individual badge status for each platform.

Why?
Overall, I feel like this is a downgrade (although it's hard to tell due to poor documentation by GitHub) and I'm not sure what changed since the original GitHub Actions PR.

---

Can/Will we later remove these releases easily to clean up the repo?

Did you look into marking releases a pre-release (GitHub has a feature for that I believe)?

I'd personally rather avoid this PR because it feels like another temporary hack.
I'd rather just upload the latest CI release to a comment on one of the xqemu or xqemu.com issues / PRs, and link that manually, until https://github.com/actions/upload-artifact/issues/51 (or similar) is resolved.

mborgerson:
>> I've merged workflows which means we won't have individual badge status for each platform.

>Why? Overall, I feel like this is a downgrade (although it's hard to tell due to poor documentation by GitHub) and I'm not sure what changed since the original GitHub Actions PR.

The reason I have done this is that within a workflow you can make one job depend on another, e.g. the release step should only happen if all 3 platforms can build. I could probably make some custom workflow logic to detect success of other workflows but that would require more of my time and be going against the design pattern for actions.

It's not really a downgrade for us. The only real downside is not having individual badges for each platform build; but for that I don't really care too much because if CI is failing for any platform, we should fix it.

> Can/Will we later remove these releases easily to clean up the repo?

Yes we can. Will we? If necessary. Likely we only need to keep a few versions around for people to help test with.

> Did you look into marking releases a pre-release (GitHub has a feature for that I believe)?

I did not. Afaik pre-release is mostly just a flag for users to be aware. We can turn it on, but I think it's pretty clear that XQEMU has been "pre-release" forever.

> I'd personally rather avoid this PR because it feels like another temporary hack.

Can you elaborate here? I'm not sure why this is considered a "hack." We have always considered the head of master to be our latest and greatest version of XQEMU, so we should be releasing it. I think abusing and linking to build artifacts to be the hack, this is an official release distribution page.

> I'd rather just upload the latest CI release to a comment on one of the xqemu or xqemu.com issues / PRs, and link that manually, until actions/upload-artifact#51 (or similar) is resolved.

I have zero interest in spending time to manually upload release tarballs and update links on the website whenever someone wants a new release. That just sounds like way more work than necessary after I have invested a significant amount of my time getting CI to be fast, stable, and auto publish releases. I listed multiple solutions for releasing builds in [my original GitHub Actions PR](https://github.com/xqemu/xqemu/pull/266#issuecomment-560786471), but I consider this to be the most convenient and stable way for now.

PR #259, opened 2019-10-02
Merge upstream QEMU v4.1.0

mborgerson:
Don't merge me. I want to clean up the merge commit.

Making PR for review. Don't try to rebase or merge this through GitHub UI, it should be merged manually (unless merges are temporarily enabled again).

~~Requesting others test this on all platforms.~~
* I've smoke-tested on Ubuntu 18.04.
* I've smoke-tested on macOS 10.14.5.
* [Windows build smoke-tested by Vinny Mendoza](https://discordapp.com/channels/428359196719972353/428359372842991616/629128273934417930)

PR #273, opened 2020-02-19
ci: Switch to checkout v2, ignore irrelevant files

Margen67:
There's probably a few files I missed so feel free to point them out.

---

Disable fail-fast.
fetch-depth: 1 is no longer needed since v2 does this by default.
Use v1 of upload-artifact instead of master, and fix inconsistency with ARTIFACT_NAME.

### DIFF
--- a/.github/workflows/build_macos.yml
+++ b/.github/workflows/build_macos.yml
@@ -1,11 +1,50 @@
 name: Build (macOS)
 
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - 'docs/*'
+      - 'docs/*/*'
+      - '.shippable.yml'
+      - '.cirrus.yml'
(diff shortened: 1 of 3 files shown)


GXTX:
I agree we shouldn't be running for those ignores, however XQEMU as a project doesn't currently edit or maintain any of those files (besides `README.md`) - they're direct upstream which is only updated/merged as part of larger patches (see #259 as an example). So I'm not sure I see the benefit.

Margen67:
> 
> 
> I agree we shouldn't be running for those ignores, however XQEMU as a project doesn't currently edit or maintain any of those files (besides `README.md`) - they're direct upstream which is only updated/merged as part of larger patches (see #259 as an example). So I'm not sure I see the benefit.

I was thinking it might be useful for rebases or future PRs or something.

PR #282, opened 2020-06-02
lpc: Replace assertions with warnings

binarymaster:
Third-party software might be buggy and thus trigger these assertions,
however it's not critical and the emulation should not be stopped.

I was able to trigger it by running CPU-Z v1.92 in ReactOS installation.
Without this change XQEMU release build just halts the emulation.

image

This change is inspired by:
https://github.com/StrikerX3/OpenXBOX/blob/1a3d5fbdda6891ee1b6813a515d973d9f9b35b3d/src/core/openxbox/hw/basic/superio.cpp#L121

### DIFF
--- a/hw/xbox/lpc47m157.c
+++ b/hw/xbox/lpc47m157.c
@@ -108,10 +108,18 @@ static void lpc47m157_io_write(void *opaque, hwaddr addr, uint64_t val,
     if (addr == 0) {
         /* INDEX_PORT */
         if (val == ENTER_CONFIG_KEY) {
-            assert(!s->configuration_mode);
+            if (s->configuration_mode) {
+                printf("lpc47m157 io write: Attempted to reenter configuration mode\n");
+            }
+            DPRINTF("lpc47m157 io write: Entering configuration mode\n");
+
             s->configuration_mode = true;
         } else if (val == EXIT_CONFIG_KEY) {
-            assert(s->configuration_mode);
+            if (!s->configuration_mode) {
+                printf("lpc47m157 io write: Attempted to reexit configuration mode\n");
+            }
+            DPRINTF("lpc47m157 io write: Exiting configuration mode\n");
+
             s->configuration_mode = false;
 
             update_devices(s);


hw/xbox/lpc47m157.c
```diff
@@ -108,10 +108,18 @@ static void lpc47m157_io_write(void *opaque, hwaddr addr, uint64_t val,
...
+            if (s->configuration_mode) {
+                printf("lpc47m157 io write: Attempted to reenter configuration mode\n");
+            }
+            DPRINTF("lpc47m157 io write: Entering configuration mode\n");
+
             s->configuration_mode = true;
         } else if (val == EXIT_CONFIG_KEY) {
-            assert(s->configuration_mode);
+            if (!s->configuration_mode) {
+                printf("lpc47m157 io write: Attempted to reexit configuration mode\n");
```
JayFoxRox, line 119: I haven't worked with the QEMU codebase in a while, but it feels like this should go to `stderr` or some warning / error handling code?

JayFoxRox:
Is it actually re-entering or re-exiting when accessing these registers during configuration mode / outside of configuration mode?

Also why would CPU-Z even have the permission to access the hardware? Could this be a bug in reactOS where it leaves the serial in bad state?

The change doesn't sound logical to me, so I'm a bit hesitant about merging this.

---

> (bikshed) This feels like this would be useful as a macro, wouldn't it? (`ABORT_ON_ASSERT_WARN` and `ASSERT_WARN`)

We should keep such discussions out of this PR; it's a general issue in XQEMU (and potentially QEMU) that we don't separate some of the asserts, errors, and warnings.

I also think that we should have features like `UNIMPLEMENTED(message)`, `WARN(message)`, `ERROR(message)`, `UNTESTED(message)`, and potentially with `ASSERT_` variations or categories for convenience, but this is not the place to discuss it.

hw/xbox/lpc47m157.c
binarymaster: `stderr` makes more sense, but I just followed the style used for DEBUG macro in the same file. There should be a separate function that accepts a file handle.

hw/xbox/lpc47m157.c
```diff
@@ -108,10 +108,18 @@ static void lpc47m157_io_write(void *opaque, hwaddr addr, uint64_t val,
     if (addr == 0) {
         /* INDEX_PORT */
         if (val == ENTER_CONFIG_KEY) {
-            assert(!s->configuration_mode);
+            if (s->configuration_mode) {
+                printf("lpc47m157 io write: Attempted to reenter configuration mode\n");
```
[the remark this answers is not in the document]
binarymaster: Need to think about better wording... :thinking:

binarymaster:
> Is it actually re-entering or re-exiting when accessing these registers during configuration mode / outside of configuration mode?

Actually I don't know, maybe technical specification have some notes about it. But also I think the code I referred as source of inspiration exists for a reason. :)

> Also why would CPU-Z even have the permission to access the hardware?

It has its own kernel driver, so it could do anything. Writing to these registers in such way is probably a part of their LPCIO detection algorithm.

> Could this be a bug in reactOS where it leaves the serial in bad state?

Unlikely. At least in XQEMU it always works for me (before and after running CPU-Z).

binarymaster:
Superseded by https://github.com/mborgerson/xemu/pull/96

PR #283, opened 2020-06-09
Improve AC97 for use with nxdk homebrew

JayFoxRox:
There are currently plans to synchronize analog and digital audio buffers in nxdk. This PR is necessary to support this.

This mostly works around (or fixes) bugs in the existing code.

The integration isn't very clean, but that's already a problem in the original QEMU / XQEMU codebase. So, it's still very different from what a real Xbox would do.

If necessary, I grouped S/PDIF with the PCM output. In the mixer code for example (which isn't present on Xbox).
A more Xbox-focused AC97 implementation can follow in the future.

This PR also gets rid of the confusing "ac97: invalid bm_index(3) in voice_set_active" messages.

### DIFF
--- a/hw/audio/ac97.c
+++ b/hw/audio/ac97.c
@@ -176,6 +176,7 @@ enum {
 
 #define GET_BM(index) (((index) >> 4) & 7)
 
+static void so_callback (void *opaque, int free);
 static void po_callback (void *opaque, int free);
 static void pi_callback (void *opaque, int avail);
 static void mc_callback (void *opaque, int avail);
@@ -272,6 +273,10 @@ static void voice_set_active (AC97LinkState *s, int bm_index, int on)
         AUD_set_active_in (s->voice_mc, on);
         break;
 
+    case SO_INDEX:
+        AUD_set_active_out (s->voice_so, on);
+        break;
+
     default:
         AUD_log ("ac97", "invalid bm_index(%d) in voice_set_active\n", bm_index);
         break;
@@ -366,6 +371,18 @@ static void open_voice (AC97LinkState *s, int index, int freq)
                 &as
                 );
             break;
+
+        case SO_INDEX:
+            s->voice_so = AUD_open_out (
+                &s->card,
+                s->voice_so,
+                "ac97.so",
+                s,
+                so_callback,
+                &as
+                );
+            break;
+
         }
     }
     else {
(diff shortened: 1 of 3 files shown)


hw/audio/ac97.c
```diff
@@ -1112,9 +1167,11 @@ static void transfer_audio (AC97LinkState *s, int index, int elapsed)
             if (r->civ == r->lvi) {
                 dolog ("Underrun civ (%d) == lvi (%d)\n", r->civ, r->lvi);
 
-                new_sr |= SR_LVBCI | SR_DCH | SR_CELV;
+                //new_sr |= SR_LVBCI | SR_DCH | SR_CELV;
```
JayFoxRox, line 1170: I don't understand why this is necessary. On hardare it only seems to set `SR_LVBCI`. However, if I do that, the interrupts won't be triggered anymore and audio drains.

hw/audio/ac97.c
```diff
@@ -1112,9 +1167,11 @@ static void transfer_audio (AC97LinkState *s, int index, int elapsed)
...
                 stop = 1;
                 s->bup_flag = (r->bd.ctl_len & BD_BUP) ? BUP_LAST : 0;
+
+                fetch_bd (s, r);
```
JayFoxRox, line 1174: This isn't necessary, but I assume this happens on hardware.

hw/audio/ac97.c
```diff
@@ -291,6 +297,7 @@ static void reset_bm_regs (AC97LinkState *s, AC97BusMasterRegs *r)
     r->cr = r->cr & CR_DONT_CLEAR_MASK;
     r->bd_valid = 0;
 
+printf("bm_reset 0x%X\n", r - s->bm_regs);
```
JayFoxRox, line 300: The commit with the debug changes would be removed before merge. They are only here for debugging purposes.

*(This particular message is also using an incorrect argument type, potentially causing stack corruption)*

hw/audio/ac97.c
```diff
@@ -434,6 +462,12 @@ static void update_combined_volume_out (AC97LinkState *s)
     rvol = (rvol * prvol) / 255;
 
     AUD_set_volume_out (s->voice_po, mute, lvol, rvol);
+
+    mute = 1;
+    lvol = 0;
+    rvol = 0;
+
+    AUD_set_volume_out (s->voice_so, mute, lvol, rvol);
```
JayFoxRox, line 470: This forces silence on "ac97.so" (digital audio port).

Ideally we'd have some S/PDIF decoder in XQEMU-Manager and find a way to choose between "ac97.po" (analog audio port) and this one.

I'll create an issue on xqemu/xqemu or xqemu/xqemu-manager after merge.

PR #286, opened 2020-07-12
Added support for FreeBSD

Martinfx:
Hi.
I added support for FreeBSD with build gmake.


GXTX:
Couple of quickies:

No need to edit `.cirrus.yml` this is tracked from upstream QEMU and isn't used by anything here.
Commit title should have prefix of `build:`

I'm not entirely sure if it's your changes or not that are breaking the macOS CI runner. I think probably(?) since it expects to be building `_cgl` not `_glx` for Darwin.

build.sh
```diff
@@ -17,6 +17,7 @@ debug_opts='--enable-debug'
 user_opts=''
 build_cflags='-O0 -g'
 job_count='4'
+make_opts='make'
```
JayFoxRox, line 20: `make_opts` sounds misleading for this, because this is just the name of the make-command, and not just the options for `make`.

build.sh
```diff
@@ -45,6 +46,12 @@ case "$(uname -s)" in # adjust compilation option based on platform
         sys_cflags='-march=native -Wno-error=redundant-decls -Wno-error=unused-but-set-variable'
         sys_opts='--enable-kvm --disable-xen --disable-werror'
         ;;
+    FreeBSD)
```
JayFoxRox, line 49: nit: I'd add this block as first case, so we have the reverse order of the default order-by-popularity (which then also happens to be alphabetical): Windows, MacOS [Darwin], Linux, BSD.

I'm not sure why the current order in this file is reverse to begin with (or why it's somewhat random).

hw/xbox/nv2a/gl/Makefile.objs
```diff
@@ -2,3 +2,4 @@ obj-y += gloffscreen_common.o
 obj-$(CONFIG_WIN32) += gloffscreen_wgl.o
 obj-$(CONFIG_DARWIN) += gloffscreen_cgl.o
 obj-$(CONFIG_LINUX) += gloffscreen_glx.o
+obj-$(CONFIG_BSD) += gloffscreen_glx.o
```
JayFoxRox, line 5: MacOS is both, `DARWIN` and `BSD`, so you'd probably need something like:
```
ifeq ($(CONFIG_DARWIN),y)
obj-$(CONFIG_DARWIN) += gloffscreen_cgl.o 
else
obj-$(CONFIG_BSD) += gloffscreen_glx.o 
endif
```

or (less QEMU-ish, because they never seem to check with `ifneq` for those `CONFIG_` flags):

```
ifneq ($(CONFIG_DARWIN),y)
obj-$(CONFIG_BSD) += gloffscreen_glx.o 
endif
```

I'm not sure if there's a more elegant way

JayFoxRox:
The commit name is bad. Please follow the existing prefix-style.
Check the history of touched files to find a suitable prefix, and try to mimic the code and commit style.

build.sh
Martinfx: FreeBSD uses gmake tool

GXTX: I think what he's suggesting is the variable name is slightly misleading. `make_opts` would imply that you're changing some options *for* make not *what* you're using to compile.

hw/xbox/nv2a/gl/Makefile.objs
Martinfx: This line its not correct ?

build.sh
Martinfx: I changed name variable

hw/xbox/nv2a/gl/Makefile.objs
Martinfx: I changed to right code style

build.sh
Martinfx: @GXTX  ?

GXTX: What?

Martinfx: @GXTX do you have next comments for this PR ?

GXTX:
Looks good to me, not breaking CI stuff and I imagine you've tested locally. Thanks!

build.sh
```diff
@@ -17,6 +17,7 @@ debug_opts='--enable-debug'
 user_opts=''
 build_cflags='-O0 -g'
 job_count='4'
+make_tool='make'
```
JayFoxRox, line 20: nit: I still think `gnu_make` or just `make` would be a better choice. But current name is okay.

hw/xbox/nv2a/gl/Makefile.objs
```diff
@@ -1,4 +1,8 @@
 obj-y += gloffscreen_common.o
+obj-$(CONFIG_LINUX) += gloffscreen_glx.o
```
JayFoxRox, line 2: nit: Why was this re-orderered? Avoid touching unaffected lines.

This also leads to a weird ordering of:
- Linux
- Windows
- macOS
- BSD

which is neither alphabetical, or by popularity.

build.sh
JayFoxRox: This is near the bottom now, and not near the top (like I had suggested). So I consider this unresolved and more problematic than before.

*Please do not resolve conversations which you didn't start; give reviewers the chance to resolve conversation when they believe the PR author has addressed all issues.*

JayFoxRox:
Only nits remain, although I'd prefer if they were fixed.

Another nit: the prefix "build:" (used for newer commits) should probably be preferred over "build.sh:" (was mainly used for older commits).

And a final note: Avoid changing the *PR* title to avoid log-spam. We only really care for *commit* titles, as the PR info doesn't end up in the Git repository. So as long as your PR title is descriptive (even if it's not ideal), you can stick to it, but commit titles should be refined if necessary (to keep the Git history clean).

hw/xbox/nv2a/gl/Makefile.objs
Martinfx: ok

GXTX:
Looks good! Thank you.

Martinfx:
Please merge this pull request and close. Thank you

build.sh
```diff
@@ -65,6 +66,12 @@ case "$(uname -s)" in # adjust compilation option based on platform
         sys_opts='--python=python3 --disable-cocoa --disable-opengl --disable-fortify-source'
         postbuild='package_windows' # set the above function to be called after build
         ;;
+    Free*|Open*|Net*|DragonFly*|BSD)
+        echo 'Compiling for BSD…'
+        sys_cflags='-march=native'
+        sys_opts='--disable-kvm --disable-xen --disable-werror --enable-bsd-user'
```
thrimbor, line 72: I don't think BSD userland emulation should get enabled here.

PR #298, opened 2023-02-08
Allow xqemu to be built on macOS Ventura.

calmsacibis995:
When trying to build xqemu, the script fails with this: "ERROR: "cc" either does not exist or does not work". The issue in this case is the '-march=native' flag. Clang will complain that it is an unknown argument.


MayeulC (approved):
I'd say go for it, especially as it is built with `-O0 -g` by default (non-release).

